PR #19, opened 2018-11-22
new directory structure: fixes #15

### LINKED ISSUE #15: Update directory structure
As a result of investigating various component libraries, (https://github.com/BBC-News/simorgh/issues/972), we found this folder structure to be useful:

```
src
├── components
│   ├── example-button
│   └── headings
└── contexts
└── utilities
```

We should update the directory structure to this.  
We can do this by updating the `lerna.json` from 
```
"packages": ["packages/*"]
```
to 
```
"packages": ["src/components/*"]
```

We should ensure that all the npm scripts in the main `package.json` file all work correctly. e.g. `npm run build`, etc.

ChrisBAshton:
We want a structure that enables things like contexts and utilities alongside components:

src
├── components
│   ├── example-button
│   └── headings
└── contexts
└── utilities


ChrisBAshton:
Hmm, Snyk can't detect issues in the example-button/package.json: https://support.snyk.io/webhooks/failed-to-detect-issues

It's looking in the wrong place:

screen shot 2018-11-22 at 10 18 09

Do we need to update a config on snyk.io?

jtart:
There is a convention in lerna for packages to live inside packages, i.e. `packages` -> `components`. Is there any reason to not do that?

ChrisBAshton:
We don't necessarily want to expose contexts & utilities as packages - though we could then follow this structure:

```
packages
  ├── example-button
  └── headings
contexts
utilities
```

(ie contexts and utilities at the root level)

Come to think of it, wouldn't we _want_ to expose contexts as packages? (Perhaps not utilities though)

jtart:
I don't believe you have to publish _every_ package under Lerna to NPM, but it would be beneficial if Lerna managed all these things :)

ChrisBAshton:
Soooo are we saying #15 is invalid? Because changing back to `packages/` structure just changes it back to what it was.

If so, feel free to close this PR and the issue 👍

sareh:
@jtart The convention of using `packages` makes sense, but when some of the children of a `packages` directory are not actually packages (e.g. utilities & themes) it just creates confusion.

ChrisBAshton:
Sounds like we're agreeing to close this PR & related issue, and move towards this sort of directory structure:

```
packages
  ├── example-button
  └── headings
contexts
utilities
```

PR #27, opened 2018-11-22
fix #16: prepares test env for simorgh components

### LINKED ISSUE #16: Remove Enzyme tests - for now
We'll move the Headings component from Simorgh to Psammead in #12.

`Headings` has its own set of snapshot tests but is not tested with Enzyme. This is the case for all of the Simorgh components so far.

So that we are not held up on the component migration, let's port the snapshot tests over as-is. An investigation as to a better testing library will be happening in https://github.com/BBC-News/simorgh/issues/671, at which point we may update the component tests.

Rather than copy the component _and_ change the testing framework in one PR, this PR can handle removing Enzyme and the Enzyme tests.

ChrisBAshton:
Resolves #16

removes example spec file (was failing anyway), and prepares Jest environment ready for Simorgh component migration.

- [ ] Test engineer approval


package.json
```diff
@@ -51,12 +51,17 @@
     "storybook-readme": "^4.0.0-beta1"
   },
   "jest": {
-    "resetModules": true,
-    "testMatch": [
-      "**/?(*.)(spec).js?(x)"
+    "collectCoverageFrom": [
+      "**/(src|scripts)/**/*.{js,jsx}",
```
sareh: We don't have a `scripts` directory.

package.json
```diff
@@ -51,12 +51,17 @@
...
+      "!**/src/app/helpers/tests/**",
```
bcmn: Does this make strange assumptions about the structure of the file system? I think especially L56, the others seem fairly agnostic. 🤔

package.json
```diff
@@ -51,12 +51,17 @@
...
+      "!**/*.stories.jsx"
+    ],
+    "setupFiles": [
+      "./jest-setup.js"
     ],
-    "testPathIgnorePatterns": [
-      "<rootDir>node_modules/"
+    "testMatch": [
+      "**/__tests__/**/*.js?(x)",
+      "**/?(*.)+(spec|test).js?(x)"
```
sareh: We should only allow `test` files rather than allowing both `spec` and `test` files.

jest-setup.js
```diff
@@ -0,0 +1,3 @@
+import fetch from "jest-fetch-mock"; // eslint-disable-line import/no-extraneous-dependencies
```
bcmn: Is there a reason this is an extraneous dependency? Surely it could be declared in devDependencies? 🤔 I feel like I must be missing context... 😅

package.json
ChrisBAshton: Oh, it does! Well spotted. Can probably delete this line.

jest-setup.js
ChrisBAshton: It can't just be put into devDependencies as we actually need to overwrite the fetch module at runtime, which this is doing...

...BUT we only need to do that where we're testing the `await fetch('data.json')` stuff - which is Simorgh rather than Psammead specific 🤔 so again, can remove! Good spot.

radiocontrolled (dismissed):

sareh (changes requested):

bcmn (approved):

sareh (approved):

PR #60, opened 2018-11-27
Configure Stylelint to error on !important

### LINKED ISSUE #39: Stylelint should error on `!important`
Currently if you add an `!important` style to your component, the style linter doesn't complain.

An error should be raised by Stylelint if `!important` is used. It should be possible to suppress the error by either removing `!important`, or by adding a magic comment to explicitly allow it.

N/A

**Testing notes**
[Tester to complete]

Dev insight: manual verification that the linter does as expected should be sufficient.

**Additional context**
`!important` is generally bad-practice and we should discourage its usage.

- [ ] Initially labelled with ["Refinement needed"](https://github.com/BBC-News/psammead/labels/Refinement%20Needed)

bcmn:
Resolves #39 

Adds the [declaration-no-important rule](https://github.com/stylelint/stylelint/tree/master/lib/rules/declaration-no-important) to `.stylelintrc`.

- [ ] Test engineer approval

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,4 +1,7 @@
 {
+  "rules": {
+    "declaration-no-important": true
+  },
   "processors": ["stylelint-processor-styled-components"],
   "extends": [
     "stylelint-config-recommended",


ChrisBAshton (approved):
Looks good!

I was a bit worried about the horrible-looking ["If you always want !important in your declarations... you can safely add them using postcss-safe-important"](https://github.com/stylelint/stylelint/tree/master/lib/rules/declaration-no-important#declaration-no-important) but it should be possible to ignore on a case-by-case basis [the normal way](https://github.com/stylelint/stylelint/issues/714#issuecomment-244906775).

dr3 (approved):

PR #74, opened 2018-11-28
Use babel on all packages

### LINKED ISSUE #73: Build packages for publish
**Describe the bug**
Currently our packages are ES6 when published and thus have issues when pulling them into simorgh as it doesnt run node modules through babel.

**To Reproduce**

1. Import @bbc/psammead-styles into simorgh
2. See error `Unexpected Token Export`

**Expected behavior**
The module should import correctly

**Solutions**
- Build them pre-publish
- Whitelist the packages in simorgh node modules to be babel-ified

- [x] Initially labelled with ["refinement-needed"](https://github.com/BBC-News/psammead/labels/refinement-needed)

dr3:
Resolves #73

Moves constant package files into a src folder, so they can be transpiled using babel and published.

You generate the built versions using `npm run build` which runs pre publish

- [ ] Test engineer approval


dr3:
Synk doesnt know where to look for the example-button package now that its been moved. Will require a force push, when merged we can tell it to use the correct location

lerna.json
```diff
@@ -1,6 +1,6 @@
 {
   "packages": [
-    "packages/*"
+    "packages/*/*"
```
ChrisBAshton, line 3: Is this a valid glob pattern?

packages/utilities/gel-constants/breakpoints.js
```diff
@@ -1,36 +1 @@
...
-export const GEL_GROUP_4_SCREEN_WIDTH_MAX = `80em`; // 1279px
-
-export const GEL_GROUP_5_SCREEN_WIDTH_MIN = `80em`; // 1280px
-
-export const MEDIA_QUERY_TYPOGRAPHY = {
-  SMART_PHONE_ONLY: `@media (min-width: ${GEL_GROUP_B_MIN_WIDTH}em) and (max-width: ${GEL_GROUP_B_MAX_WIDTH}em)`,
-  SMART_PHONE_AND_LARGER: `@media (min-width: ${GEL_GROUP_B_MIN_WIDTH}em)`,
-  LAPTOP_AND_LARGER: `@media (min-width: ${GEL_GROUP_CD_MIN_WIDTH}em)`,
-};
+module.exports = require('./dist/breakpoints'); // eslint-disable-line import/no-unresolved
```
ChrisBAshton, line 1: Not keen on having to change the way we write our components to get an ES2015 version out 🤔 

Can we take inspiration from [the way material-ui does it](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/package.json#L29-L30)?

I'd hope it's just a case of adding a script or two to our `package.json`.

ChrisBAshton (changes requested):
We should ideally be able to keep the existing `packages/` structure, and all components untouched.

sareh:
@dr3 Since this is in Code Review, I looked up creating ES5 versions in the this with Webpack 4 & Babel 7. 
Here's one way to generate ES5 versions in the dist directory.

```
cd packages/utilities/gel-constants
npm i -D @babel/core @babel/preset-env babel-loader webpack webpack-cli
mkdir dist
touch .babelrc
touch webpack.config.js
```

`.babelrc`
```
{
    "presets": [
        "@babel/preset-env"
    ]
}
```
`webpack.config.js`
```
var webpack = require('webpack');
var path = require('path');

var config = {
  entry: {
    breakpoints: path.resolve(__dirname, './src/breakpoints.js'),
    spacings: path.resolve(__dirname, './src/spacings.js'),
  },
  output: {
    filename: '[name].js',
    path: path.resolve(__dirname, './dist'),
  },
  module: {
    rules: [
    {
      test: /(\.js)$/,
      loader: 'babel-loader',
    },
    ]
  }
};

module.exports = config;
```

In `package.json`:
```
  "scripts": {
    "build": "webpack --mode production",
    "dev": "webpack --mode development"
  },
```
Then an `npm run dev` will generate the files in the `dist` directory. An `npm run build` will generate the minified files in the directory. 
And your top-level `package.json` file already has the prepublish script `npm run build`.

lerna.json
dr3: It will cover the pattern `/packages/components/thing` and `/packages/utilities/thing`

packages/utilities/gel-constants/breakpoints.js
dr3: Does material UI just dyamically build a new version of the packages? sorry not sure which part your referring to

dr3: Im not sure how we could do that when it comes to using lerna. You can use a lerna command to link all packages togeather, but that might not work if we generate the files somewhere else. And if we do it in place it might get a bit crazy with .gitignores to make sure we dont put them in git. If you can think of a nice way to do it though im all ears :)

dr3: Bare in mind this is only for these constants libraries where we have chosen to address them using `@bbc/psammead-styles/colours`, pretty much every normal component wont need to do this. IMO this seems like the best thing to do for now, espesh as its blocking everything, if we can think of a better way we could come back in alpha 3

ChrisBAshton: I guess we'd define a `build` command in the root-level `package.json` which goes through all the components and transpiles them.

```
packages/
  my-package/
    src/
      index.jsx
    dist/ # generated by top-level 'build' command. git-ignored
    package.json
```

And the component-level package.json would have `main: "dist/index.js"` and `files: ["dist"]` properties. That way it should be possible for consumers to write `import foo from "@bbc/my-package"` as well as `import bar from "@bbc/my-package/fonts"`.

Only downsides I can think of are 1) symlinking (need to run the `build` command when debugging locally) and 2) unable to pull in untranspiled versions for modern browsers (unless we expand the `files` property to `files: ["dist", "src"]`).

Upsides are that we don't have to change the way we write components - it's still ES6, it's just down a level (in `src/`). And we can still use Lerna to manage deployments (would be `npm run build && lerna publish --independent` from the root).

dr3: Running it locally I cant get `"files": ["dist"]` to work, is it only when publishing? Mine only works when that `module.exports = require('./dist/colours');` file is at the base dir

dr3: Looking at https://docs.npmjs.com/files/package.json#files it looks as if this doesnt dictate what files it points at, just what are packaged up. Not sure what you meant about symlinking though

ChrisBAshton: What I mean is if `main: "dist/index.js`, and we then `npm link` the module,
then in our application we `npm link @bbc/psammead-foo`,
when we `import foo from "@bbc/psammead-foo"` then it will try to pull in `dist/index.js` - but `dist/index.js` doesn't exist until you run the `build` step at the root of Psammead.

dr3: Yeah thats fine, the reason for these ES5 files is that were not using an index.js for these packages. If we did ```import { colours } from '@bbc/psammead-styles'``` we could, but as were going with ```import { foo } from '@bbc/psammead-styles/colours'``` we cant use an index.js and thus need this pass through

We are doing what you said in example-button though, which we will do for all normal components that arnt imported like these

sareh: @ChrisBAshton @dr3 I'd suggest merging this PR as it is, since it works well and provides the API for consumers of the package.  
The optimisation of the package internals can be dealt with at a later date.  Right now, the absence of the functionality this PR is adding is blocking all our other tickets.  
I'd be happy to open an issue that captures all of your concerns, Chris, and restructure the package internals later.

sareh (approved):
👍 Looks good.  `npm run build` works well. 

As a result of the discussion in the thread, https://github.com/BBC-News/psammead/pull/74/files#r237470595 I'll be happy to open a separate issue that restructures the internals of our packages at a later date.

packages/utilities/gel-constants/breakpoints.js
pjlee11: Agreed with Sareh, however could we have the issue raised and commented here before the PR is merged. It could be labbeled with `alpha-3`

dr3: I have created https://github.com/BBC-News/psammead/issues/78

pjlee11 (approved):
Nice work!

ChrisBAshton (approved):
Happy to proceed as-is. Nice one 👍

PR #80, opened 2018-11-29
Move test helpers to psammead

dr3:
Test helpers are required for https://github.com/BBC-News/psammead/pull/79 so having them as a package is useful IMO

- ~[ ] Tests added for new features~
- ~[ ] Test engineer approval~


packages/utilities/psammead-test-helpers/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Installation
+
+```jsx
+npm install @bbc/psammead-test-helpers --save
+```
+
+## Usage
+
+```jsx
```
sareh, line 13: I think it should be important to explain what the `shouldMatchSnapshot` function does. e.g. It renders the component passed the second argument using `react-test-renderer`, converts it to JSON and asserts that it matches the given snapshot, which will be saved in the `__snapshots__` directory. 
The first argument is the title for the test. 

Likewise, since we've added `shallowRender`,  `shouldShallowMatchSnapshot`, `isNull`, we should explain these as well.

packages/utilities/psammead-test-helpers/README.md
```diff
@@ -0,0 +1,36 @@
+# psammead-test-helpers &middot; [![GitHub license](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](https://github.com/BBC-News/psammead/blob/latest/LICENSE) [![npm version](https://img.shields.io/npm/v/@bbc/psammead-test-helpers.svg)](https://www.npmjs.com/package/@bbc/psammead-test-helpers) [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](https://reactjs.org/docs/how-to-contribute.html#your-first-pull-request)
+
+This package is a collection helper methods used for snapshot testing, with support for styled components.
+
+## Installation
+
+```jsx
+npm install @bbc/psammead-test-helpers --save
```
sareh: To be able to use `psammead-test-helpers`, a consumer would need to also install `react react-dom react-test-renderer`.

sareh (changes requested):
Looks great. Just some points to add in the Readme to make it clear how to use the component.

sareh (approved):
Looks great. 👍

ChrisBAshton (commented):
Hmmm tbh I don't see much value in this being a package, rather than just living at the root level of the repo - it's not a component, and it's not a utility that will be consumed by components. Not something we'll need versioning for.

e.g. will we be using this code directly to run tests, or will we have an `npm install @bbc/psammead-test-helpers` in our CI? Is that confusing?

I don't feel too strongly about it, but want to know that we've considered it first. 🤔

sareh:
@ChrisBAshton We'll need these test helpers in both Psammead and in Simorgh, since we'll still have Containers in Psammead that need snapshot tests. For that reason, it makes sense to have this as a single package that can be used by both repositories. 

In future, other teams with their own component libraries could use the package as a quick way to get started with their tests.

ChrisBAshton (approved):
Fair 👍

PR #94, opened 2018-12-04
Update Psammead to Babel 7 & Babel-Loader 8

radiocontrolled:
Resolves https://github.com/BBC-News/psammead/issues/76

Updates Psammead to use the latest Babel releases: Babel 7 & babel loader 8.

- [N/A] Tests added for new features
- [ ] Test engineer approval

PR #96, opened 2018-12-04
Update babel cli

ChrisBAshton:
Upgrades `babel-cli` to `@babel/cli`. Trickier than hoped because of Jest - see git commits for comments.

- [ ] Test engineer approval


package.json
```diff
@@ -65,6 +66,7 @@
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
     "react-test-renderer": "^16.6.3",
+    "regenerator-runtime": "^0.13.1",
```
dr3: Whats this? I don't need it locally

ChrisBAshton: I think you'll find you do 🤔 I tried removing it and it errored, so I added it back in.

Needed by Jest.

radiocontrolled (approved):
👍

package.json
dr3: Hmm running `npm run test` and `npm run build` I dont need them :/ Not sure why I dont and you do, strange

ChrisBAshton: Just double-checked - you're right! `jest` used to fail without it, but perhaps I was missing the env stuff in babel.config.js at that time.

Have now removed. Thanks!

dr3 (approved):

ChrisBAshton:
For #94.

PR #107, opened 2018-12-05
psammead-test-helpers - use react-testing-library

### LINKED ISSUE #95: Deep snapshots have ForwardRef instead of component names
**Describe the bug**
When using deep rendering with `styled-components` V4, snapshot references to other React components are replaced with `<ForwardRef />`. [An example of this is visible in the original Simorgh PR for styled-components V4](https://github.com/BBC-News/simorgh/pull/695/files#diff-21c2870128b7d951044c5459438579f6).

Luckily, as most Psammead components are currently only composed to a depth of 1, this isn't yet visible in this repo. However, it will become an issue as we introduce more complex or nested components.

**Possible Solutions/Approaches**
This issue is caused by our dependency `jest-styled-components`. If we stick with using jest-styled-components alone, we will have to await or contribute to their resolution of this issue. They have a [ticket on it here](https://github.com/styled-components/jest-styled-components/issues/191). We could potentially contribute to the PRs that spawn out of it if we considered it a high priority.

Alternatively, if we are evaluating using Enzyme or react-testing-library alongside jest-styled-components, one of the benefits from either of these moves would be that this would no longer be an issue.

- [x] Initially labelled with ["bug"](https://github.com/BBC-News/psammead/labels/bug)

sareh:
Resolves #95

To see the impact that such a change has on snapshot tests, you can look at this Simorgh PR https://github.com/BBC-News/simorgh/pull/1050  

When the `shouldMatchSnapshot` is used currently, we get this:
```
exports[`Headings with headline data should render correctly 1`] = `
<ForwardRef
  text="This is a headline!"
>
  This is a headline!
</ForwardRef>
`;
```
Reference: https://github.com/BBC-News/simorgh/pull/1026

After this PR, if we use `shouldMatchSnapshot`, we get this output: 
```
exports[`Headings with headline data should render correctly 1`] = `
<DocumentFragment>
  .c0 {
  color: #222222;
  font-family: ReithSerifNewsMedium,Helvetica,Arial,sans-serif;
  margin: 0;
  padding: 2rem 0 1rem 0;
  font-size: 1.75em;
  line-height: 2rem;
}

@media (min-width:20em) and (max-width:37.4375em) {
  .c0 {
    font-size: 2em;
    line-height: 2.25rem;
  }
}

@media (min-width:37.5em) {
  .c0 {
    font-size: 2.75em;
    line-height: 3rem;
  }
}

<h1
    class="c0"
  >
    This is a headline!
  </h1>
</DocumentFragment>
`;
```
Reference: 
https://github.com/BBC-News/simorgh/blob/cade24c1c1853c5329298b49cd7b290662a46af6/src/app/containers/Header/__snapshots__/index.test.jsx.snap  in PR https://github.com/BBC-News/simorgh/pull/1050
- Change `shouldMatchSnapshot` and `isNull` to use `react-testing-library`. 
- Installed dev dependencies `react` & `react-dom` since they're needed by `react-testing-library` and the existing `react-test-renderer`. 

- [ ] Test engineer approval


packages/utilities/psammead-test-helpers/package.json
```diff
@@ -25,7 +25,12 @@
     "snapshot"
   ],
   "dependencies": {
-    "jest-styled-components": "^6.3.1",
-    "react-test-renderer": "^16.6.3"
+    "jest-styled-components": "^6.3.1"
+  },
+  "devDependencies": {
+    "react": "^16.6.3",
```
dr3: When I run this locally I dont need `react` and `react-dom`? Are you sure these are needed?

packages/utilities/psammead-test-helpers/package.json
```diff
@@ -25,7 +25,12 @@
...
+    "react-dom": "^16.6.3",
+    "react-test-renderer": "^16.6.3",
+    "react-testing-library": "^5.3.1"
```
dr3, line 30: As this is a dev library, I think these should be `dependencies` not `devDependencies`

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -1,11 +1,11 @@
-import renderer from 'react-test-renderer';
+import { render } from 'react-testing-library';
 import 'jest-styled-components';
 import ShallowRenderer from 'react-test-renderer/shallow';
 
 export const shouldMatchSnapshot = (title, component) => {
   it(title, () => {
-    const tree = renderer.create(component).toJSON();
-    expect(tree).toMatchSnapshot();
+    const { asFragment } = render(component);
```
dr3, line 7: Running this on psammead-headings I go from 
```
exports[`Headline component should render correctly 1`] = `
.c0 {
  color: #222222;
  font-family: ReithSerifNewsMedium,Helvetica,Arial,sans-serif;
  margin: 0;
  padding: 2rem 0 1rem 0;
  font-size: 1.75em;
  line-height: 2rem;
}

@media (min-width:20em) and (max-width:37.4375em) {
  .c0 {
    font-size: 2em;
    line-height: 2.25rem;
  }
}

@media (min-width:37.5em) {
  .c0 {
    font-size: 2.75em;
    line-height: 3rem;
  }
}

<h1
  className="c0"
>
  This is my headline.
</h1>
`;
```
to
```
exports[`Headline component should render correctly 1`] = `
<DocumentFragment>
  <h1
    class="sc-bdVaJa gMUPYM"
  >
    This is my headline.
  </h1>
</DocumentFragment>
```
It feels like were losing a lot of information here

dr3 (changes requested):

packages/utilities/psammead-test-helpers/src/index.js
sareh: Interesting. 🤔 Will look into why it's not including the styles in this case, but was when I included it in Simorgh container. I'm not sure about the source of the inconsistency.

dr3:
Have raised https://github.com/kentcdodds/react-testing-library/issues/236 because of the issue found linking these packages locally

ChrisBAshton:
Decided in standup yesterday we'd close this ticket for now.

Thanks for looking at this, Sareh.

PR #180, opened 2018-12-10
Add script to publish packages

### LINKED ISSUE #132: Create script to publish unpublished packages
When a PR is merged, all unpublished packages should be published. This will happen on the CI (in #166) however we can have a script for now to do this manually.

A script that checks all packages:

* if `"publish": "false"`, ignore
* else, compares version with the latest version in the NPM registry
* if version different, publishes

Summary output should include which packages have changed, from what version to what version, and what succeeded and what failed.

**Additional thoughts**
We should also have a way to stop a package from being published if they dont want it to be. Maybe look for a `"publish": "false"` value in their package json?

A POC PR has been made https://github.com/BBC-News/psammead/pull/123, it can be reopened once refined if needed.

**Testing notes**
[Tester to complete]

Dev insight: If we bump a package version, then run the publish script, it should publish to NPM.

dr3:
Resolves #132

Creates node script to publish packages

image

image

- [ ] Test engineer approval


scripts/publish/src/getOtpTag.js
```diff
@@ -0,0 +1,14 @@
+// Builds opt tag for publish command.
+// Accepts both '--otp 123456' and '--otp=123456' format.
+module.exports = () => {
```
sareh: I think using https://www.npmjs.com/package/yargs-parser could be better than implementing argument parsing here.

scripts/publish/src/getPackages.js
```diff
@@ -0,0 +1,7 @@
+const { exec } = require('shelljs');
+
+// Get all packages in lerna scope.
+module.exports = () =>
+  exec('npx lerna ls --parseable', { silent: true })
```
sareh, line 5: Is there a reason why you're using lerna scope instead of looking at `/packages/*/*`? We're not using lerna publish, so it might be better to just simplify things and just look in the directory structure.

dr3: I chose this as it would always match the scope of lerna, which we use to build and install packages. But yes I could just use `fs`

scripts/publish/src/slackNotification.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable no-console */
+const request = require('request');
+const chalk = require('chalk');
+
+// This value will need to be made available to the script in jenkins.
+const slackToken = 'XXXXXXXXX/XXXXXXXXX/XXXXXXXXXXXXXXXXXXXXXXXX';
```
sareh: Can be passed in as a `process.env.SLACK_TOKEN` & we ensure Jenkins has this environment variable.

scripts/publish/src/getRegistry.js
```diff
@@ -0,0 +1,7 @@
+const { exec } = require('shelljs');
+
+// Get version of package in NPM regsitry. Returns 0.0.0 if doesn't exist.
+module.exports = name => {
+  const npmVersion = exec(`npm show ${name} version`, { silent: true }).stdout;
```
sareh: Would be good to use async/await whenever you're executing commands and need the response:

```
module.exports = async name => {
    const npmVersion = await exec(`npm show ${name} version`, { silent: true })
...
```

dr3:
Changed the publish restrictor from `"publish": "false"` to `"private": true` as to match NPM standards https://docs.npmjs.com/files/package.json#private

packages/components/example-button/package-lock.json
```diff
@@ -1,233 +1,233 @@
 {
```
bcmn: Why has this file changed? It looks like it's just formatting.

scripts/publish/test/publish.test.js
```diff
@@ -0,0 +1,117 @@
...
+    expect(attempted.failure.length).toEqual(1);
+    expect(attempted.failure[0]).toEqual('@foo/psammead-foobar@0.1.2');
+
+    expect(slackNotification).toHaveBeenCalledWith(
+      '@foo/psammead-foobar@0.1.2',
+      false,
+    );
+  });
+
+  it('runs correct publish command when opt argument is set and publish is successful', () => {
```
bcmn:
```suggestion
  it('runs correct publish command when otp argument is set and publish is successful', () => {
```

Typo.

scripts/publish/test/report.test.js
```diff
@@ -0,0 +1,100 @@
+/* eslint-disable no-console */
+const stripAnsi = require('strip-ansi');
+const report = require('../src/report');
+
+// const exit = jest.spyOn(process, 'exit');
```
bcmn: Why is this still here?

bcmn (commented):
Which of the five children was called Publish?

bcmn (dismissed):

scripts/publish/src/publish.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable no-console */
+const { exec } = require('shelljs');
+const chalk = require('chalk');
+const argv = require('yargs-parser')(process.argv.slice(2));
```
pjlee11, line 4: What is `slice(2)` doing in this case?

scripts/publish/src/publishConfig.js
```diff
@@ -0,0 +1,17 @@
+module.exports = packageJson => {
+  let access = 'public';
+  let tag = 'latest';
+
+  // Set access and tag based on value in package.json config.
+  if (packageJson.publishConfig) {
+    if (packageJson.publishConfig.access) {
+      ({ access } = packageJson.publishConfig);
+    }
```
pjlee11, line 9: As a future improvement it would be nice to have an else like so:
```
else {
    console.log("No access value found in config. Using default access value of 'public'");
}
```

scripts/publish/src/publishConfig.js
```diff
@@ -0,0 +1,17 @@
...
+
+    if (packageJson.publishConfig.tag) {
+      ({ tag } = packageJson.publishConfig);
+    }
```
pjlee11, line 13: As a future improvement it would be nice to have an else like so:
```
else {
    console.log("No tag value found in config. Using default tag value of 'latest'");
}
```

scripts/publish/src/report.js
```diff
@@ -0,0 +1,33 @@
...
+  );
+
+  if (successfulCount) {
+    console.log(chalk.underline.green('\nSuccessful'));
+    attempted.success.forEach(published => console.log(chalk.green(published)));
+  }
+
+  if (failureCount) {
+    console.log(chalk.underline.red('\nFailed'));
+    attempted.failure.forEach(published => console.log(chalk.red(published)));
```
pjlee11: Does the variable in the foreach being named `published` make sense? Are these not going to be failures as compared to publishes?

pjlee11:
Few comments that could be addressed, at a first pass v good work

scripts/publish/src/publish.js
dr3: Removing The first two args which is `node` and the file of the script, Based on recomended https://www.npmjs.com/package/yargs-parser#example

scripts/publish/src/publishConfig.js
dr3: Sounds good, added

scripts/publish/src/publishConfig.js
dr3: Sounds good, added

pjlee11 (approved):

sareh (approved):
👍 Looks great!

bcmn (approved):

dr3:
Talking to @sareh And @jamesbrumpton We are going to merge this and fix forward any issues that arise with https://github.com/BBC-News/psammead/issues/166

PR #185, opened 2018-12-11
Jenkins - notify Slack

ChrisBAshton:
Resolves #165

Notifies Slack of build fail/pass/succeed.

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Tests added for new features
- [x] Test engineer approval

PR #179, opened 2018-12-10
Add Code Climate badges to main README

bcmn:
Resolves #171

Adds Code Climate badges to main README. Blocked on #174, as it added the other badges to the README.

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Tests added for new features
- [ ] Test engineer approval

PR #188, opened 2018-12-12
Simplify jenkinsfile

bcmn:
Resolves #N/A. Blocker for #179. Blocker for #185.

The stages in the Jenkinsfile looks to have been copied from an infrastructure repo Jenkinsfile, which clones the application repo to a `APP_DIRECTORY`, in order to perform tests across both. This is unnecessary for this repo, & was only causing headaches as I tried to add to the CI steps.

In addition, this PR ensures that `npm install` is not run on the packages -- instead, we should be using `npm ci`.

Furthermore, it appears the application test stage was only being run on non-latest branches. This does not seem desirable for this repo, so I have removed this conditional.

- [ ] Test engineer approval


package.json
```diff
@@ -5,10 +5,11 @@
   "main": "index.js",
   "scripts": {
     "build": "lerna exec --parallel -- babel src -d dist --ignore '**/*.test.jsx,**/*.stories.jsx' --root-mode upward",
+    "ci:packages": "lerna exec -- npm ci",
     "clear_styled_components": "for d in ./packages/*/*; do rm -rf $d/node_modules/styled-components; done; echo \"Cleared nested packages/**/node_modules/styled-components\"",
     "deploy-storybook": "build-storybook -o storybook_dist && gh-pages -d storybook_dist",
     "install:packages": "lerna exec -- npm install",
-    "postinstall": "npm run install:packages",
+    "postinstall": "npm run ci:packages",
```
dr3, line 12: IMO we should remain doing `npm run ci:packages` as a post install step here, Until we decide and create a process for ensuring packages don't go stale.

Obs it wont scale, but for now I think its our best bet untill we work out a process

dr3: We could leave this as it is and do `make install` => `npm ci --ignore-scripts; npm run ci:packages`

bcmn: `--ignore-scripts;` is unusable -- [I believe it ignores the postinstall scripts of dependencies, too.](https://stackoverflow.com/a/23506053)

bcmn: Just to summarise the arguments put forward in [another place](https://en.wikipedia.org/wiki/Another_place):

* We could explore changing the postinstall step to something like `if !ENV.IS_CI; do npm run install:packages; done`, but we like knowing that our local env is the same as CI env, & that tests will produce the same results.
* The "obviously it doesn't scale" refers to: Keeping `install:packages` as the postinstall step locally will result in version bumps in packages that you aren't currently working on. When someone wishes to make changes to `psammead-copyright`, & `psammead-headlines` has received new dependency versions then they have to be careful not to commit those changes, & those changes may cause them issues if dependency bumps have accidentally introduced errors or behaviour changes. This will only increase as we gain more components.

bcmn: [An example of us already bumping into the issue described above on a PR](https://github.com/BBC-News/psammead/pull/184#discussion_r241009497), albeit worsened by the npm versioning woes.

ChrisBAshton: So the summary is: we're sticking with `npm ci`?

sareh (dismissed):
👍 Looks great! https://ci.news.tools.bbc.co.uk/blue/organizations/jenkins/psammead/detail/simplify-jenkinsfile/9/pipeline

dr3 (dismissed):

bcmn:
Quickly added a line to ensure the app directory is removed before running tests (workspaces don't seem to be cleaned on each run). If it isn't, it can lead to a Jest Haste error. This seems like the most reliable solution for now.

ChrisBAshton (approved):

dr3 (approved):

sareh (approved):

jamesbrumpton:
Happy for this to be merged after speaking with the devs around the work/checks.

PR #190, opened 2018-12-12
Locks us into Node v8.11.3

ChrisBAshton:
See https://github.com/BBC-News/simorgh/pull/1067 for more details

This PR aligns us all onto v8.11.3.

To test:

* checkout this branch
* `rm -rf node_modules package-lock.json`
* `npm install`
* `git diff` - should not be any different!
* `npm test` - should still pass

Note that the above testing instructions are time sensitive (i.e. if a dependency changes a subdependency version, we likely will have a diff).

- [ ] Test engineer approval


bcmn (approved):

sareh:
We should update the Jenkinsfile here to ensure Jenkins is using 8.11.3 as well. https://github.com/BBC-News/psammead/pull/191/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R4

Package locks for the packages need to be updated too.

dr3 (approved):
LGTM Locally

PR #230, opened 2018-12-19
Add README decorator to Image & ImagePlaceholder stories

bcmn:
Resolves #N/A

The ImagePlaceholder story doesn't display the README, which is noticeably inconsistent with all other stories. This PR fixes that.

- [ ] ~Tests added for new features~
- [ ] Test engineer approval

### DIFF
--- a/packages/components/psammead-image-placeholder/src/index.stories.jsx
+++ b/packages/components/psammead-image-placeholder/src/index.stories.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
-import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
+import { storiesOf } from '@storybook/react';
+import { withReadme } from 'storybook-readme';
+import Readme from '../README.md';
 import ImagePlaceholder from '.';
 
 const landscapeImageRatio = 56.25;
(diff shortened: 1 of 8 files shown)


sareh (dismissed):

packages/components/psammead-image-placeholder/src/index.stories.jsx
```diff
@@ -1,12 +1,15 @@
 import React from 'react';
-import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
+import { storiesOf } from '@storybook/react';
```
ChrisBAshton, line 2: Why do we no longer need the eslint-disable?

ChrisBAshton (dismissed):
Looks good 👍 I wonder if there's a way of avoiding doing this for each one? (Having a 'default' Storybook with decorators etc, from the .storybook level)

packages/components/psammead-image-placeholder/src/index.stories.jsx
sareh: That `eslint-disable` was needed in Simorgh, but isn't in Psammead, since there is a global ignore of this rule.

bcmn:
> Looks good 👍 I wonder if there's a way of avoiding doing this for each one? (Having a 'default' Storybook with decorators etc, from the .storybook level)

The global decorator part of this can be done, & we've done it for stuff like global styles in the Simorgh/styled-components-v4 work. However, the README for each package is relative to that package. I'm not clear how we'd achieve (& if we did, whether it would be a fragile solution that fails horribly when the README isn't present for some reason).

sareh (approved):

ChrisBAshton (approved):

jamesbrumpton:
LGTM

PR #239, opened 2018-12-20
Use lerna bootstrap to hoist dependencies

### LINKED ISSUE #217: Add package linking command and documentation
We need a way of linking packages locally for local development

Add an NPM command `npm run link` -> `lerna link --force-local`

And document on the readme (near the top) what this does

**Additional context**

You need `--force-local`, If a package requires `^0.1.0` and you have `0.2.0` locally, it wont link unless you force it, which is weird, but IMO its what we want  https://github.com/lerna/lerna/tree/master/commands/link#readme

### LINKED ISSUE #240: Remove `clear_styled_components` hack
We have an NPM script `clear_styled_components` which removes any `styled-components` in any components' `node_modules/` folder, so that we don't have multiple instances of styled components in Storybook. (When we have multiple instances, the component's CSS isn't always applied - hence the hack).

We've since discovered there is a `lerna boostrap --hoist` command which is the official way of doing things. Follow the thread here: https://github.com/BBC-News/psammead/pull/238#issuecomment-448945116

Delete the hacky script, add the bootstrap method.

N/A

**Testing notes**
[Tester to complete]

Dev insight: test snapshots should be exactly the same as before. Tests should continue to pass. Components should continue to have styling in Storybook.

bcmn:
Resolves #240.

We can use `lerna bootstrap` to both install dependencies for all packages, & hoist common dependencies out of each nested package's `node_modules` & into the monorepo's top-level `node_modules`. This hoisting is [recommended for lerna repos by styled-components](https://www.styled-components.com/docs/faqs#usage-with-lerna), & allows for the removal of the `clear-styled-components` script. By default, `lerna bootstrap` links local packages together, which helps address a discussion point in #217.

`lerna bootstrap` is configured to behave the same as our previous `postinstall` command -- running `npm ci`, instead of `npm install`, on all nested packages.

In doing this, I noticed many of the package-locks committed on `latest` are out of date. This PR fixes that.

**Additional info**

[Lerna bootstrap command documentation.](https://github.com/lerna/lerna/tree/master/commands/bootstrap#readme)

[Lerna's advice on hoisting.](https://github.com/lerna/lerna/blob/master/doc/hoist.md)

[Information on what to run when a nested package-lock is conflicting is in this comment.](https://github.com/BBC-News/psammead/pull/239#issuecomment-450847002)

- [ ] Test engineer approval


ChrisBAshton (dismissed):
Looks good to me 👍

ChrisBAshton (dismissed):

packages/components/psammead-caption/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-caption",
-  "version": "0.1.0",
+  "version": "0.1.1",
```
sareh: I'm surprised that the previous PRs didn't catch these. That's quite worrying! I wonder if the mismatched package/package-lock versions published on npm will cause `npm ci` to error. 
I suppose it'll be fixed next time we publish these packages to npm.

sareh (dismissed):
👍 Looks great.

sareh (dismissed):

ChrisBAshton (dismissed):

sareh:
Investigating why the Jenkins build is failing with this error:
```
npm ERR! Missing: @bbc/psammead-image@^0.1.0
```

I removed the `--ci` flag from the `postinstall` script and ran `npm install` locally.
This removed the `nanoid` package from the `package-lock.json`.  
After that, I returned the `--ci` flag and ran `npm install` again. It errored:
```
npm ERR! Missing: @bbc/psammead-image@^0.1.0
npm ERR! Missing: nanoid@^2.0.0
```
Need to investigate what exactly `lerna bootstrap` is doing under the hood to fix this issue.

bcmn:
Got it passing again, but not clear what the issue was. Going to scrutinise [the compare between the first fail & last pass](https://github.com/BBC-News/psammead/compare/853c39b...cc6494).

bcmn:
Resolved conflicting package-locks above -- Doing so simply seems to be a case of running `lerna exec -- npm install`. I don't believe this brakes the hoisting.

If you want to recreate this to verify:

```
git checkout cc6494c
git pull origin latest
```

This will get you in the scenario of conflicting package-locks. 

Then simply run `lerna exec -- npm install` & run the tests to verify the hoisting hasn't broken.

dr3 (approved):

dr3 (changes requested):
This doesnt appear to be installing the modules correctly. I got a fresh clone, ran `npm i` and in the packages the node_modules arnt correct.

The only packages in the node_modules folder are bbc packages. Others, like styled-components arnt installed

bcmn:
> This doesnt appear to be installing the modules correctly. I got a fresh clone, ran npm i and in the packages the node_modules arnt correct.
> 
> The only packages in the node_modules folder are bbc packages. Others, like styled-components arnt installed

I think this is the understood behaviour of `--hoist`. There's more in [this link from the PR description](https://github.com/lerna/lerna/edit/master/doc/hoist.md), but this seems relevant: 

> When the `--hoist` flag is used:
> 
> * Common dependencies will be installed _only_ to the top-level
>   `node_modules`, and omitted from individual package `node_modules`.
> * Mostly-common dependencies are still hoisted, but outlier packages
>   with different versions will get a normal, local `node_modules`
>   installation of the necessary dependencies.
>   * In this instance, `lerna bootstrap` will always use `npm install`
>     with the `--global-style` flag, regardless of client configuration.
> * Binaries from those common packages are symlinked to individual
>   package `node_modules/.bin` directories, so that `package.json`
>   scripts continue to work unmodified.
> * Well-behaved Node-based software should continue to work unmodified.

I suspect most of our dependencies are "common" or "mostly common".

The tests still pass & the storybook functions.

dr3 (dismissed):
Hmm ok the hoisting seems to be working, curious what issues might arise, but cant see any atm, lets see how it goes.

ChrisBAshton:
I ran an `npm install`and I have a `node_modules/@bbc/psammead-image` folder.

My `package-lock.json` also has removed `@bbc/psammead-image` (around line 993).

Not sure why `@bbc/psammead-image` is being treated differently to our other components 🤔 For example, it's used as a devDependency in ` @bbc/psammead-figure`. But `@bbc/psammead-test-helpers` is also a dev dependency of the same module, and _it_ isn't in `node_modules`. I would have expected `@bbc/psammead-image` to be hoisted & `npm link`ed like everything else...?

dr3:
This issue doesn't happen with --force-local IMO we should use this for the time being, and file an issue with lerna to try and fix this bug.

bcmn:
It's been agreed that we'll use `--force-local` to get this PR merged, & open an issue with Lerna to document some quirks we've observed with their semver range comparisons. The goal is to remove the flag once these quirks have been fixed and/or explained adequately.

dr3 (dismissed):

bcmn:
Following up on my last comment, & this comment, following discussion in [another place](https://en.wikipedia.org/wiki/Another_place):

> I ran an `npm install`and I have a `node_modules/@bbc/psammead-image` folder.
> 
> My `package-lock.json` also has removed `@bbc/psammead-image` (around line 993).
> 
> Not sure why `@bbc/psammead-image` is being treated differently to our other components 🤔 For example, it's used as a devDependency in ` @bbc/psammead-figure`. But `@bbc/psammead-test-helpers` is also a dev dependency of the same module, and _it_ isn't in `node_modules`. I would have expected `@bbc/psammead-image` to be hoisted & `npm link`ed like everything else...?

It's important to be clear about what the scenario was here:

- [`lerna bootstrap` will symlink together packages which are dependencies of each other, so long as the semver range matches.](https://github.com/lerna/lerna/tree/master/commands/bootstrap#readme) If it doesn't match, it will install the requested version into `node_modules`.
- `psammead-image` has a committed version of `0.2.1`.
- `psammead-figure` has a dependency on `psammead-image`, with a range of `^0.1.0`.
- NPM's caret range syntax is defined as ["allowing changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple"](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004). More detailed examples are available in their documentation.
- This is based on [item 4 of Semantic Versioning 2.0.0](https://semver.org/#spec-item-4), which treats _anything_ under 1.0.0 as a potentially breaking change.

As such, this is the correct behaviour -- If `psammead-image` had a version of `0.1.x`, it would be symlinked to the repo version. However, by changing to `0.2.x`, we have exceeded the range specified by `^0.1.0`.

As such, I don't intend to open additional issues on this point. We'll discuss whether we want to keep the `--force-local` flag separately.

dr3 (approved):
LGTM

pjlee11 (approved):
Great work 😎

jamesbrumpton:
LGTM 👍

Issue #147, opened 2018-12-07
Validate components against our Definition of Done


ChrisBAshton:
**Is your feature request related to a problem? Please describe.**
It would be nice to have an automated checker that ensures components correspond to our Definition of Done. Ie.

* they should have a README that includes the expected required sections (contributing, Code of Conduct, licence etc)
* they should have a package.json which includes "repository, author, licence, bugs, homepage"
* they should have a `src/` folder that outputs a `dist/` folder.

...etc.

**Describe the solution you'd like**
A script that validates anything that can be automated, and that runs on the PR and offers a status check. It should have a red cross when the component doesn't validate, to prevent merge.

**Describe alternatives you've considered**
Continuing as we are, which is to check these things manually.

**Testing notes**
[Tester to complete]

Dev insight: ideally a validation script should have unit tests which makes sure it works as expected. For manually testing, you should be able to have an invalid README or package.json and see the automated checklist fail when you open a PR.

**Additional context**
Original issue in #64.

rossgaskell:
Will resolve in https://github.com/bbc/psammead/issues/130

PR #245, opened 2018-12-21
Make component documentation more consistent

### LINKED ISSUE #233: Add Usage example to all components
Feedback from R&L team - whilst the components are great, they had to dig into the source code to figure out how to use some of them (e.g. VisuallyHiddenText).

[Some component READMEs contain a Usage section](https://github.com/BBC-News/psammead/tree/latest/packages/components/psammead-sitewide-links), but we should apply this to all the [other component READMEs that don't](https://github.com/BBC-News/psammead/tree/latest/packages/components/psammead-visually-hidden-text).

We should also update our Definition of Done to include a mandatory Usage section (and Installation section).

A PR which adds Usage sections to any components that are missing them, and an update to the definition of done.

**Testing notes**
[Tester to complete]

Dev insight: Documentation only.

bcmn:
Resolves #233 

Adds "usage" section to all components READMEs, & helps ensure documentation is more consistent.

- [ ] Test engineer approval

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,9 +109,9 @@ The `README.md` should follow this template:
 ## Props
 (a table of all the component's props, following the template below)
 
-| Argument  | Type                | Required | Example         |
-|-----------|---------------------|----------|-----------------|
-| prop-name | description of type | Yes/No   | `example input` |
+| Argument  | Type                | Required | Default | Example         |
+|-----------|---------------------|----------|---------|-----------------|
+| prop-name | description of type | Yes/No   | `value` | `example input` |
 
 ## Usage
 (A code block containing an example usage of the component, including importing)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,10 +128,12 @@ We welcome feedback and help on this work. By participating in this project, you
 Psammead is [Apache 2.0 licensed](https://github.com/BBC-News/psammead/blob/latest/LICENSE).
 ```
 
+When linking to other pages from a component's README, we advise you do so only by absolute URL, and avoid the use of relative paths. Component READMEs are displayed in a variety of places, including NPM and Storybook, and this helps ensure the intended meaning is preserved.
+
 If you wish to exclude one or more sections of this template, we advise you do so by simply commenting out the heading, like so:
 
 ```
-<!--# Roadmap-->
+<!-- ## Roadmap -->
 ```
 
 ### After merge
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,7 +114,7 @@ The `README.md` should follow this template:
 | prop-name | description of type | Yes/No   | `example input` |
 
 ## Usage
-(An code block containing an example usage of the component, including importing)
+(A code block containing an example usage of the component, including importing)
 
 ## Accessibility notes
 (ensuring they make sense to non-technical audiences)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,24 +79,60 @@ The changelog should follow the following convention:
 
 **README**
 
-The `README.md` should cover the following territory:
+The `README.md` should follow this template:
 
 ```
+# (package name) &middot; (badges)
+
 ## Description
 (high-level summary that will be accessible to product/UX/Test, plus link to Storybook, explain the different cases)
+
 ## When to use this component
 (including where it is currently used)
-## When to not use this component
+
+## When not to use this component
 (if this is not known, can leave the section empty)
+
+## Installation
+`npm install (package-name)`
+
+## Props
+(a table of all the component's props, following the template below)
+
+| Argument  | Type                | Required | Example         |
+|-----------|---------------------|----------|-----------------|
+| prop-name | description of type | Yes/No   | `example input` |
+
+## Usage
+(An code block containing an example usage of the component, including importing)
+
 ## Accessibility notes
 (ensuring they make sense to non-technical audiences)
+
 ## Roadmap
 (what we have planned for this component)
+
 ## Additional notes
 (link to any relevant ADRs)
+
+## Contributing
+
+Psammead is completely open source. We are grateful for any contributions, whether they be new components, bug fixes or general improvements. Please see our primary contributing guide which can be found at [the root of the Psammead respository](https://github.com/BBC-News/psammead/blob/latest/CONTRIBUTING.md).
+
+### [Code of Conduct](https://github.com/BBC-News/psammead/blob/latest/CODE_OF_CONDUCT.md)
+
+We welcome feedback and help on this work. By participating in this project, you agree to abide by the [code of conduct](https://github.com/BBC-News/psammead/blob/latest/CODE_OF_CONDUCT.md). Please take a moment to read it.
+
+### License
+
+Psammead is [Apache 2.0 licensed](https://github.com/BBC-News/psammead/blob/latest/LICENSE).
 ```
 
-In addition, the `README.md` should also point to Psammead's [CODE_OF_CONDUCT](https://github.com/BBC-News/psammead/blob/latest/CODE_OF_CONDUCT.md), [CONTRIBUTING](https://github.com/BBC-News/psammead/blob/latest/CONTRIBUTING.md) guide, and [LICENSE](https://github.com/BBC-News/psammead/blob/latest/LICENSE).
+If you wish to exclude one or more sections of this template, we advise you do so by simply commenting out the heading, like so:
+
+```
+<!--# Roadmap-->
+```
 
 ### After merge
 
--- a/packages/components/psammead-brand/README.md
+++ b/packages/components/psammead-brand/README.md
@@ -12,7 +12,7 @@ The `Brand` component provides the BBC News logo (as SVG), nested inside a style
 
 | Argument  | Type                | Required | Example         |
 |-----------|---------------------|----------|-----------------|
-| No props. |
+| No props. |                     |          |                 |
 
 ## Usage
 
--- a/packages/components/psammead-brand/README.md
+++ b/packages/components/psammead-brand/README.md
@@ -10,9 +10,9 @@ The `Brand` component provides the BBC News logo (as SVG), nested inside a style
 
 ## Props
 
-| Argument  | Type                | Required | Example         |
-|-----------|---------------------|----------|-----------------|
-| No props. |                     |          |                 |
+| Argument  | Type                | Required | Default | Example         |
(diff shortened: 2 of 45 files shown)


ChrisBAshton:
Lots of conflicts!

bcmn:
Fixed (for now, I'm sure!). 🙂

CONTRIBUTING.md
```diff
@@ -89,24 +89,62 @@ The changelog should follow the following convention:
...
+
+## Installation
+`npm install (package-name)`
+
+## Props
+(a table of all the component's props, following the template below)
+
+| Argument  | Type                | Required | Example         |
+|-----------|---------------------|----------|-----------------|
+| prop-name | description of type | Yes/No   | `example input` |
```
ChrisBAshton: Should we have a `Default` column?

And a `Description`?

Otherwise it's just:

```
| foo | boolean | Yes  | true |
```

...and would need some context! 😄 

So I'd suggest:

```
| Argument  | Type                      | Required | Default | Description                                |
|------------|------------------|----------|---------|-------------------------------|
| foo             | boolean                 | No           | false     | true if you want to bar the baz |
```

(Probably don't need Example _and_ Description, come to think of it)

CONTRIBUTING.md
```diff
@@ -89,24 +89,62 @@ The changelog should follow the following convention:
...
+
+## Usage
+(An code block containing an example usage of the component, including importing)
```
ChrisBAshton:
```suggestion
(A code block containing an example usage of the component, including importing)
```

CONTRIBUTING.md
```diff
@@ -89,24 +89,62 @@ The changelog should follow the following convention:
...
+
+We welcome feedback and help on this work. By participating in this project, you agree to abide by the [code of conduct](https://github.com/BBC-News/psammead/blob/latest/CODE_OF_CONDUCT.md). Please take a moment to read it.
+
+### License
+
+Psammead is [Apache 2.0 licensed](https://github.com/BBC-News/psammead/blob/latest/LICENSE).
 ```
 
-In addition, the `README.md` should also point to Psammead's [CODE_OF_CONDUCT](https://github.com/BBC-News/psammead/blob/latest/CODE_OF_CONDUCT.md), [CONTRIBUTING](https://github.com/BBC-News/psammead/blob/latest/CONTRIBUTING.md) guide, and [LICENSE](https://github.com/BBC-News/psammead/blob/latest/LICENSE).
+When linking to other pages from a component's README, we advise you do so only by absolute URL, and avoid the use of relative paths. Component READMEs are displayed in a variety of places, including NPM and Storybook, and this helps ensure the intended meaning is preserved.
```
ChrisBAshton, line 153: Good shout 👍

CONTRIBUTING.md
```diff
@@ -89,24 +89,62 @@ The changelog should follow the following convention:
...
+
+If you wish to exclude one or more sections of this template, we advise you do so by simply commenting out the heading, like so:
+
+```
+<!-- ## Roadmap -->
+```
```
ChrisBAshton, line 159: Why? Feels a bit unclean 🤔

packages/components/psammead-brand/README.md
```diff
@@ -1,10 +1,24 @@
...
+
+## Installation
+
+`npm install @bbc/psammead-brand`
+
+## Props
+
+| Argument  | Type                | Required | Example         |
+|-----------|---------------------|----------|-----------------|
+| No props. |
```
ChrisBAshton: Invalid table structure, I think?

packages/components/psammead-caption/README.md
```diff
@@ -1,36 +1,53 @@
...
+## Props
+
+| Argument  | Type                | Required | Example         |
+|-----------|---------------------|----------|-----------------|
+| No props. |
+
 ## Usage
 
-`psammead-caption` can be used together with `psammead-visually-hidden-text` in this way. This visually hidden text is to give context to users of Assistive Technology.
+`psammead-caption` can be used together with the [`psammead-figure`](https://github.com/BBC-News/psammead/tree/latest/packages/components/psammead-figure) and [`psammead-visually-hidden-text`](https://github.com/BBC-News/psammead/tree/latest/packages/components/psammead-visually-hidden-text) in this way. This visually hidden text is to give context to users of Assistive Technology.
```
ChrisBAshton:
```suggestion
`psammead-caption` can be used together with the [`psammead-figure`](https://github.com/BBC-News/psammead/tree/latest/packages/components/psammead-figure) and [`psammead-visually-hidden-text`](https://github.com/BBC-News/psammead/tree/latest/packages/components/psammead-visually-hidden-text) components in this way. This visually hidden text is to give context to users of Assistive Technology.
```

packages/components/psammead-caption/README.md
```diff
@@ -1,36 +1,53 @@
...
 const Wrapper = (captionText) => (
-  <figure>
+  <Figure>
     ...
     <Caption>
       <VisuallyHiddenText>Image caption, </VisuallyHiddenText>
       {captionText}
     </Caption>
-  <figure>
+  <Figure>
```
ChrisBAshton:
```suggestion
  </Figure>
```

packages/components/psammead-copyright/README.md
```diff
@@ -1,19 +1,52 @@
...
+
+## Installation
+
+`npm install @bbc/psammead-copyright`
+
+## Props
+
+| Argument  | Type                | Required | Example         |
+|-----------|---------------------|----------|-----------------|
+| No props. |
```
ChrisBAshton: Table structure

Actually, wouldn't we just delete the table altogether and just have a sentence "No props"?

packages/components/psammead-headings/README.md
```diff
@@ -4,14 +4,43 @@
...
+| Argument  | Type                | Required | Example         |
+|-----------|---------------------|----------|-----------------|
+| No props. |
+
+## Usage
+
+```jsx
+import { Headline, SubHeading } from '@bbc/psammead-headings';
+
+const Wrapper = (src, alt, width, height) => (
```
ChrisBAshton: Weird props! ^

packages/components/psammead-headings/README.md
```diff
@@ -4,14 +4,43 @@
...
+    <Fragment>
+        <Heading>Some headline</Heading>
+        <SubHeading>Some subheadline</SubHeading>
+    </Fragment>
+);
+```
+
+All `SubHeading` components can be used as page anchors, with their ID being generated from their text, with any whitespace replaced with hyphens.
```
ChrisBAshton: Can you give an example? I wouldn't have known these were linkable otherwise! Thanks for adding the sentence 👍

packages/components/psammead-image/README.md
```diff
@@ -1,8 +1,22 @@
...
 The `psammead-image` component is a styled `img` element.
 
+## Installation
+
+## Props
+
+| Argument  | Type          | Required | Example                      |
+|-----------|---------------|----------|------------------------------|
+| alt       | String        | Yes      | "A description of the image" |
+| height    | Number/String | No*       | 400                          |
```
ChrisBAshton: Dodgy indentation here

packages/components/psammead-image/README.md
```diff
@@ -13,21 +27,41 @@ const Wrapper = (src, alt, width, height) => (
...
+  ratio,
+  src,
+  width,
+}) => (
+  <Figure>
+    <ImagePlaceholder ratio={ratio}>
+      <Image alt={alt} src={src} width={width} />
+    </ImagePlaceholder>
+  </Figure>
+);
```
ChrisBAshton: It seems redundant to be pulling in Figure/ImagePlaceholder as part of the example usage here. We can just have an example of the Image in isolation.

This is better practice because otherwise if/when the API changes we have to update example code in multiple places.

(Same goes for the ImagePlaceholder readme, we can just have `<Image ...props />` in there or something, the focus is on the ImagePlaceholder.

packages/components/psammead-visually-hidden-text/README.md
```diff
@@ -1,9 +1,31 @@
...
+
+```jsx
+import VisuallyHiddenText from '@bbc/psammead-visually-hidden-text';
+
+const WrappingComponent = () => (
+  <VisuallyHiddenText>Some offscreen text</VisuallyHiddenText>
+);
+```
+
+Please bear in mind that if CSS is disabled, any `children` will be shown inline. Testing your usage of the component should include disabling page styling.
```
ChrisBAshton, line 27: Nice 👍

ChrisBAshton (changes requested):
Good start, thanks Ben!

packages/components/psammead-brand/README.md
bcmn: screenshot 2019-01-03 at 10 50 51

As shown above, Github is fine with it. Feels a bit weird though, I'll fix it.

(Also: you can preview .md files in the Changes view now, which is handy for stuff like this.)

packages/components/psammead-headings/README.md
bcmn:
```suggestion
const Wrapper = () => (
```

CONTRIBUTING.md
bcmn: I wanted to try & encourage consistency. One of the most irritating parts of doing this work was the fact that "adding a usage section to every README" was muddied with the fact that not every "usage" section was called "usage". It feels we should have a clear & consistent structure for READMEs, as this will make it more reliable for users & for linking to specific subheadings, & that structure should remain consistent even as parts of the READMEs change.

This was kinda my logic for leaving the template of the table even under "No props" scenarios, as we already had 3 variations of table columns, as far as I remember.

ChrisBAshton: Makes it more automatically testable too 👍

bcmn: Oh yeah, you jogged my memory -- I think I'd briefly looked into #147 at the time, & considered that this would allow us to assert than every package still contains the string `## Roadmap` without requiring some READMEs have a section header with an empty body.

CONTRIBUTING.md
bcmn: I've added a `Default` column. I agree somewhat on the `Description`, but updating these across all READMEs has already proven a _very_ time-consuming task & I don't feel it's worth the investment -- if the purpose isn't clear from the combination of a name, type (which in some cases already has a little more info), default & example, then it's a strong sign we should be re-evaluating the props & their identifiers in the first place.

radiocontrolled (approved):

dr3 (approved):

ChrisBAshton (approved):

jamesbrumpton:
LGTM 👍

Issue #156, opened 2018-12-07
Improve test coverage for our utilities packages


dr3:
**Is your feature request related to a problem? Please describe.**
We currently have a very red test coverage output. This is mainly due to out utilities packages

**Describe the solution you'd like**
We add tests, maybe testing the number of exports, how they're exported etc 

**Describe alternatives you've considered**
We turn off code coverage for utilities.

![image](https://user-images.githubusercontent.com/11341355/49654868-e14b9000-fa30-11e8-9583-0f6ffe35b236.png)

- [x] Initially labelled with ["Refinement needed"](https://github.com/BBC-News/psammead/labels/Refinement%20Needed)

ChrisBAshton:
Fixed in #259

PR #258, opened 2019-01-04
Add and use method to test a package's exported values

pjlee11:
Part 1 of 2 for #156

Follows up on https://github.com/BBC-News/psammead/pull/224

Adds a test helper method that takes an actual import and validates it has all the exported values it is expected to have. This can then be used to simplifying testing the other utility packages exports as well as `psammead-test-helpers`

Has error handling logic for both new files (EG: font.js) and for exported files within those files when:
* The actual import has additional exports when compared to the expected values. This is needed so that we do not forget to change the expectations when adding a new export.
* The expected value has additional exports as compared to the actual exports. This is needed in the case of a breaking change when an export is to be removed. This is very useful at catching regressions in the utility packages.

- [ ] Test engineer approval

Testing this involves ensuring no regression in the `psammead-test-helpers` package, which can be verified via running `npm run build & npm run test`. Also the `psammead-test-helpers` package test coverage should increase from `0%` to `51.28%`


packages/utilities/psammead-test-helpers/README.md
```diff
@@ -34,6 +35,8 @@ There is currently a plan to migrate from `react-test-renderer` to `enzyme` as i
 
 ## Contributing
 
+When **adding** a new export to this utility package the [export tests](https://github.com/BBC-News/psammead/blob/5d7395fd60bd8d73796d5a23775b4b5b36db1445/packages/utilities/psammead-test-helpers/index.test.jsx#L8-L14) also need to be updated and the export should be adding to the [README](https://github.com/BBC-News/psammead/tree/5d7395fd60bd8d73796d5a23775b4b5b36db1445/packages/utilities/psammead-test-helpers#exported-functions). When **removing** an exisiting export from this utility package the [export tests](https://github.com/BBC-News/psammead/blob/5d7395fd60bd8d73796d5a23775b4b5b36db1445/packages/utilities/psammead-test-helpers/index.test.jsx#L8-L14) need to be updated, the export should be removed from the [README](https://github.com/BBC-News/psammead/tree/5d7395fd60bd8d73796d5a23775b4b5b36db1445/packages/utilities/psammead-test-helpers#exported-functions) and the package version requires a major change (EG: 1.2.1 -> 2.0.0) as this would be considered a breaking change due to functionality being removed.
```
ChrisBAshton, line 38: This got changed back from when I last suggested a change?

bcmn (dismissed):

packages/utilities/psammead-test-helpers/README.md
pjlee11:
```suggestion
When **adding** a new export to this utility package, the export must be added to the [export tests](https://github.com/BBC-News/psammead/blob/5d7395fd60bd8d73796d5a23775b4b5b36db1445/packages/utilities/psammead-test-helpers/index.test.jsx#L8-L14). When **removing** an existing export from this utility package, the export will need removing from the export tests and the package version requires a major change (EG: 1.2.1 -> 2.0.0). In all cases, the README requires updating too.
```

bcmn (dismissed):

ChrisBAshton (dismissed):

pjlee11:
snyk stalling on jobs so re-committing ...

ChrisBAshton (approved):

pjlee11:
Confirmed that synk is hanging as the webhooks are looking at the wrong github repo

bcmn (approved):

jamesbrumpton:
LGTM 👍

PR #319, opened 2019-03-04
Fix dependancies

dr3:
**Overall change:** The package locks are in a weird state which I cant replicate

**Code changes:**

- Ran `npx lerna exec --concurrency 1 -- 'rm -rf node_modules && rm package-lock.json && npm install'`
- Ran `npm run updateSnapshots`
- Added webpack for storybook to fix styling issue with no hoisting
- Bumped and changelog'd about 4 trillion packages

**Code review**
Please ensure you can run these in **any** order
1. `npm run ci:packages && npm run test` (all pass)
2. `npm run ci:packages && npm run storybook` (styles load)
2. `npm run install:packages && npm run storybook` (styles load)
4. `npm run install:packages && npm run test` (all pass)
5. `npm run install:packages:link && npm run test` (tests fail)

ezgif-4-f7ede509e320

---

- [ ] Test engineer approval


pjlee11:
@dr3 could you elaborate on why we expect the tests to fail on `npm run install:packages:link && npm run test`? Is it because the snapshot output of the linked packages is different to the package in isolation? 

NB: The following screenshot is the failure for this on my local and I'm a little confused with the swap from `em` to `rem` ...

image

sareh (approved):
👍 Thanks for making these changes. It all looks good for me locally. 
To test the linking with storybook was working well, I also made an update to a utility package and used it within a component package, then used `npm run install:packages:link && npm run storybook` to see the changes reflected.  
Great work!

dr3:
> @dr3 could you elaborate on why we expect the tests to fail on `npm run install:packages:link && npm run test`? Is it because the snapshot output of the linked packages is different to the package in isolation?

@pjlee11 This is because our tests need to be in 1 of two states, either whats on NPM or whats local. As packages need to be published, to be pulled in to another package, they arnt the same.

We picked to set our snapshots to be whats published, rather than whats local when linked

pjlee11 (approved):

jamesbrumpton:
👍

PR #335, opened 2019-03-07
RTL testing

### LINKED ISSUE #328: Investigate how we add RTL/LTR Support to components
We need to be able to test how components react to LTR/RTL text direction.

We want to install Storybook Addon RTL to toggle the text direction per component.

N/A

**Testing notes**
N/A

N/A

**Additional context**
https://github.com/unindented/storybook-addon-rtl

j-pendlebury:
Resolves #328

**Overall change:**
We (the WS team) need to make sure that the current components will work with RTL left scripts. This is the beginning of such work.

**Code changes:**

- Added the `direction` CSS property to the component. The value of "RTL" can be passed via a prop, or if not, it will default to "LTR".
- A new story was added to reflect this change and to show an example of a RTL script. The caption was taken from an existing article on BBC Persian.
- Updated snapshots accordingly

---

- [ ] Test engineer approval

### DIFF
--- a/packages/components/psammead-caption/CHANGELOG.md
+++ b/packages/components/psammead-caption/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 | Version | Description |
 |---------|-------------|
+| 0.3.0   | [PR#335](https://github.com/BBC/psammead/pull/335) Add `direction` CSS property for RTL scripts |
 | 0.2.0   | [PR#318](https://github.com/BBC/psammead/pull/318) Update to new font face and family. |
 | 0.1.6   | [PR#323](https://github.com/BBC/psammead/pull/323) Update readme storybook badge |
 | 0.1.5   | [PR#264](https://github.com/BBC/psammead/pull/319) Resolving package-lock issues. |
--- a/packages/components/psammead-caption/package-lock.json
+++ b/packages/components/psammead-caption/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-caption",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/packages/components/psammead-caption/package.json
+++ b/packages/components/psammead-caption/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-caption",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "main": "dist/index.js",
   "description": "React styled components for a Caption",
   "repository": {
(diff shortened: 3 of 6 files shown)


dr3:
Just being nosey 🙈 Direction is usually done at the page level once, rather than the component level, perhaps we could try something like https://www.npmjs.com/package/storybook-addon-rtl? Saves each component being polluted by prop :)

j-pendlebury:
@dr3 you're right, and I suggested doing it that way (even had a PR almost ready with the add-on config'ed), but was told to do it this way. We're still in a investigation phase so I guess we're looking at the best way of doing things. Thanks for your input though 😄

chris-hinds:
I had suggested we look at passing a prop to the component for textDirection incase there was any other layout work we needed to do with the individual components and this way you would easily be able to test each component in isolation and see how they handle rtl.

But thinking about it I don't really think this is a genuine use case right now, and we can probably handle this by just setting  `dir="rtl"` in the parent as Jamie and Drew suggested. So ignore my initial thoughts :P

chris-hinds:
One other thing to note is that the storybook addon places the `dir` attribute on the body tag however W3C states that this should be placed on the HTML tag. Just something to note when we actually implement this.

image

https://www.w3.org/International/questions/qa-html-dir#quickanswer

dr3:
Sounds good @hindsc52, should custom direction logic be needed in future we can discuss possible wrappers or _direction_ context passing at that point :P 

On the `body` vs `head` comment, I imagine it's fine for our storybook use case, but we could push for https://github.com/unindented/storybook-addon-rtl/issues/2 to be addressed/fork if its not?

chris-hinds:
> Sounds good @hindsc52, should custom direction logic be needed in future we can discuss possible wrappers or _direction_ context passing at that point :P
> 
> On the `body` vs `head` comment, I imagine it's fine for our storybook use case, but we could push for [unindented/storybook-addon-rtl#2](https://github.com/unindented/storybook-addon-rtl/issues/2) to be addressed/fork if it's not?

@dr3 agree! And yes for Storybook its fine, just a good reminder to make sure we place the attribute in the right place in the applications :)

And yes, we can comment again on that issue asking for it to be implemented.

dr3:
> One other thing to note is that the storybook addon places the `dir` attribute on the body tag however W3C states that this should be placed on the HTML tag. Just something to note when we actually implement this.

This has now been resolved by https://github.com/unindented/storybook-addon-rtl/pull/5 and published as 0.2.0 :)

Also ~0.3.0 will be coming out soon to support SB5~ v0.2.2 now supports SB5 🎉

j-pendlebury:
Closing this PR as we're going with #336 instead 🎉

PR #353, opened 2019-03-13
Add support for different scripts typographies

### LINKED ISSUE #342: Extend support for scripts in gel-foundation
typography.js definitions do not comprise World Service scripts

Make the types into functions that take the script as a parameter

We have considered having an all-knowing theme context object that would be called on by components, but that was a terrible idea.

**Testing notes**
[Tester to complete]

**Additional context**

- [ ] Initially labelled with ["Refinement needed"](https://github.com/bbc/psammead/labels/Refinement%20Needed)

j-pendlebury:
Resolves #342

**Overall change:** Add support to allow components to use Typography type sizes for different scripts.

**Code changes:**

- Create files with the GEL Types Styles as defined by UX for each script (arabic, burmese, cjk_thai, hindi_nepali_bengali, latin, mundo, sinhala, and tamil).
- Add in the `latin` script the default values GEL for all the Types Styles. 
- Turn the Types (e.g. GEL Trafalgar) into functions (e.g. `getTrafalgar`), which takes the script name as an argument and looks for the type sizes related to that script and returns the relevant font-size and line-size.
- Load script files dynamically.

---

Testers are not expected to review this until the changes to the components are merged.

---

- [ ] Test engineer approval


j-pendlebury:
@briankabiro and I worked on simplifying the import of different typography settings. We created a `typographyConfig` helper within the scripts directory for easier import. It basically imports all the typography settings and then in `typography.js` we pick which one is going to be used.

Note: we had to disable linting for `import/no-useless-path-segments` in typographyConfig because after running the `npm run build` command to compile the code, the JSON files are not accessible from the `dist` directory, and rather than having them duplicated, this seemed like an easier solution.

j-pendlebury:
## Before merging make sure to:
- [x] Bump `package.json` version
- [x] Update `package-lock.json` (`npm install`)
- [x] Update CHANGELOG
- [x] Update README
- [x] Update PR description

packages/utilities/gel-foundations/src/scripts/typographyConfig.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable import/no-useless-path-segments */
+import arabicConfig from '../../src/scripts/arabic.json';
+import latinConfig from '../../src/scripts/latin.json';
+import mundoConfig from '../../src/scripts/mundo.json';
+
+const scriptConfig = {
```
dr3: 👋 Hey, just being nosey but It might be worth thinking about how we can tree shake/bundle split this

mgurgel: @dr3 Maybe you've come across this problem before and can give us a hand. At this point, there is a 1-to-1 correspondence between a `script` variable and the filename from which we will be pulling a json object. Would it be the case here to just to an fs.read for the specific file named `../../src/scripts/${script}.json` (considering it's not executable code, but just JSON)?

dr3: Hey :) Yeah I appreciate this is a new concern, we would be happy to go over in more detail, but the jist is that doing `import mundoConfig` etc for all services causes every services config to be bundled when you may only need one.

One possible solution is to get the application to pass the config e.g. `getAtlas(scriptConfig)` and then its down to the application to load only the config it needs

packages/utilities/gel-foundations/src/scripts/latin.json
```diff
@@ -0,0 +1,128 @@
...
+    "groupB": {
+      "fontSize": "24",
+      "lineHeight": "28"
+    },
+    "groupD": {
+      "fontSize": "32",
+      "lineHeight": "36"
+    }
+  },
+  "paragon": {
```
alex-magana: Consider removing the empty definitions of paragon, doublePica and greatPrimer.

packages/utilities/gel-foundations/src/scripts/typographyConfig.js
alex-magana: We've looked into this and implemented the possible solution below.
`require('./scripts/'+ script + '.json');`
We would add the code above to `typography.js`, in method, `getFont`.
Additionally, with this, we will need
- to disable the `import/no-dynamic-require` rule, and
- add `--copy-files` to script, `build` in the top-level `package.json`

packages/utilities/gel-foundations/src/typography.js
```diff
@@ -1,11 +1,78 @@
-import { MEDIA_QUERY_TYPOGRAPHY } from './breakpoints';
+import { MEDIA_QUERY_TYPOGRAPHY } from '../breakpoints';
 
 const fontFamilyBase = ', Helvetica, Arial, sans-serif';
 
 export const GEL_FF_REITH_SERIF = `ReithSerif${fontFamilyBase}`;
 export const GEL_FF_REITH_SANS = `ReithSans${fontFamilyBase}`;
 export const GEL_FF_REITH_SANS_COND = `ReithSansCondensed${fontFamilyBase}`;
 
+function getTypeSizes(type = 'default', script = 'latin') {
```
alex-magana: Consider setting the default type to an existing value in the script `.json` files or introducing a guard clause on `type`.
I raise this because I haven't encountered a type, `default`, in the script `.json` files. As such, the function might end up setting `font-size` and `line-height` to `undefined`.

packages/utilities/gel-foundations/src/typography.js
```diff
@@ -1,11 +1,78 @@
...
+  // eslint-disable-next-line global-require, import/no-dynamic-require
+  const importedScript = require(`./scripts/${script}.json`);
+  const typographyType = importedScript[type];
+
+  const GROUP_A_FONT_SIZE_PX = typographyType.groupA.fontSize;
```
sareh: What is the reason for having pixels as exports instead of just `rem`?

packages/utilities/gel-foundations/src/typography.js
DenisHdz: Hey @alex-magana, take a look to the last commit. We are actually going to be always passing the typography style name to the `getTypeSizes` function so there is no need to set a default one.

dr3 (changes requested):
**COPIED FROM SLACK** 

drew [9:52 AM]
```const importedScript = require(`./scripts/${script}.json`);```
wont work for us with webpack sadly
A sollution for us would be to export the config out of the package, and allow the user to pass the config in when its invoked
So the function itself wouldnt contain any config

marcos [9:53 AM]
Do you have an example of this being done elsewhere in Psammead or Simorgh?
Is this the first time we’ve needed to load JSON in either of these repos?

drew [9:56 AM]
In service config
```import { sinhalaTypographyConfig } from '@bbc/gel-foundations/typography/scripts';```

In container
```
import { getTrafalgar } from '@bbc/gel-foundations/typography

>Pass sinhalaTypographyConfig from Service Context

getTrafalgar(sinhalaTypographyConfig)
```

Instead json we should use js files that export objects too
This way, the application can load just `sinhalaTypographyConfig` and use the function, without loading any other scripts config

**Comment by DenisHdz** [comment]
> **COPIED FROM SLACK**
> 
> drew [9:52 AM]
> `` const importedScript = require(`./scripts/${script}.json`); ``
> wont work for us with webpack sadly
> A sollution for us would be to export the config out of the package, and allow the user to pass the config in when its invoked
> So the function itself wouldnt contain any config
> 
> marcos [9:53 AM]
> Do you have an example of this being done elsewhere in Psammead or Simorgh?
> Is this the first time we’ve needed to load JSON in either of these repos?
> 
> drew [9:56 AM]
> In service config
> `import { sinhalaTypographyConfig } from '@bbc/gel-foundations/typography/scripts';`
> 
> In container
> 
> ```
> import { getTrafalgar } from '@bbc/gel-foundations/typography
> 
> >Pass sinhalaTypographyConfig from Service Context
> 
> getTrafalgar(sinhalaTypographyConfig)
> ```
> Instead json we should use js files that export objects too
> This way, the application can load just `sinhalaTypographyConfig` and use the function, without loading any other scripts config

I have renamed the script files from JSON to JS and exported the Typography Types as objects as suggested. Also, I removed the dynamic require from the `getTypeSizes` function because we will be passing the script object directly as an argument.

File: packages/utilities/gel-foundations/src/typography.js [hunk 5]
```diff
@@ -1,11 +1,76 @@
...
+
+export const getImperial = script => getTypeSizes('imperial', script);
+
+export const getRoyal = script => getTypeSizes('royal', script);
+
+export const getFoolscap = script => getTypeSizes('foolscap', script);
+
+export const getBodyCopy = script => getTypeSizes('bodyCopy', script);
+
+/* ORIGINAL CONSTANTS */
```
sareh, line 40: What is the reason for making the GEL standard line heights and font sizes saved under 'news'? Can these not be under 'default', since these are what the gel-foundations typography settings should be if they don't need to be overridden.

Why is there a distinction between these typography settings and those in `latin.js`? I'd imagine they should be identical.


The reason our constant values for GEL typography are using `rem` instead of `px` values is for accessibility reasons - when we zoom into a page with `CTRL +` or `CMD +` or use a browser's zoom feature, using rem will ensure that the text scales well in conjuction with the rest of the page. 
[When we made this change, we wrote up a table of our type sizes here. ](https://github.com/bbc/simorgh/blob/67decf15656f785ae20caf0cef7e9e5657920766/src/app/lib/README.md#typography) 

It would be good to have all the scripts consistently using `rem`.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Core Components Library Developed & Maintained By The Articles and Reach & Languages Team",
   "main": "index.js",
   "scripts": {
-    "build": "lerna exec --parallel -- babel src -d dist --ignore '**/*.test.jsx,**/*.stories.jsx' --root-mode upward",
+    "build": "lerna exec --parallel -- babel src -d dist --ignore '**/*.test.jsx,**/*.stories.jsx' --root-mode upward --copy-files",
```
dr3: This change shouldnt be needed if you use js files not json files

packages/utilities/gel-foundations/src/typography.js
DenisHdz: Because as the UX team have defined them using pixels, we thought it is easier and more clear to declare them using px if anyone needs to add, modify or correct any type in the script files, and convert them to rem within the `getTypeSizes` function.

packages/utilities/gel-foundations/src/typography.js
DenisHdz: 1. Do you mean why we named it `news.js` instead of `default.js`? 

If not, please note that after Drew's yesterday suggestion about removing the dynamic script import in the `getTypeSizes` function because it won't work with webpack, now we will need to pass directly the script object with all the GEL Type Styles as an argument.

The script object will be imported from the Service Context and passed to the component via props, and as the `script` prop will be required we will need to always import either a NEWS or WS script.

Please find below an implementation example using the Caption component.

**Import**
`import { arabic } from '@bbc/gel-foundations/scripts;`
or
`import { news } from '@bbc/gel-foundations/scripts';`

**Usage**
`<Caption script={arabicTypographyConfig}>This is a caption.</Caption>,`

**Component declaration**
Screenshot 2019-03-19 at 09 52 53



If you have an altenative solution, could you please provide an example?

2. There is a distinction between `latin.js` and `news.js` because we assumed it would make it easier for UX to change something in WS and not in NEWS, and at the moment we don't have `Body Copy` in WS and we are using its sizes as `Pica`.

3. We are still using `rem`, then only difference is that we declare the GEL Types Styles using pixels in the scrips and then we convert them to `rem`, as explained in the previous comment.

packages/utilities/gel-foundations/README.md
```diff
@@ -31,18 +31,20 @@ import { css } from 'styled-components';
...
 
-Our typography uses `em` for font-size and `rem` for line-height. `em` allows modularity of components: you can change the component font-size by changing the font-size of its container. `rem` is relative to the document root, so we use that for line-height and spacing for a consistent look-and-feel across the document. You can read our [detailed analysis of "REMs vs EMs for spacing"](https://github.com/bbc/simorgh/blob/latest/docs/Spacing-Units.md) for more information.
+Our typography uses `rem` for font-size and line-height. `rem` is relative to the document root, so we use that for font-size and line-height and spacing for a consistent look-and-feel across the document. You can read our [detailed analysis of "REMs vs EMs for spacing"](https://github.com/bbc/simorgh/blob/latest/docs/Spacing-Units.md) for more information.
 
+You can pass a script name as a prop to the utility typography function (e.g. `getLongPrimer('arabic')`), to get the font-size and line-height for the Typography Style of that script. By default it will use the `news` script as found [here](https://www.bbc.co.uk/gel/guidelines/typography#type-sizes).
```
dr3: I think this needs updating to the new method of passing script config

packages/utilities/gel-foundations/src/scripts/arabic.js
```diff
@@ -0,0 +1,88 @@
+const arabicTypographyConfig = {
```
dr3: These files currently dont have a nice import path, You would have to do 
```
import arabicTypography from '@bbc/gel-foundations/dist/scripts/arabic'
```

packages/utilities/gel-foundations/README.md
DenisHdz: Thanks for spotting this @dr3. I have updated it.

packages/utilities/gel-foundations/src/scripts/arabic.js
dr3: Hey @DenisBBC im not sure https://github.com/bbc/psammead/pull/353/commits/c75ff29436c8fe161606e881b4ad35b44ca1b7a9 changes anything You still have to reference in the dist folder `@bbc/gel-foundations/dist/scripts/arabic`

dr3: An example of treeshaking we want is https://survivejs.com/webpack/optimizing/tree-shaking/#demonstrating-tree-shaking

I suggest putting a file at the base of this package called `scripts.js` and inside have

```
module.exports = {
    enGb: require('./dist/scripts/enGB'),
    arabic: require('./dist/scripts/arabic'),
    ...etc
}
```

That allows the user to do 

`import { arabic } from '@bbc/gel-foundations/scripts'`

And tree shake all other typographies

packages/utilities/gel-foundations/src/typography.js
```diff
@@ -1,11 +1,76 @@
-import { MEDIA_QUERY_TYPOGRAPHY } from './breakpoints';
+import { MEDIA_QUERY_TYPOGRAPHY } from '../breakpoints';
 
 const fontFamilyBase = ', Helvetica, Arial, sans-serif';
 
 export const GEL_FF_REITH_SERIF = `ReithSerif${fontFamilyBase}`;
 export const GEL_FF_REITH_SANS = `ReithSans${fontFamilyBase}`;
 export const GEL_FF_REITH_SANS_COND = `ReithSansCondensed${fontFamilyBase}`;
 
+function getTypeSizes(typogStyleName, importedScript) {
```
dr3:
```suggestion
const getTypeSizes = (typogStyleName, importedScript) => {
```

dr3: Also, could this function be moved into its own file? and unit tested

packages/utilities/gel-foundations/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 | Version | Description |
 |---------|-------------|
+| 0.2.6   | [PR#353](https://github.com/BBC/psammead/pull/353) Add support for different scripts typographies |
```
ChrisBAshton:
```suggestion
| 0.3.0  | [PR#353](https://github.com/BBC/psammead/pull/353) Add support for different scripts typographies |
```

I'd suggest given the size and scope of this PR it's more than a patch release.

packages/utilities/gel-foundations/README.md
```diff
@@ -31,17 +31,19 @@ import { css } from 'styled-components';
 
 const SomeStyledComponent = css`
   font-family: ${GEL_FF_REITH_SANS};
+  ${getLongPrimer(script)};
```
dr3, line 34: Can you please add an example of where 'script' comes from

ChrisBAshton, line 34: What is `script` in this snippet? Would be nice to have a readily copy-and-pastable example.

packages/utilities/gel-foundations/README.md
```diff
@@ -31,17 +31,19 @@ import { css } from 'styled-components';
...
 
-Our typography uses `em` for font-size and `rem` for line-height. `em` allows modularity of components: you can change the component font-size by changing the font-size of its container. `rem` is relative to the document root, so we use that for line-height and spacing for a consistent look-and-feel across the document. You can read our [detailed analysis of "REMs vs EMs for spacing"](https://github.com/bbc/simorgh/blob/latest/docs/Spacing-Units.md) for more information.
+Our typography uses `rem` for font-size and line-height. `rem` is relative to the document root, so we use that for font-size and line-height and spacing for a consistent look-and-feel across the document. You can read our [detailed analysis of "REMs vs EMs for spacing"](https://github.com/bbc/simorgh/blob/latest/docs/Spacing-Units.md) for more information.
+
+You can pass a script config object to the utility typography function (e.g. `getLongPrimer('arabicTypography')`), to get the Typography Type Styles (font-size and line-height) specifics for that script. By default it will use the `news` script as found [here](https://www.bbc.co.uk/gel/guidelines/typography#type-sizes).
```
ChrisBAshton:
```suggestion
You can pass a script config object to the utility typography function (e.g. `getLongPrimer('arabicTypography')`), to get the Typography Type Styles (`font-size` and `line-height`) specifics for that script. By default it will use the `news` script as found on the [GEL typography guidelines](https://www.bbc.co.uk/gel/guidelines/typography#type-sizes).
```

packages/utilities/gel-foundations/scripts.js
```diff
@@ -0,0 +1,12 @@
...
+module.exports = {
+  news: require('./dist/scripts/news').default,
+  arabic: require('./dist/scripts/arabic').default,
+  burmese: require('./dist/scripts/burmese').default,
+  cjkThai: require('./dist/scripts/cjk_thai').default,
+  hindiNepaliBengali: require('./dist/scripts/hindi_nepali_bengali').default,
+  latin: require('./dist/scripts/arabic').default,
+  mundo: require('./dist/scripts/arabic').default,
+  sinhala: require('./dist/scripts/sinhala').default,
+  tamil: require('./dist/scripts/tamil').default,
```
ChrisBAshton: What's the plan for adding additional services? Is there one?

packages/utilities/gel-foundations/src/scripts/arabic.js
DenisHdz: @dr3 I added Tree shaking as suggested but when importing the script doing `import { arabic } from '@bbc/gel-foundations/scripts` it returns the script object with a default key, so I had to add a `.default` at the end of each require.

packages/utilities/gel-foundations/src/scripts/arabic.js
```diff
@@ -0,0 +1,88 @@
+const arabicTypography = {
```
ChrisBAshton, line 1: I'd suggest the filename is enough to denote the service name, and that this should be pure config (i.e. `arabic.js` -> `arabic.json`, make it a pure JSON file)

packages/utilities/gel-foundations/src/typography.js
```diff
@@ -1,11 +1,76 @@
-import { MEDIA_QUERY_TYPOGRAPHY } from './breakpoints';
+import { MEDIA_QUERY_TYPOGRAPHY } from '../breakpoints';
```
ChrisBAshton: Why has this changed? 🤔

packages/utilities/gel-foundations/src/typography.js
```diff
@@ -1,11 +1,76 @@
...
+        ${MEDIA_QUERY_TYPOGRAPHY.SMART_PHONE_ONLY} {
+          font-size: ${GROUP_B_FONT_SIZE};
+          line-height: ${GROUP_B_LINE_HEIGHT};
+        }
+   
+        ${MEDIA_QUERY_TYPOGRAPHY.LAPTOP_AND_LARGER} {
+          font-size: ${GROUP_D_FONT_SIZE};
+          line-height: ${GROUP_D_LINE_HEIGHT};
+        }
+      `;
```
ChrisBAshton: This function is awesome 🎉 strange indentation in the return statement though?

packages/utilities/gel-foundations/src/typography.js
ChrisBAshton: Agreed, this function (and inherited functions) lends itself well to being in a separate file and unit tested.

ChrisBAshton (commented):
Looks good to me. The explanation of `px` / `rem` makes sense to me too. Please consider Sareh's comments addressed, in her absence.

I have a few comments of my own, some are fixes but some are just suggestions on best practices.

packages/utilities/gel-foundations/README.md
DenisHdz: @dr3 Updated :)

DenisHdz: Hi @ChrisBAshton, I have just updated the README. Let me know if it is enough self explanatory now or I should add an example.

packages/utilities/gel-foundations/src/typography.js
DenisHdz: Moved it into a separate file and added test.

packages/utilities/gel-foundations/src/scripts/arabic.js
DenisHdz: @ChrisBAshton Could we agree on this with the rest of the team, please? Because we defined them initially as JSON files but replaced them to JS after [this](https://github.com/bbc/psammead/pull/353#pullrequestreview-215505449) discussion

ChrisBAshton: Ha, didn't know that! Apologies for the mixed messages. I'm happy leaving as-is and assume others have looked into the implications of it more than me.

packages/utilities/gel-foundations/scripts.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable global-require */
+module.exports = {
+  news: require('./dist/scripts/news').default,
```
sareh:
```suggestion
```
News should not be a separate script. We should be using `latin`. 

Currently, the `news` file has these extra values that are missing in the `latin` file:
`atlas`, `elephant`, `imperial`, `royal`, `foolscap`, `paragon`, `doublePica`, `greatPrimer`, `bodyCopy`. 
Can you please add them?

I saw your comment below ```There is a distinction between latin.js and news.js because we assumed it would make it easier for UX to change something in WS and not in NEWS, and at the moment we don't have Body Copy in WS and we are using its sizes as Pica.```

However, News UX and WS UX has to be unified and not independent, otherwise our shared approach will not work. News is just a special case of a Latin WS site.

packages/utilities/gel-foundations/scripts.js
```diff
@@ -0,0 +1,12 @@
...
+  arabic: require('./dist/scripts/arabic').default,
+  burmese: require('./dist/scripts/burmese').default,
+  cjkThai: require('./dist/scripts/cjk_thai').default,
+  hindiNepaliBengali: require('./dist/scripts/hindi_nepali_bengali').default,
+  latin: require('./dist/scripts/arabic').default,
```
sareh:
```suggestion
  latin: require('./dist/scripts/latin').default,
```

packages/utilities/gel-foundations/scripts.js
```diff
@@ -0,0 +1,12 @@
...
+  mundo: require('./dist/scripts/arabic').default,
```
sareh:
```suggestion
  mundo: require('./dist/scripts/mundo').default,
```

packages/utilities/gel-foundations/scripts.js
```diff
@@ -0,0 +1,12 @@
...
+  sinhala: require('./dist/scripts/sinhala').default,
```
sareh:
```suggestion
  sinhalese: require('./dist/scripts/sinhalese').default,
```
The script is called the Sinhalese. We should update the filenames too to reflect the name.

sareh (dismissed):

packages/utilities/gel-foundations/scripts.js
sareh: Why is `mundo` the name of this script? Should it not be named `latin-with-diacritics`?

packages/utilities/gel-foundations/scripts.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable global-require */
+module.exports = {
+  news: require('./dist/scripts/news').default,
+  arabic: require('./dist/scripts/arabic').default,
+  burmese: require('./dist/scripts/burmese').default,
+  cjkThai: require('./dist/scripts/cjk_thai').default,
+  hindiNepaliBengali: require('./dist/scripts/hindi_nepali_bengali').default,
```
sareh: Since we're defining scripts, not services, we should name these accordingly.

Hindi & Nepali use the `Devanagari` script, so `devanagari` should be the script name here.  (https://en.wikipedia.org/wiki/Devanagari)

Bengali uses the `Bengali-Assamese` script and should be named as a separate file.  https://en.wikipedia.org/wiki/Bengali%E2%80%93Assamese_script

We shouldn't just group different scripts together for convenience, but have them be distinct files.

DenisHdz: @sareh Applied the new scripts group names.

packages/utilities/gel-foundations/scripts.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable global-require */
+module.exports = {
```
sareh: Looks great! Thanks for updating this to reflect the groupings for scripts.

packages/utilities/gel-foundations/index.test.jsx
```diff
@@ -56,6 +56,21 @@ const typographyExpectedExports = {
...
+  getLongPrimer: 'function',
+  getBrevier: 'function',
+  getMinion: 'function',
+  getAtlas: 'function',
+  getElephant: 'function',
+  getImperial: 'function',
+  getRoyal: 'function',
+  getFoolscap: 'function',
+  getBodyCopy: 'function',
 };
```
sareh, line 96: Please add tests here to check we export all of the scripts e.g. `arabic-pashto`, etc.
This will ensure we don't break any script exports in this module as we maintain this package.

packages/utilities/gel-foundations/README.md
```diff
@@ -31,17 +31,20 @@ import { css } from 'styled-components';
...
+Our typography uses `rem` for font-size and line-height. `rem` is relative to the document root, so we use that for font-size and line-height and spacing for a consistent look-and-feel across the document. You can read our [detailed analysis of "REMs vs EMs for spacing"](https://github.com/bbc/simorgh/blob/latest/docs/Spacing-Units.md) for more information.
+
+The utility typography function (e.g. `getLongPrimer('arabic')`) receives a script argument which is an object with the different script GEL Types Sizes and returns the `font-sizes` and `line-heights` for the specific GEL Type (e.g. `Long Premier`).
+You can import the script doing `import { arabic } from '@bbc/gel-foundations/scripts`. By default you should import the `news` script as found on the [GEL typography guidelines](https://www.bbc.co.uk/gel/guidelines/typography#type-sizes).
```
sareh:
```suggestion
You can import the script doing `import { arabic } from '@bbc/gel-foundations/scripts`. By default you should import the `latin` script as found on the [GEL typography guidelines](https://www.bbc.co.uk/gel/guidelines/typography#type-sizes).
```

packages/utilities/gel-foundations/scripts.js
```diff
@@ -0,0 +1,18 @@
...
+  latin: require('./dist/scripts/latin-and-cyrillic').default,
```
sareh: It would be good to briefly explain in the Readme why these scripts have been mapped to the these typography settings. Perhaps in a table.

This is so the Readme can be used by those new to supporting multiple scripts, to understand the choices made. It can also be useful for maintaining this package.

pjlee11: I agree with Sareh here. I might also be good to link the Dropbox analysis doc assuming it is okay to both be public and is clearly described as a working doc.

sareh: @pjlee11 Instead of linking to Dropbox, any relevant analysis could be summarised in a document in this package. This increases the visibility and longevity of the documentation - since often dropbox documents can be moved.

packages/utilities/gel-foundations/README.md
```diff
@@ -43,8 +43,27 @@ To allow the typography to be fully accessible and responsive, please note that
...
+| arabic                      | Urdu, Arabic, Persian | Group as they use the Arabic alphabet | 
+| arabic-pashto               | Pashto | Contains a subset of glyphs that mean it differs from Arabic | 
+| no-ascenders-or-descenders  | Korean, Japanese, Zhongwen, Ukchina, Tigrinya, Amharic | Group all alphabets without ascenders or descenders. This could be implemented as chinese, korean, ethiopic | Hindi, Nepali, Gujarati, Punjabi, Marathi, Telegu | The typography values for Devanagari and Gurmukhi characters are identical |
+| devanagari-and-gurmukhi     |  Hindi, Nepali, Gujarati, Punjabi, Marathi, Telegu | The typography values for Devanagari and Gurmukhi characters are identical |   
+| thai                        | Thai     | Bespoke to the Thai characters       | 
+| bengali                     | Bengali  | Bespoke to the Bengal characters     | 
+| burmese                     | Burmese  | Bespoke to the Burmese characters    | 
+| sinhalese                   | Sinhala  | Bespoke to the Sinhalese characters  | 
+| tamil                       | Tamil    | Bespoke to the Tamil characters      |
```
sareh, line 66: Great. Thanks for adding this. 🌟

packages/utilities/gel-foundations/src/scripts/arabic-pashto.js
```diff
@@ -0,0 +1,144 @@
...
+      lineHeight: '30',
+    },
+    groupD: {
+      fontSize: '22',
+      lineHeight: '34',
+    },
+  },
+  pica: {
+    groupA: {
+      fontSize: '',
```
pjlee11: Is there a reason we are leaving these values empty?

packages/utilities/gel-foundations/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/gel-foundations",
-  "version": "0.3.0",
+  "version": "0.3.1",
```
pjlee11: Personally I think this is a much bigger change than a `patch` so would suggest `0.4.0` or maybe even `1.0.0`. Would be keen to hear what others think as well

DenisHdz: I agree, I would go for 
```suggestion
  "version": "1.0.0",
```

packages/utilities/gel-foundations/src/scripts/arabic-pashto.js
DenisHdz: Because UX hasn't tested some of the GEL Types for some scripts, but I populated them with the current values from PAL for now.

pjlee11 (approved):
Great work 👍

dr3 (approved):
In future we should try to avoid such large PRs, perhaps by only adding a couple services at a time. But I appreciate this was always going to be huge.

Great work everyone :)

amywalkerdev:
No testing needed as nothing can be visually tested in storybook yet

PR #366, opened 2019-03-18
Image widths

sareh:
Part of https://github.com/bbc/simorgh/issues/1401

**Overall change:** Removes width: 100% from figure and image

**Code changes:**

-  Removes width: 100% from figure and image
- Styled components moved to be a dev and peer dependency instead of a dependency, so that these components can be locally npm linked to the application. There needs to be a single instance of styled components. 

---

- [ ] Test engineer approval


sareh:
This PR is being tested within the Simorgh application, on this branch: https://github.com/bbc/simorgh/pull/1429

packages/components/psammead-image-placeholder/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Psammead Image Placeholder Changelog
 
+<!-- prettier-ignore -->
```
dr3, line 3: Instead of having this here, can you put an `excludeFiles` matching all `CHANGELOG.md` files in the global prettier config

ChrisBAshton: Could do, but it's kinda useful having Prettier run across the rest of the file. It's only the table that it borks things for. I'd sooner have this comment as part of the `CHANGELOG.md` template 🤔

dr3: But this currently only runs prettier on the first 2 lines of the file.. :/

ChrisBAshton: Does it? My understanding was that it only ignores the next 'block':

> A JavaScript comment of // prettier-ignore will exclude the next node in the abstract syntax tree from formatting.

https://prettier.io/docs/en/ignore.html#javascript

dr3 (changes requested):
https://github.com/bbc/simorgh/pull/1429#pullrequestreview-217722041

packages/components/psammead-image-placeholder/src/index.jsx
```diff
@@ -5,14 +5,17 @@ import { BBC_BLOCKS } from '@bbc/psammead-assets/svgs';
 
 const ImagePlaceholder = styled.div`
   position: relative;
-  height: 0;
+  height: ${props => props.ratio}vw;
```
pjlee11, line 8: `vw` doesn't seem to be supported in [opera mini](https://caniuse.com/#search=vw) are we sure using `height & vw` is a better solution than `padding-bottom & %`?

packages/components/psammead-image/src/index.jsx
```diff
@@ -6,7 +6,7 @@ export { default as AmpImg } from './index.amp';
 
 const StyledImg = styled.img`
   display: block;
-  width: 100%;
```
pjlee11, line 9: Is there a reason we wouldn't want this to have both a `width` and a `max-width`? 
The idea being the `img` should always be 100% width unless it reaches the max-width

dr3 (changes requested):
I get this locally where you can see the placeholder underneath 
IMAGE

packages/components/psammead-image-placeholder/src/index.jsx
pjlee11: caption

@dr3 found a bug with the `image-placeholder` logic when using `vw` it's easiest to replicate using browserstack. I think we should look at limiting the width of the wrapping div rather than the figure, this way we can persist with the `padding-bottom & %` which appears to be more consistent than `height & vw`

ChrisBAshton:
Blocked on https://github.com/bbc/simorgh/issues/1508.

PR #410, opened 2019-03-29
Update caption padding to match article

ChrisBAshton:
Resolves #393

Update caption padding to match article. This is a temporary change until nested grid items are added to simorgh

---

- [ ] Test engineer approval

### DIFF
--- a/packages/components/psammead-caption/src/index.jsx
+++ b/packages/components/psammead-caption/src/index.jsx
@@ -19,7 +19,7 @@ const Caption = styled.figcaption`
   color: ${C_CLOUD_DARK};
   font-family: ${GEL_FF_REITH_SANS};
   ${FS_ITALIC};
-  padding: ${GEL_SPACING} 0;
+  padding: ${GEL_SPACING};
   width: 100%;
   ${MEDIA_QUERY_TYPOGRAPHY.LAPTOP_AND_LARGER} {
     padding: ${GEL_SPACING} ${GEL_SPACING_DBL};
--- a/packages/components/psammead-caption/src/index.jsx
+++ b/packages/components/psammead-caption/src/index.jsx
@@ -1,5 +1,5 @@
 import styled, { css } from 'styled-components';
-import { GEL_SPACING, GEL_SPACING_DBL } from '@bbc/gel-foundations/spacings';
+import { GEL_SPACING } from '@bbc/gel-foundations/spacings';
 import { MEDIA_QUERY_TYPOGRAPHY } from '@bbc/gel-foundations/breakpoints';
 import {
   GEL_LONG_PRIMER,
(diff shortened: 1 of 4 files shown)


packages/components/psammead-caption/src/index.jsx
```diff
@@ -19,7 +19,7 @@ const Caption = styled.figcaption`
   color: ${C_CLOUD_DARK};
   font-family: ${GEL_FF_REITH_SANS};
   ${FS_ITALIC};
-  padding: ${GEL_SPACING} 0;
+  padding: ${GEL_SPACING};
   width: 100%;
   ${MEDIA_QUERY_TYPOGRAPHY.LAPTOP_AND_LARGER} {
     padding: ${GEL_SPACING} ${GEL_SPACING_DBL};
```
sareh: Could you please fix forward instead of reverting?

```suggestion
     padding: ${GEL_SPACING} 0;
```

ChrisBAshton: Happy to do so provided we know what we're doing! I tried this locally and it looked odd to me. Given we're launching Mon/Tues, wanted to get it back from "badly broken on mobile" to "looks a bit weird on desktop".

Does this line definitely fix it? 😀

dr3 (changes requested):
.

packages/components/psammead-caption/src/index.jsx
```diff
@@ -19,10 +26,14 @@ const Caption = styled.figcaption`
...
   font-family: ${GEL_FF_REITH_SANS};
   ${FS_ITALIC};
-  padding: ${GEL_SPACING};
+  padding: ${GEL_SPACING} 0;
   width: 100%;
-  ${MEDIA_QUERY_TYPOGRAPHY.LAPTOP_AND_LARGER} {
-    padding: ${GEL_SPACING} 0;
+
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MIN}) {
+    padding: ${GEL_SPACING} ${GEL_MARGIN_BELOW_400PX};
```
ChrisBAshton: Can we be mobile first please?

I'd have this padding value as the default ^ (on line 29)

Then we can remove this media block, and add one at the end instead:

```suggestion
@media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
  padding: ${GEL_SPACING} 0;
```

...so that it reads from mobile to bigger.

Current implementation defaults to the desktop view and then overrides for specific smaller breakpoints.

ChrisBAshton (commented):
Thanks for jumping on this. Please tweak the media query order 👍

pjlee11 (dismissed):
LGTM - Thanks for the GIFs 👍

packages/components/psammead-caption/src/index.jsx
dr3: done :)

pjlee11 (approved):

dr3 (approved):
Approving on behalf of @ChrisBAshton

jamesbrumpton:
👍

PR #428, opened 2019-04-09
Add custom propTypes shape to Headings component

### LINKED ISSUE #425: Create prop types shape for typography scripts
We should be more restrictive about what 'object' we allow

Current:
```
SubHeading.propTypes = {
  script: objectOf(object).required,
};
```

Following the [articles example](https://github.com/bbc/simorgh/blob/latest/src/app/containers/Article/index.jsx#L64), export a proptypes shape to use in downstream components

Desired:

```
SubHeading.propTypes = {
  script: objectOf(scriptPropTypes).required,
};
```

**Testing notes**
[Tester to complete]

Dev insight: When run locally, you dont get a prop type warning.

**Additional context**

DenisHdz:
Resolves #425

**Overall change:** 
We should be more restrictive about what script 'object' we allow to pass to the component.
For this reason, we should add a custom `propType` shape matching the current `scripts` values.

**Code changes:**
- Replace `defaultProps` with `propTypes` in the `Headline` and `SubHeading` components
- Import script `propType` shape from `gel-foundations`

---

- [ ] Test engineer approval


dr3 (commented):
PR looks great, the shape format is awesome. 

My only 2 concerns are, I think the propTypes shape should be published to NPM, so it can be used in simorgh, so if a container expects a script object, we can use it in the proptypes.

Second, While I love the `packages/models/propTypes`, having it disconnected from the gel-foundations package, when it directly relates to it, could cause problems. E.g. if its published, and then we add a new script name, we need to update this packages at the same time or avoid possibly an invalid propType declaration as it explicitly asks for the names of scripts. This means the 2 packages need to be tied together forever, with every package for which it has proptypes for.

curious on other opinions @pjlee11 @sareh @jblaut

pjlee11:
@dr3 the disconnect between `propTypes` and `gel-foundations` is an interesting point to raise. Rather than the packages being "tied together" is this not just a case of dependency rather than circular dependency? In that to use `gel-foundations/typography` you are dependant on the typography group being defined in `propTypes` package? Therefore, `propTypes` isn't dependant on `gel-typography` but `gel-typography` is dependant on `propTypes`? 

That said I do think because of the dependency these two should live in a single package.

**Scenario 1**
_`propTypes` and `gel-foundations` are seperate packages._
If you needed to add a new typography group (`foobar`) you first add `foobar` to `propTypes` and publish the package. You then go to `gel-foundations` and add `foobar` and its values, while pulling in the new publish of `propTypes` in as well, then you publish `gel-foundations`. Following this your component can pull in both `gel-foundations` and `propTypes` into your component.

**Scenario 2**
_`gel-foundations` contains `propTypes`._
If you needed to add a new typography group (`foobar`) you can add the propType definition to `gel-foundation/modelled-prop-types` and the new values to `gel-foundation/typography` and then publish. Following the publish both a component or a container can use the new `foobar` typography and propTypes. 

By combining these in `gel-foundations` we no longer have to worry about the dependency order and can update both typography and the propType definition in a single PR. We can also expose the `propTypes` to a container via `gel-foundations` the same way we do for `typography` or `breakpoints`

dr3:
I agree with Phil, My concern was more what happens when its used in simorgh

Say we have `psammead-proptypes` which contains proptypes for scripts and colours (for example)

We use `psammead-proptypes` in simorgh containers for prop type validation. 

We bump `psammead-proptypes` due to new scripts
We bump `psammead-proptypes` to new colours

We want to integrate the new scripts into simorgh, but in doing so, we pull in the new colours prop types, throwing prop type warnings as theyre not yet integrated.

dr3:
@DenisBBC Can you please move this work into the `gel-foundations` package please? :)

pjlee11:
>I agree with Phil, My concern was more what happens when its used in simorgh

@dr3 by importing propTypes as `gel-foundations/prop-types.js`, or similar, mitigate that worry? We will make the changes in the `gel-foundations` package before pulling it into simorgh?

dr3:
@pjlee11 yeah `import { scriptPropType } from '@bbc/gel-foundations/prop-types'` works :)

DenisHdz:
Blocked on [#431](https://github.com/bbc/psammead/pull/431)

chris-hinds (approved):
LGTM 👍

j-pendlebury (approved):

PR #445, opened 2019-04-15
First implementation of section divider

### LINKED ISSUE #437: Create a section divider
For the ws frontpages we need a section divider for between sections of cards

Create a section divider component [according to UX designs](https://app.zeplin.io/project/5c8a5e021d454124ef90669f?seid=5cac6e35d4655b65fd4169a3). This component should accept a text prop which is placed in the divider when provided. The padding around the component should be confirmed with UX. If text isn't provided this component should render the just a line, but maintain the same dimensions.

Work with UX to confirm the variations of this component, its likely we will need to support:
- with inline text
- without text
- with text below line

Consider the UX for a really long text, work to agree a design

These can all be done with 1 component, but we should have a discussion of how we want to vary the output.

Check Accessibility, confirm with devs and @greenc05 what type of `H` tag it would need to be. Does it need an aria label? Should the bar have semantic meaning ie. be an `hr` tag?

It should also accept scripts to be localised and work RTL

image

**Testing notes**
[Tester to complete]

Dev insight: Snapshots and storybook

andrew-nowak:
Resolves #437

**Code changes:**

---

- [ ] Test engineer approval

### DIFF
--- a/packages/components/psammead-section-divider/src/index.jsx
+++ b/packages/components/psammead-section-divider/src/index.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import styled from 'styled-components';
-import { boolean, node, shape } from 'prop-types';
+import { boolean, shape, string } from 'prop-types';
 import { scriptPropTypes } from '@bbc/gel-foundations/prop-types';
 import {
   getDoublePica,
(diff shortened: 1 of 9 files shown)


packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+const SectionRule = styled.div`
+  background: ${C_CLOUD_LIGHT};
+  height: 1px;
+  border: none;
+  flex-grow: 100;
+  margin: ${props => (props.inline ? 'auto 8px' : '8px')};
+`;
+
+const SectionDividerWrapper = styled.div`
```
dr3, line 37: Hey, have you perhaps considered a more simpler (imo) solution that doesn't use flex? Ive made https://codepen.io/drewmcmillan/pen/OGOzLR which mirrors the news frontpage implementation at the moment, curious of your thoughts

andrew-nowak:
Closing in favour of #449

PR #449, opened 2019-04-16
Create Section Label

### LINKED ISSUE #437: Create a section divider

andrew-nowak:
Resolves #437 

**Overall change:** Implements the ~~section divider~~ section label component

**Code changes:**

- A second attempt after #445 at implementing without using flexbox.

New:
  - ~~Exports two components; a `SectionLabel` and a `SectionLabelWithoutBar` to accommodate the two UX designs~~

Even newer:
  - Control whether you see a bar by passing in a prop `bar={false}`
  - Receive a prop to use as `id` for the label, to allow using `aria-labelledby` in the container.
  - Not providing a strapline counts as an error

---

- [ ] Test engineer approval


packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+import { GEL_SPACING } from '@bbc/gel-foundations/spacings';
+import {
+  getDoublePica,
+  GEL_FF_REITH_SANS,
+} from '@bbc/gel-foundations/typography';
+import { C_CLOUD_LIGHT, C_EBON, C_WHITE } from '@bbc/psammead-styles/colours';
+
+const TOP_MARGIN_PX = 1;
+
+const calcTop = ({ inline, script }) =>
```
dr3: I know eslint is making it this way, but is there anyway to appease it, while making it look nice and easy to read? :P

packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+`;
+
+const SectionTitle = styled.h2`
+  ${props => getDoublePica(props.script)};
+  color: ${C_EBON};
+  background-color: ${C_WHITE};
+  font-family: ${GEL_FF_REITH_SANS};
+  display: inline-block;
+  position: relative;
+  margin: ${TOP_MARGIN_PX}px 0 0 0;
```
dr3: This can probably just be this right? 
```suggestion
  margin-top: ${TOP_MARGIN_PX}px;
```

packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+  html:not([dir='rtl']) & {
```
dr3: Whats the browser support for this like? Im not sure if I prefer this, vs passing `dir` as a prop to the component

dr3 (changes requested):
Amazing work, really really nice, just a couple things

packages/components/psammead-section-divider/src/index.jsx
dr3: If we do, We can probably reduce the complexity here if we just account for the `html[dir='rtl']`

Have it default to `padding: 0 ${GEL_SPACING} 0 0;` then override if `html[dir='rtl']` to `padding: 0 0 0 ${GEL_SPACING};`

dr3: chatting to @pjlee11 in #si_repo_psammead on slack, we should pass `dir` as a prop and toggle on `rtl` as a value

packages/components/psammead-section-divider/src/index.jsx
andrew-nowak: Can it? My thinking was to force the default margins to 0 so that margins added to the Wrapper or through the container can control the placement

packages/components/psammead-section-divider/src/index.jsx
andrew-nowak: Yep, that sounds good. I don't have any attachment to this method, I was just trying to make the rtl toggle on storybook work (which may need some thought about how that would work with dir as a prop)

packages/components/psammead-section-divider/src/index.jsx
dr3: Ah good point

packages/components/psammead-section-divider/src/index.jsx
andrew-nowak: ~~How does 722d3ab look?~~

andrew-nowak:
~~Blocked on #456~~

packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+        {text}
+      </SectionTitle>
+    )}
+  </SectionDividerWrapper>
+);
+
+SectionDivider.defaultProps = {
+  children: '',
+  dir: 'ltr',
+  script: {},
```
dr3:
```suggestion
  script: null,
```

packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+    {text && (
+      <SectionTitle script={script} dir={dir}>
+        {text}
+      </SectionTitle>
+    )}
+  </SectionDividerWrapper>
+);
+
+SectionDivider.defaultProps = {
+  children: '',
```
dr3:
```suggestion
  children: null,
```

packages/components/psammead-section-divider/src/index.test.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { shouldMatchSnapshot } from '@bbc/psammead-test-helpers';
+import { latin, arabic } from '@bbc/gel-foundations/scripts';
+import SectionDivider from './index';
+
+describe('SectionDivider', () => {
+  describe('With title', () => {
+    shouldMatchSnapshot(
```
dr3: This tests that `ltr` is the default, can we have one that passes it in explicitly please :)

packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+const paddingDir = ({ dir }) => `padding-${dir === 'ltr' ? 'right' : 'left'}`;
+
+const SectionTitle = styled.h2`
+  ${({ script }) => script && getDoublePica(script)};
+  color: ${C_EBON};
+  background-color: ${C_WHITE};
+  font-family: ${GEL_FF_REITH_SANS};
+  display: inline-block;
+  position: relative;
+  margin: ${MARGIN_TOP_PX}px 0 0 0;
```
dr3: Can this just use `margin-top`? Also we should use `rem` here :)

packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import styled from 'styled-components';
+import { oneOf, shape, string } from 'prop-types';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+import {
+  GEL_GROUP_3_SCREEN_WIDTH_MIN, // 600px
```
dr3: Im not a big fan of these comments as they technically could go out of sync with the package they reference. Can they be removed please? Lines 6->22

packages/components/psammead-section-divider/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@bbc/psammead-section-divider",
+  "version": "0.1.0",
+  "description": "React styled component for a section-divider",
```
dr3:
```suggestion
  "description": "React styled component for a section divider",
```

packages/components/psammead-section-divider/src/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+const SectionDividerWrapper = styled.div`
+  position: relative;
+  margin-top: ${GEL_SPACING_TRPL};
+  margin-bottom: ${GEL_SPACING_DBL};
+  &::before {
+    content: '';
+    position: absolute;
+    height: 1px;
+    border: none;
+    background: ${C_CLOUD_LIGHT};
```
dr3: In zeplin its using the colour `#aeaeb5`, is this wrong?

dr3 (changes requested):
Looking good, Im not sure about the behaviour at 1280px, we may need to chat over how its gunna work in simorgh with the story promo list as Im not convinced it can work like it is atm

dr3:
image

For now, we can remove the code for the 1280px breakpoint, as it will involve additional logic to escape the grid column/component magic

packages/components/psammead-section-divider/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+### When to use this component
+
+This component should be used to divide the content in a page into logical elements.
+
+<!-- ### When not to use this component -->
+
+### Accessibility notes
+
+Although this component has the appearance of a horizontal rule, it does not use an `<hr>` tag, and therefore does not have the associated semantic meaning.
```
sareh: It's great that you've added this note about the semantics of this component. 👍

packages/components/psammead-section-divider/README.md
```diff
@@ -0,0 +1,60 @@
...
+| dir | string | no | 'ltr' | 'rtl' |
+| script    | object | yes | latin | { canon: { groupA: { fontSize: '28', lineHeight: '32',}, groupB: { fontSize: '32', lineHeight: '36', }, groupD: { fontSize: '44', lineHeight: '48', }, }, trafalgar: { groupA: { fontSize: '20', lineHeight: '24', }, groupB: { fontSize: '24', lineHeight: '28', }, groupD: { fontSize: '32', lineHeight: '36', }, }, }|
+
+## Usage
+
+```jsx
+import SectionDivider from '@bbc/psammead-section-divider';
+import { latin } from '@bbc/gel-foundations/scripts';
+
+const WrappingComponent = () => <SectionDivider script={latin} dir="ltr">Text here</SectionDivider>;
```
sareh: Is the recommended usage of this component to be with a heading element? for example, with an `<h2>`? If so, it would be good for the sample usage be using this. e.g.

```jsx
import SectionDivider from '@bbc/psammead-section-divider';
import { SubHeading } from '@bbc/psammead-headings';
import { latin } from '@bbc/gel-foundations/scripts';

const Wrapper = () => (

const WrappingComponent = () => (
  <SectionDivider script={latin}>
    <SubHeading script={latin}>Some subheadline</SubHeading>
  </SectionDivider>
);
```

Also, is it necessary for there to be a `dir="ltr"` on this SectionDivider element? I believe just having it further up the page should be fine.  The reason I'm noting this, is because often people copy and paste the Readme example into their code, so it's good to make sure we have examples that people can clearly use immediately.

packages/components/psammead-section-divider/README.md
```diff
@@ -0,0 +1,60 @@
+# psammead-section-divider - [![Known Vulnerabilities](https://snyk.io/test/github/bbc/psammead/badge.svg?targetFile=packages%2Fcomponents%2Fpsammead-section-divider%2Fpackage.json)](https://snyk.io/test/github/bbc/psammead?targetFile=packages%2Fcomponents%2Fpsammead-section-divider%2Fpackage.json) [![Storybook](https://raw.githubusercontent.com/storybooks/brand/master/badge/badge-storybook.svg?sanitize=true)](https://bbc.github.io/psammead/?path=/story/section-divider--default) [![GitHub license](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](https://github.com/bbc/psammead/blob/latest/LICENSE) [![npm version](https://img.shields.io/npm/v/@bbc/psammead-section-divider.svg)](https://www.npmjs.com/package/@bbc/psammead-section-divider) [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](https://github.com/bbc/psammead/blob/latest/CONTRIBUTING.md)
+
+## Description
+
+The `@bbc/psammead-section-divider` package exports a single Section Divider component.
```
sareh: This is a comment about the naming of this component within Psammead.  Can it please be renamed as `@bbc/psammead-divider`? This is because the use of `section-divider` appears to refer to the HTML semantic `<section>`, which is not the case here.

packages/components/psammead-section-divider/src/index.jsx
andrew-nowak: > Can this just use `margin-top`?

No, this is to override the browser's default margins for an h2. I'll add a comment explaining why

> Also we should use `rem` here :)

:+1:

packages/components/psammead-section-divider/README.md
andrew-nowak: > Is the recommended usage of this component to be with a heading element? 

No, it isn't. This component puts the children it receives (and as the prop-types make clear, it should only be passed a string) in its own `<h2>`, and intentionally not the SubHeading component, as it uses Double Pica sizing, rather than Trafalgar as in the SubHeading. I will make this more clear on the readme.

> Also, is it necessary for there to be a `dir="ltr"` on this SectionDivider element? I believe just having it further up the page should be fine. The reason I'm noting this, is because often people copy and paste the Readme example into their code, so it's good to make sure we have examples that people can clearly use immediately.

No it's not required, but it is valid, the example would work, and IMO helps to remind someone who is copy-and-pasting that, unlike the majority of other components, the styling won't 'just work' for rtl text and whether the prop is being set needs to be considered.

packages/components/psammead-section-divider/src/index.jsx
andrew-nowak: Caught up with UX, the colour has changed from C_CLOUD_LIGHT to C_PEBBLE (`#aeaeb5`)... except C_PEBBLE doesn't exist in `psammead-styles`, so I raised #501

andrew-nowak:
~~Blocked on #501 - see [this comment](https://github.com/bbc/psammead/pull/449#discussion_r281716950)~~
No longer blocked :)

packages/components/psammead-section-divider/README.md
andrew-nowak: I had a catch up with UX about this component yesterday, they are now referring to this component as a 'Section Label'. Would it be best to rename this component to align as closely as possible with their name? This would also have the issue that `section-label` might be referring to `<section>`, but `label` also feels very vague.

j-pendlebury (dismissed):

packages/components/psammead-section-label/src/index.jsx
```diff
@@ -0,0 +1,147 @@
...
+);
+
+SectionLabel.defaultProps = {
+  children: null,
+  dir: 'ltr',
+  script: null,
+};
+
+SectionLabel.propTypes = {
+  children: string,
```
pjlee11: Are `children` required?

packages/components/psammead-section-label/src/index.jsx
```diff
@@ -0,0 +1,147 @@
...
+  dir: oneOf(['ltr', 'rtl']),
+  script: shape(scriptPropType),
```
pjlee11: Same as above, can we render a section label without `script` passed in? Should it be required?

packages/components/psammead-section-label/src/index.jsx
andrew-nowak: ~~not in my implementation, I've handled the case where children aren't provided. My thinking was to handle the case where a strapline was missing from the data, and rendering the bar without the extra spacing for the missing text - see <https://github.com/bbc/psammead/pull/449/files#diff-a826825d0bceab40bf039858608d0224R56>~~

This is no longer the case, it's been confirmed that (at least for now) we will consider a missing strapline an error and won't handle that case

ghost:
The current implementation exports `SectionLabel` and `SectionLabelWithoutBar`. Have you looked into exporting `SectionLabel` and `SectionDivider` and have 
```jsx
<SectionDivider>
  <SectionLabel>
    Most Read
  </SectionLabel>
</SectionDivider>
```
just as you stated on your slack. IMO this will be a much better implementation. Glad to hear your thoughts on this.

packages/components/psammead-section-label/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Description
+
+The `@bbc/psammead-section-label` package exports two components - a Section Label, and a variant of the Section Label that does not feature a horizontal bar.
+
+It uses `@bbc/psammead-styles` for colours and font family and `@bbc/gel-foundations` for spacing and for GEL Typography implemented in Styled Components.
```
ghost: We might need to reword this 
```suggestion
For colours and font family it uses `@bbc/psammead-styles` and `@bbc/gel-foundations` for spacing and GEL Typography implemented in Styled Components.
```

andrew-nowak:
> The current implementation exports `SectionLabel` and `SectionLabelWithoutBar`. Have you looked into exporting `SectionLabel` and `SectionDivider` and have
> 
> ```js-jsx
> <SectionDivider>
>   <SectionLabel>
>     Most Read
>   </SectionLabel>
> </SectionDivider>
> ```
> 
> just as you stated on your slack. IMO this will be a much better implementation. Glad to hear your thoughts on this.

Good question! The short answer is that it doesn't really work without some pretty ugly logic.

1) To set the divider line to be vertically centered next to the label, I use `position: absolute;` and set `top` to half the text's line height. However, if the Label and the Divider are separate components, the Divider shouldn't require the `script` prop -- it shouldn't have text as a direct child. Even if we do accept the `script` prop, using those values without a Label would effectively increase the top margins and result in too large a spacing between sections. You technically *could* inspect the children, and if the child was a Label then you could use the values from its script prop, but that is definitely quite ugly.

2) The divider should control the spacing between sections, ie. sets a margin-top and margin-bottom, but if we export both a Divider and a Label and allow them to be used independently, how should the margins be applied? When trying to make this work, I ended up with

```css
*:not(${/* sc-selector */ SectionDivider}) > & {
  ${margins};
}
```

which IMO is worse than just using props to control whether the bar exists, and then being able to control this logic internally.

After some more discussion on slack, there was an agreement to move from the solution which you reviewed and use props to control the visibility of the horizontal bar, rather than a second export, which I have now pushed.

packages/components/psammead-section-label/README.md
```diff
@@ -0,0 +1,79 @@
...
+For colours and font family it uses `@bbc/psammead-styles` and `@bbc/gel-foundations` for spacing and GEL Typography implemented in Styled Components.
+
+The only provided child should be a *string*, which will be wrapped in an `<h2>` element by the component – (see [the Accessibility notes](#accessibility-notes)).
+
+## Installation
+
+`npm install @bbc/psammead-section-label`
+
+## Props
+
```
ghost, line 20: You may also need to add prettier-ignore comment here.

packages/components/psammead-section-label/README.md
```diff
@@ -0,0 +1,79 @@
...
+| Argument  | Type | Required | Default | Example |
+| --------- | ---- | -------- | ------- | ------- |
+| bar | boolean | no | `true` | `false` |
+| children | string | no | N/A | `'Most Read'` |
+| dir | string | no | `'ltr'` | `'rtl'` |
+| script | object | yes | N/A | `{ canon: { groupA: { fontSize: '28', lineHeight: '32',}, groupB: { fontSize: '32', lineHeight: '36' }, groupD: { fontSize: '44', lineHeight: '48' } }, trafalgar: { groupA: { fontSize: '20', lineHeight: '24' }, groupB: { fontSize: '24', lineHeight: '28' }, groupD: { fontSize: '32', lineHeight: '36' } } }` |
```
ghost:
```suggestion
| script | object | yes | N/A | { canon: { groupA: { fontSize: '28', lineHeight: '32',}, groupB: { fontSize: '32', lineHeight: '36' }, groupD: { fontSize: '44', lineHeight: '48' } }, trafalgar: { groupA: { fontSize: '20', lineHeight: '24' }, groupB: { fontSize: '24', lineHeight: '28' }, groupD: { fontSize: '32', lineHeight: '36' } } } |
```
The back-ticks causes the script example to be displayed on one line, which becomes too long forcing a horizontal scroll on the browser.

packages/components/psammead-section-label/src/index.jsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+import { boolean, oneOf, shape, string } from 'prop-types';
```
ghost:
```suggestion
import { bool, oneOf, shape, string } from 'prop-types';
```

packages/components/psammead-section-label/src/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+
+SectionLabel.defaultProps = {
+  bar: true,
+  children: null,
+  dir: 'ltr',
+  script: null,
+};
+
+SectionLabel.propTypes = {
+  bar: boolean,
```
ghost:
```suggestion
  bar: bool,
```

packages/components/psammead-section-label/README.md
```diff
@@ -0,0 +1,80 @@
...
+import { SectionLabel } from '@bbc/psammead-section-label';
+import { latin } from '@bbc/gel-foundations/scripts';
+
+const WrappingComponent = () => <SectionLabel script={latin} dir="ltr">Text here</SectionLabel>;
+```
+
+Or, without a section title:
+
+```jsx
+import { SectionLabel } from '@bbc/psammead-section-label';
```
ghost: Section label is exported as a default
```suggestion
import  SectionLabel from '@bbc/psammead-section-label';
```

packages/components/psammead-section-label/README.md
```diff
@@ -0,0 +1,80 @@
...
+| --------- | ---- | -------- | ------- | ------- |
+| bar | boolean | no | `true` | `false` |
+| children | string | no | N/A | `'Most Read'` |
+| dir | string | no | `'ltr'` | `'rtl'` |
+| script | object | yes | N/A | { canon: { groupA: { fontSize: '28', lineHeight: '32',}, groupB: { fontSize: '32', lineHeight: '36' }, groupD: { fontSize: '44', lineHeight: '48' } }, trafalgar: { groupA: { fontSize: '20', lineHeight: '24' }, groupB: { fontSize: '24', lineHeight: '28' }, groupD: { fontSize: '32', lineHeight: '36' } } } |
+
+## Usage
+
+```jsx
+import { SectionLabel } from '@bbc/psammead-section-label';
```
ghost:
```suggestion
import SectionLabel from '@bbc/psammead-section-label';
```

packages/components/psammead-section-label/README.md
```diff
@@ -0,0 +1,80 @@
...
+import { latin } from '@bbc/gel-foundations/scripts';
+
+const WrappingComponent = () => <SectionLabel />;
+```
+
+Or, without a horizontal bar:
+
+```jsx
+import { SectionLabelWithoutBar } from '@bbc/psammead-section-label';
```
ghost:
```suggestion
import SectionLabel from '@bbc/psammead-section-label';
```

ghost (commented):
Looks great. Just a few things

packages/components/psammead-section-label/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Psammead Section Label Changelog
+
+<!-- prettier-ignore -->
+| Version | Description |
+|---------|-------------|
+| 0.1.0-alpha.0 | [PR#449](https://github.com/BBC/psammead/pull/449) Create initial package with Section Label components. |
```
ghost, line 6:
```suggestion
| 0.1.0-alpha.1 | [PR#449](https://github.com/BBC/psammead/pull/449) Create initial package with Section Label components. |
```
Alpha versions usually start from `.1`

packages/components/psammead-section-label/package-lock.json
```diff
@@ -0,0 +1,393 @@
+{
+  "name": "@bbc/psammead-section-label",
+  "version": "0.1.0-alpha.0",
```
ghost, line 3:
```suggestion
  "version": "0.1.0-alpha.1",
```

packages/components/psammead-section-label/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@bbc/psammead-section-label",
+  "version": "0.1.0-alpha.0",
```
ghost, line 3:
```suggestion
  "version": "0.1.0-alpha.1",
```

packages/components/psammead-section-label/README.md
andrew-nowak:
```suggestion
import SectionLabel from '@bbc/psammead-section-label';
```

packages/components/psammead-section-label/CHANGELOG.md
andrew-nowak: do you have a reference for this? I don't see it mentioned on <https://semver.org>, and my understanding of the team's decision was to start at `.0`, and it was used for [story promo list](https://www.npmjs.com/package/@bbc/psammead-story-promo-list)

ghost: I can see it was indeed used for `story promo list`. I wasn't on the discussion on the `alpha` releases, just ignore the suggestions. Ideally it should be either `0.1.0-alpha` or `0.1.0-alpha.1`.

ghost (dismissed):
👍 Looks Great.

packages/components/psammead-section-label/src/index.jsx
ghost: @aacn500 I think script should be required. Failure to provide script causes the label to be out of line.
Screenshot 2019-05-17 at 12 06 26

DenisHdz (dismissed):
👍

packages/components/psammead-section-label/src/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+`;
+
+Title.propTypes = {
+  dir: oneOf(['ltr', 'rtl']).isRequired,
+  script: shape(scriptPropType).isRequired,
+};
+
+const SectionLabel = ({ bar, children: title, dir, labelId, script }) => (
+  <Wrapper script={title && script} bar={bar}>
```
ghost: I don't understand why we have `title && script`

andrew-nowak: Woops, I missed one. Well caught, thanks!

ghost (approved):

andrew-nowak:
There are remaining questions over how to display this component when there is not a `strapline` available. However, if required, that will be raised as a future issue. For now, we will assume that straplines are always provided.

dr3 (approved):
LGTM, great work 👍

david-boydell:
All good, thanks @aacn500, this is rfm.

PR #461, opened 2019-04-23
Make timestamp typography style agnostic

### LINKED ISSUE #460: Make the timestamp component typography styling agnostic
###  Precursor: 
Confirm with UX that we are likely to need the timestamp in different typography styles given the scenario the component is used in

---

Our current timestamp component is [hardcoded](https://github.com/bbc/simorgh/blob/latest/src/app/components/Timestamp/index.jsx#L15) to use `GEL_BREVIER`. While this is correct for articles, this may not be the case going forward for other usecases, such as WS frontpages or elsewhere.

Make the timestamp component styling agnostic in a way that allows its font styling to be influenced. This could be done via passing props, or via style inheritance from its wrapper.

We could pass in a `get` typography function, such as `getPica`, or we can style it from a wrapper

Allow it to be configured as a knob in storybook to switch styles. 

**Testing notes**
[Tester to complete]

Dev insight: Unit tests snapshots and inspection in storybook should be sufficient.

**Additional context**

ghost:
Resolves #460

**Overall change:** Make the `Timestamp` component style agnostic to enable passing in of different typography styles.

**Code changes:**

-  Removed `GEL_BREVIER` to allow the style to be passed in via props.
-  Introduced `typographyStyle` prop which sets the  style  on `StyledTimestamp`.

---

- [ ] Test engineer approval


packages/components/psammead-timestamp/README.md
```diff
@@ -22,12 +22,20 @@ When a date or time is to be displayed inline inside a paragraph.
...
+| `typographyStyle` | string | Yes | N/A | `font-size: 1rem` |
 
 ## Usage
 
-```js
+```jsx
+import { latin } from '@bbc/gel-foundations/scripts';
+import { getPica } from '@bbc/gel-foundations/typography';
+
+const typographyStyle = getPica(latin);
```
dr3: Im curious of others opinions on this,

I imagined passing in the function `get*` (e.g. getGreatPrimer) to be used in psammead-timestamp itself. Means we can also default to `getPica` internally for usability

thoughts @pjlee11 @sareh @DenisBBC?

packages/components/psammead-timestamp/src/index.stories.jsx
```diff
@@ -1,26 +1,57 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
-import { text, withKnobs } from '@storybook/addon-knobs';
+import { text, select, withKnobs } from '@storybook/addon-knobs';
+import { latin } from '@bbc/gel-foundations/scripts';
+import * as typography from '@bbc/gel-foundations/typography';
 import notes from '../README.md';
 import Timestamp from '.';
 
+const styles = [
```
dr3: Rather than having these hardcoded here? Could you maybe use the `typography` object to get the key values? Means if we were to ever add a new one, it would just work with a version bump

dr3 (changes requested):
Just a couple things

packages/components/psammead-timestamp/README.md
pjlee11: Are you saying something like:
```
  <Timestamp datetime="2019-03-01T14:00+00:00" typographyFunc={getPica} script={latin}>
    Friday 1 March 2019
  </Timestamp>
```

dr3: Yeah, with an internal proptypes default of `getPica`

pjlee11: That would keep inline with the other components which currently internally have the `script` mocked in their tests/stories. Feels like a more consistent approach

packages/components/psammead-timestamp/src/index.jsx
```diff
@@ -5,14 +5,11 @@ import {
...
-import {
-  GEL_BREVIER,
-  GEL_FF_REITH_SANS,
-} from '@bbc/gel-foundations/typography';
+import { GEL_FF_REITH_SANS } from '@bbc/gel-foundations/typography';
 import { C_CLOUD_DARK } from '@bbc/psammead-styles/colours';
 
 const StyledTimestamp = styled.time`
-  ${GEL_BREVIER};
+  ${props => props.typographyStyle}
```
DenisHdz: I wonder if this will fail if we don't pass the `typographyStyle` prop?
Shouldn't be `${props => (props. typographyStyle ? props.typographyStyle : '')};` ?

I am just asking because in the others components we are checking if we pass the  `script` prop before calling the GEL function (e.g.  ` ${props => (props.script ? getLongPrimer(props.script) : '')};`) and I would like to know what is the right way.

packages/components/psammead-timestamp/src/index.test.jsx
```diff
@@ -1,15 +1,26 @@
 import React from 'react';
 import { shouldMatchSnapshot } from '@bbc/psammead-test-helpers';
+import { latin } from '@bbc/gel-foundations/scripts';
+import { getPica } from '@bbc/gel-foundations/typography';
```
DenisHdz, line 4: We should be using `getBrevier` here as well?

ghost: `getBrevier` is the default `typographyFunc` and will be used on the first test case that doesnt pass `typographyFunc` and `script` to `Timestamp` therefore not needed

dr3 (approved):

DenisHdz (approved):
Great work 👍

david-boydell:
Discussed with Isaac that no testing is required.

PR #492, opened 2019-05-02
Make width optional

### LINKED ISSUE #467: Image width prop should be optional
In the `psammead-image` we currently have props for both height and width. However, height is an optional value due to [this issue](https://github.com/bbc/psammead/pull/234) and width is required making them inconsistent. For context `<amp-img>` requires both height and width to be AMP valid. This means in some cases height and width are required and in other cases they aren't strictly required HTML attributes on the element. 

Also we are aiming for our psammead components to be as re-usable as possible meaning we shouldn't enforce width/height being passed in to the component but instead enforce this in the container bespoke to our rendering scenario. 

- Make width prop optional on the image component 
- Explain the benefits of passing width and height props in the psammead-image README.md

**Testing notes**
[Tester to complete]

Dev insight: There should be no regression in the component. There should be stories and unit test for the optional behaviour of the width prop.

**Additional context**

ghost:
Resolves #467 

**Overall change:** Made image width prop optional.

**Code changes:**
- Made `width` an optional prop
- Added a story with the image width not specified
- Added test to cover the optional width for `Img`
- Updated test snapshots 

---

- [ ] Test engineer approval


packages/components/psammead-image/src/testHelpers/stories.jsx
```diff
@@ -69,5 +69,20 @@ const stories = (
...
       { notes },
     );
+  if (Component.name === 'Img') {
+    imageStories.add(
+      'image without width',
+      () => (
+        <Component
+          {...getProps(landscape, includeHeight)}
+          {...additionalProps}
+          width={190}
```
DenisHdz: If this new story is for `images without width` why are we passing a `width` prop? Just asking :)

ghost: Still in progress. i was just testing out some stuff.

packages/components/psammead-image/README.md
```diff
@@ -59,12 +59,17 @@ const WrappingContainer = ({ alt, src, height, width }) => (
...
 | `src`    | string        | Yes      | -       | "https://bbc.com/300/cat.jpg"                                                                          |
 | `srcset` | string        | No       | null    | "https://bbc.com/300/cat.jpg 300w, https://bbc.com/450/cat.jpg 450w, https://bbc.com/600/cat.jpg 600w" |
-| `width`  | number/string | Yes      | -       | 600                                                                                                    |
+| `width`  | number/string | No       | null    | 600                                                                                                    |
+
+The `height` and `width` props are optional, in some cases to preserve the image ratio you might specify either `height` or `width` and let the browser scale the image accordingly.
+
+However when not specified the browser will not be able to determine the size of the image, the browser will therefore build the page twice or more depending on the number of images you have. First build is for the browser to display all the text and once the image is downloaded and size determined a second build to wrap the texts around the image.
```
dr3: By build, do you mean paint?

packages/components/psammead-image/src/index.jsx
```diff
@@ -15,6 +15,9 @@ export const Img = ({ alt, src, srcset, height, width }) => {
   if (srcset) {
     props.srcSet = srcset;
   }
+  if (!width) {
```
dr3: We dont need this, if `width` is falsey, it wont have an affect on the component

packages/components/psammead-image/src/index.test.jsx
```diff
@@ -1,10 +1,6 @@
-import Image, { Img } from '.';
+import Image from '.';
 import snapshotTests from './testHelpers/snapshotTests';
 
-describe("Image - imported as '{ Img }'", () => {
-  snapshotTests(Img);
-});
-
-describe("Image - imported as default 'Image'", () => {
+describe("Image - Img'", () => {
```
dr3: Why have you changed this test? We need to test both export methods

packages/components/psammead-image/src/testHelpers/snapshotTests.jsx
```diff
@@ -58,6 +58,18 @@ const snapshotTests = (Component, additionalProps) => {
       width={landscape.width}
     />,
   );
+  if (Component.name === 'Img') {
```
dr3: Why are you doing this `if`?

packages/components/psammead-image/src/testHelpers/stories.jsx
```diff
@@ -69,5 +69,20 @@ const stories = (
       ),
       { notes },
     );
+  if (Component.name === 'Img') {
```
dr3: Why are you doing this `if`?

dr3 (changes requested):

packages/components/psammead-image/src/testHelpers/stories.jsx
ghost: This is done to avoid creating stories for `AmpImg`, since `testHelpers/stories.jsx` is used to create both `Img` and `AmpImg` stories.

packages/components/psammead-image/README.md
ghost: Yah paint. Am changing it to `paint` for clarity. Thanks for bringing that up.

packages/components/psammead-image/src/testHelpers/stories.jsx
dr3: I dont think we need it, it didnt need it when `height` was made optional. Just dont pass `width`

packages/components/psammead-image/src/index.jsx
ghost: We need it in order to avoid `width=null` in the snapshot data

dr3: is that not what happens with height? React is clever enough to not render `null` in the dom

You can just not pass width, it wont render non declared null values

ghost: I understand the React bit. I would like to get your opinion on this. Would you consider
```jsx
<img width={null} src="http://example.com/car.png"/>
```
to be a snapshot of 
```jsx
<img src="http://example.com/car.png"/>
```
That is actually rendered on the `DOM`.  IMO a snapshot should be what the component looks like when rendered.

dr3: But if you didn't pass 'width=null' in the first place, it wouldn't show in the snapshot

ghost: Since `width` is not a required prop what would the default value be?

dr3: Null, we should just copy the approach taken with height, which didn't/doesnt need this logic

ghost: The approach taken with `height` doesn't need it because there is no test defined for `image` without `height`. Going with that approach means no tests for optional `width` too. But on the [issue description](https://github.com/bbc/psammead/issues/467#issue-436713861), it is stated
> There should be stories and unit test for the optional behaviour of the width prop.

dr3: I agree we should have a test, and a snapshot with `width={null}` is fine :)

packages/components/psammead-image/src/testHelpers/snapshotTests.jsx
ghost: This is to avoid creating extra test cases and snapshots for `AmpImg` without `width`. Since `width` is a required prop in `AmpImg`, my assumption is, it will always be provided. I therefore saw no benefit in the extra test cases and snapshots.

packages/components/psammead-image/src/testHelpers/stories.jsx
ghost: I could remove the `if` but that means we create a story of `AmpImg` without `width`. `width` being a required prop on `AmpImg`, it doesn't make sense to have a story to preview the behaviour of `AmpImg` when `width` is not provided.

dr3: Why didnt we need one for height?

dr3: If you want a seperate one, can you make it a separate export that we can just use for the normal image? Reusable tests shouldnt have specific edge case handling

packages/components/psammead-image/src/index.test.jsx
```diff
@@ -1,10 +1,29 @@
...
+describe("Image - imported as default 'Image'", () => {
   snapshotTests(Img);
 });
 
-describe("Image - imported as default 'Image'", () => {
+describe("Image - imported as '{ Img }'", () => {
   snapshotTests(Image);
 });
+
+describe("Image - imported as default 'Image'", () => {
```
dr3: These two can probably be nested inside the describes in lines 7 and 11, as they appear to have name duplicates :)

packages/components/psammead-image/src/testHelpers/stories.jsx
```diff
@@ -21,8 +21,8 @@ const stories = (
   includeHeight = false,
   additionalProps = {},
   styleDecorator = storyFn => storyFn(),
-) =>
-  storiesOf(title, module)
+) => {
+  const imageStories = storiesOf(title, module)
```
dr3: Howcome this needed to be declared and returned? can this be reverted?

dr3 (changes requested):
Just two small things :)

dr3 (approved):

pjlee11 (approved):
LGTM 👍

david-boydell:
Regression on Psammead Storybook complete

PR #521, opened 2019-05-14
make the script prop optional + snapshot + story

### LINKED ISSUE #520: Paragraph: script should be optional
As described [here](https://github.com/bbc/simorgh/pull/1692#issuecomment-491305611) we want to be able to use `psammead-paragraph` but with the typography settings being optional. The initial solution to this is making `script` optional so no line-heights or font-sizes are explicitly set meaning the line-height and font-size is inherited from the parent element. 

Make the `script` value optional.

When a script is not passed in, it won't have the default paragraph styles (font usage & spacing). 

We could pass in `typographyFunc` the same way that [`psammead-timestamp` currently does](https://github.com/bbc/psammead/blob/1e8281ab965cd8ac0f95e633f5f781a00ac935f4/packages/components/psammead-timestamp/src/index.jsx#L14)

**Testing notes**
This will require regression testing of the paragraph logic within Storybook.

Dev insight: Unit tests are required and storybook regression of the `psammead-paragraph`

**Additional context**
Is related to https://github.com/bbc/simorgh/pull/1692

pjlee11:
Resolves #520

**Overall change:** Makes the script prop optional.

**Code changes:**

- Makes the script prop optional
- Adds a story for paragraph without typography styling as script is not passed
- Adds a snapshot for paragraph without typography styling as script is not passed

---

- [ ] Test engineer approval


packages/components/psammead-paragraph/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 1.0.5   | [PR#xxx](https://github.com/bbc/psammead/pull/xxx) Make script prop optional |
```
j-pendlebury:
```suggestion
| 1.0.5   | [PR#xxx](https://github.com/bbc/psammead/pull/521) Make script prop optional |
```

j-pendlebury (changes requested):
`CHANGELOG` needs updating, otherwise LGTM 👍

packages/components/psammead-paragraph/CHANGELOG.md
pjlee11:
```suggestion
| 1.0.5   | [PR#521](https://github.com/bbc/psammead/pull/521) Make script prop optional |
```

j-pendlebury (dismissed):

packages/components/psammead-paragraph/README.md
```diff
@@ -14,11 +16,18 @@ It uses `@bbc/psammead-styles` for colours and font family and `@bbc/gel-foundat
...
 
 ## Usage
 
+```jsx
+import Paragraph from '@bbc/psammead-paragraph';
+
+const WrappingComponent = () => <Paragraph>Text here</Paragraph>;
+```
+
+When passing `script` prop:
```
sareh: Could you please swap these examples around and clarify that this example is the recommended usage. 

That the one without the script should only be used when nesting the paragraph.

pjlee11: I'm personally not convinced the ideal use case is to pass the `script`. We should pass the script where it is needed but the default behaviour of a browser is that the element will inherit CSS. By suggesting we explicitly pass `script` with every `<p>` tag doesn't feel like a good solution IMO

sareh: I think it should be the default usage, since otherwise you don't get any `font-size` or `line-height` defined for the paragraph and so the component no longer follows GEL Typography specs, when we've stated clearly that our `psammead` components are all GEL Foundations-compliant.

pjlee11: Ahhh good point, in terms of GEL compliance as an individual component the default should be with a script. Will update

DenisHdz (approved):
LGTM 👍

pjlee11:
Closing in favour of https://github.com/bbc/psammead/pull/524

packages/components/psammead-paragraph/src/index.stories.jsx
```diff
@@ -13,4 +13,11 @@ storiesOf('Paragraph', module)
       <Paragraph script={script}>{paragraph}</Paragraph>
     )),
     { notes, knobs: { escapeHTML: false } },
+  )
+  .add(
+    'without script',
```
12, line 18: As `script` is now optional, I think it would make more sense to me if this was named `default`, and then this would be named `with script`:

https://github.com/bbc/psammead/blob/00be6dcaa1b72a492a620428f23fa1eb629641e9/packages/components/psammead-paragraph/src/index.stories.jsx#L10-L16

12 (changes requested):

pjlee11:
Further context here https://github.com/bbc/simorgh/pull/1692#issuecomment-492414908

PR #525, opened 2019-05-15
Update getVersionScript to return array of published package versions

### LINKED ISSUE #518: Publishing is failing due to alpha versioning
**Describe the bug**
Due to how we check for published packages in the registry, we try to publish already published alpha packages, throwing an error

**To Reproduce**

1. Run the psammead jenkins job
2. See error

```
You cannot publish over the previously published versions: 0.1.0-alpha.1. : @bbc/psammead-story-promo-list
```

**Expected behavior**
It knows its published and doesnt try

**Additional context**
This is failing as `npm show @bbc/psammead-story-promo-list version` returns `0.1.0-alpha.0`, the newest release tagged as `latest`. However, as `0.1.0-alpha.1` has been published, when it tries, it fails.

**Possible solution:**
Change [this code](https://github.com/bbc/psammead/blob/latest/scripts/publish/src/getRegistryVersion.js) to run `npm show ${name} versions` instead of `npm show ${name} version`. This returns an array of all published versions, regardless of if they are alpha or not. If our version matches any in that array, dont publish.

chris-hinds:
Resolves #518 

**Overall change:** 
Returns an array of published versions of components form the NPM registry and checks the latest version against our local version.

Previously returning a single version would not show the latest published alpha version and would prevent us from publishing updated alphas.

**Code changes:**

- Uses the NPM `versions` endpoint to return an array of package versions
- requests JSON from the endpoint
- Parse's the JSON making sure it is an array and gets the latest version

---

- [X] Tests added for new features
- [ ] Test engineer approval

### DIFF
--- a/scripts/publish/src/getRegistryVersion.js
+++ b/scripts/publish/src/getRegistryVersion.js
@@ -2,6 +2,15 @@ const { exec } = require('shelljs');
 
 // Get version of package in NPM regsitry. Returns 0.0.0 if doesn't exist.
 module.exports = name => {
-  const npmVersion = exec(`npm show ${name} version`, { silent: true }).stdout;
-  return npmVersion ? npmVersion.split('\n')[0] : '0.0.0';
+  let packageVersions = [];
+  const npmVersion = exec(`npm view ${name} versions -json`, { silent: true })
+    .stdout;
+
+  if (typeof npmVersion === 'string' && JSON.parse(npmVersion)) {
+    const jsonResponse = JSON.parse(npmVersion);
+    packageVersions = Array.isArray(jsonResponse) ? jsonResponse : [];
+  }
+
+  const latestPublishedVersion = packageVersions.pop(); // removes the last element from the array and returns it i.e. the latest published version
+  return latestPublishedVersion || '0.0.0';
 };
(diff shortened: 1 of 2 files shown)


DenisHdz (approved):
Great job, lgtm 👍🏻

j-pendlebury (approved):

scripts/publish/src/getRegistryVersion.js
```diff
@@ -2,6 +2,15 @@ const { exec } = require('shelljs');
 
 // Get version of package in NPM regsitry. Returns 0.0.0 if doesn't exist.
 module.exports = name => {
-  const npmVersion = exec(`npm show ${name} version`, { silent: true }).stdout;
-  return npmVersion ? npmVersion.split('\n')[0] : '0.0.0';
+  let packageVersions = [];
+  const npmVersion = exec(`npm view ${name} versions -json`, { silent: true })
```
DemonusPC, line 6: Sorry to be a buzzkill but this creates a serious command injection vulnerability. May I suggest using [spawn](https://nodejs.org/api/child_process.html). Using `spawn` enforces the command and parameter separation, that way you should remove this potential command injection.

DemonusPC (changes requested):
Great work overall. But there is a serious command injection vulnerability in here. It is actually quite easy to fix tho. Just use spawn instead of exec.

sareh (approved):
👍 Looks great!

DemonusPC (approved):

DemonusPC:
It was decided with @hindsc52 that we will create a separate issue that will address this command injection vulnerability.

jamesbrumpton:
Spoken with @hindsc52 about this issue. Happy for this to be merged.

PR #480, opened 2019-04-29
Control height of Brand SVG

andrew-nowak:
Resolves #466 
Resolves #499 

**Overall change:** Height of the SVG in the Brand component is dynamically scaled between the prop values `minWidth` and `maxWidth`. Also the Banner height and padding changes at 600px.

**Code changes:**

- Banner height is now 56px when window is 0-599px and 80px for 600+px
- SVG left-right padding is now 8px when window is 0-599px and 16px for 600+px
- SVG dynamically scales (while remaining centred) between the `minWidth` and `maxWidth` values. 
- SVG is vertically centred in the banner using `margin-top`
- Storybook: add knobs to Brand story which allow for passing of `minWidth` `maxWidth` and `height` to showcase the flexibility of the component and the dynamic scaling of the SVG

**Testing notes**
Check the brand component in storybook and test cross browser the functionality to dynamically scale the SVG's at small widths works as expected. 

---

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Tests added for new features
- [ ] Test engineer approval

PR #528, opened 2019-05-16
Dynamic brand link

### LINKED ISSUE #500: Make Branding SVGs link href dynamic
Currently, the SVGs are not dynamically linked but hardcoded to `bbc.co.uk/news` instead.

The SVG branding should be dynamically linked. Accept a prop to be used as the inline link href. 

Default to the bbc homepage `https://www.bbc.com` (CONFIRM WITH PRODUCT), Maybe we could not use a link if a href isnt provided

**Testing notes**
The SVGs are linked to the relative branding.

Dev insight: Unit tests and storybook, have the link be a text knob

**Additional context**
https://github.com/bbc/psammead/blob/latest/packages/components/psammead-brand/src/index.jsx#L80

ghost:
Resolves #500 

**Overall change:** Make Branding href link dynamic to allow passing of different service links.

**Code changes:**
- Added a `link` prop which is required to allow passing of the service url to brand
- Handled what should happen incase a link is not provided to `Brand`
- Added a story to preview `Brand` with link not provided.

---

- [ ] Test engineer approval


DenisHdz (dismissed):
LGTM 👍

packages/components/psammead-brand/src/index.jsx
```diff
@@ -45,10 +45,10 @@ const StyledWrapper = styled.div`
 const ConstraintWrapper = styled.div`
   max-width: ${GEL_GROUP_5_SCREEN_WIDTH_MIN};
   margin: 0 auto;
+  padding-top: ${GEL_SPACING};
```
pjlee11: Does moving this padding top from the `<a>` to here not reduce the click area? UX had explicitly set the click area to be as big as possible EG: the height of the banner

packages/components/psammead-brand/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
+| 1.1.0   | [PR#528](https://github.com/bbc/psammead/pull/528) Add link prop to allow dynamic link |
```
12: This should be a major breaking change. You're adding a required prop `link`.

packages/components/psammead-brand/README.md
```diff
@@ -10,10 +10,12 @@ The `Brand` component provides the BBC service logo (as SVG), nested inside a st
 
 ## Props
 
+<!-- prettier-ignore -->
 | Argument  | Type   | Required | Default | Example      |
 | --------- | ------ | -------- | ------- | ------------ |
 | brandName | String | yes      | N/A     | `'BBC News'` |
 | svg | Object | yes | N/A | { group: `(<g fillrule="evenodd"><path d="M84.32" /></g>)`, viewbox: { height: 24, width: 167.95 }, ratio: 6.9979 } |
+| link | String | yes | N/A | `https://www.bbc.co.uk/news` |
```
12: I personally would prefer this to be called `href`

packages/components/psammead-brand/src/index.stories.jsx
```diff
@@ -18,10 +18,19 @@ storiesOf('Brand', module)
...
+          brandName="Default Brand Name"
+          svg={svgs[choice]}
+          link="https://bbc.com/news"
+        />
+      );
     },
     { notes },
-  );
+  )
+  .add('without link', () => (
```
12: There should not be a story testing for without a link if `link` is a required proptype?

packages/components/psammead-brand/src/index.jsx
```diff
@@ -73,24 +73,40 @@ const BrandSvg = styled.svg`
...
-              focusable="false"
-              aria-hidden="true"
-              ratio={svg.ratio}
-            >
-              {svg.group}
-            </BrandSvg>
-            <VisuallyHiddenText>{brandName}</VisuallyHiddenText>
-          </StyledSpan>
-        </StyledLink>
+      {link ? (
```
12: Why does this conditional exist if `link` should always exist?

12 (changes requested):
Looks good! Got a couple of comments.

packages/components/psammead-brand/CHANGELOG.md
12: Should it definitely be a required prop?

packages/components/psammead-brand/src/index.jsx
ghost: It does reduce the clickable area. Nice catch. Fixed it.

packages/components/psammead-brand/src/index.jsx
ghost: According to this comment. If the link is not provided in any case we should just display the branding that doesn't link to anywhere.
https://github.com/bbc/psammead/issues/500#issuecomment-492656690

packages/components/psammead-brand/README.md
ghost: How about `url`

packages/components/psammead-brand/CHANGELOG.md
ghost: You are right. Since `Brand` doesn't have a default `url` it links to.

packages/components/psammead-brand/src/index.test.jsx
```diff
@@ -18,10 +18,20 @@ const svg = {
 describe('Brand', () => {
   shouldMatchSnapshot(
     'should render correctly',
-    <Brand brandName="Default Brand Name" svg={svg} />,
+    <Brand
+      brandName="Default Brand Name"
+      svg={svg}
+      link="https://www.bbc.co.uk/news"
```
12:
```suggestion
      url="https://www.bbc.co.uk/news"
```

andrew-nowak (commented):
seems good to me, but looks like it will have major merge conflicts with #480

ghost:
We can wait for #480 to be merged, then resolve the conflicts.

packages/components/psammead-brand/src/index.jsx
```diff
@@ -73,24 +74,40 @@ const BrandSvg = styled.svg`
...
-              aria-hidden="true"
-              ratio={svg.ratio}
-            >
-              {svg.group}
-            </BrandSvg>
-            <VisuallyHiddenText>{brandName}</VisuallyHiddenText>
-          </StyledSpan>
-        </StyledLink>
+    <ConstraintWrapper hasLink={url}>
+      {url ? (
```
dr3: We shouldn't really have logic expecting `url` to not be there if it's required. We should either make url not required, or expect it to be there.

dr3: We can write code to be defensive to errors for require props, e.g. if a required prop is missing, dont throw an error, but that should fail safe (e.g. null or render with missing value), not fail to an expected alternative output

ghost: Good call. Both you and @rhell are right. Since there is an alternative output, it makes sense to make `url` an optional prop.

packages/components/psammead-brand/src/index.stories.jsx
```diff
@@ -18,10 +18,19 @@ storiesOf('Brand', module)
         .map(key => key.charAt(0).toUpperCase() + key.slice(1));
 
       const choice = select('Service SVG', options, 'news').toLowerCase();
-      return <Brand brandName="Default Brand Name" svg={svgs[choice]} />;
+      return (
+        <Brand
```
12, line 40: Can we name this story something a bit more descriptive? `with brand link` should suffice. 

This story doesn't reflect the default behaviour of the component, as the default behaviour is when `url` is null, which then renders a different component.

packages/components/psammead-brand/src/index.test.jsx
```diff
@@ -18,10 +18,20 @@ const svg = {
 describe('Brand', () => {
   shouldMatchSnapshot(
     'should render correctly',
```
12:
```suggestion
    'should render correctly with link provided',
```

12 (changes requested):

12 (dismissed):
LGTM 👍 Thanks for making those changes

DenisHdz (dismissed):
Great job 👍

packages/components/psammead-brand/README.md
```diff
@@ -4,13 +4,13 @@
 
 The `Brand` component provides the BBC service logo (as SVG), nested inside a styled link and div. The link is currently hardcoded to "https://www.bbc.co.uk/news".
 
-`Brand` takes a `brandName`, `svgHeight`, `minWidth`, `maxWidth` and  `svg` as props. 
+`Brand` takes a `brandName`, `svgHeight`, `minWidth`, `maxWidth` and `svg` as props.
```
dr3: Can you update the readme to say about `url` please

packages/components/psammead-brand/src/index.jsx
```diff
@@ -66,30 +71,68 @@ const BrandSvg = styled.svg`
   /* stylelint-enable */
 `;
 
-const Brand = ({ brandName, svgHeight, minWidth, maxWidth, svg }) => (
-  <Banner svgHeight={svgHeight}>
+/* eslint-disable react/prop-types */
+const StyledBrand = ({ brandName, svgHeight, svg, maxWidth }) => (
```
dr3: `StyledLink` is a styled-components `styled.a` element, but `StyledBrand` is a React element. Can we ensure consistancy, if we need to we should only prefix directly styled-components objects as `Styled`

packages/components/psammead-brand/src/index.jsx
```diff
@@ -66,30 +71,68 @@ const BrandSvg = styled.svg`
...
+  maxWidth,
+  svg,
+  url,
+}) => (
+  <StyledLink href={url} maxWidth={maxWidth} minWidth={minWidth}>
+    <StyledBrand brandName={brandName} svg={svg} svgHeight={svgHeight} />
+  </StyledLink>
+);
+
+const Brand = ({ brandName, svgHeight, minWidth, maxWidth, svg, url }) => (
```
dr3, line 113: This is a bit tricky to easy scan, as it splits off into BrandWithLink, could we maybe just do this?
```jsx
const Brand = props => {
  const { svgHeight } = props;

  return (
    <Banner svgHeight={svgHeight}>
      {url ? (
        <StyledLink {...props}>
          <StyledBrand {...props} />
        </StyledLink>
      ) : (
        <StyledBrand {...props} />
      )}
    </Banner>
  );
};
```

dr3 (changes requested):
Just a couple things, good work

packages/components/psammead-brand/src/index.jsx
```diff
@@ -66,30 +71,68 @@ const BrandSvg = styled.svg`
...
           {svg.group}
         </BrandSvg>
         <VisuallyHiddenText>{brandName}</VisuallyHiddenText>
-      </StyledLink>
+      </Fragment>
+    )}
+  </Fragment>
+);
+
+/* eslint-disable react/prop-types */
```
dr3: Also, it can sometimes be ok to disable requiring prop types in storybook, but in prod we should always do it really, even on nested components

packages/components/psammead-brand/src/index.jsx
ghost: Thanks. That is way much nicer. 👍. Fixed

packages/components/psammead-brand/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
+| 3.0.0   | [PR#528](https://github.com/bbc/psammead/pull/528) Add link prop to allow dynamic link |
```
dr3, line 6: Should this be alpha?

ghost: I don't think so. Why would it be alpha?

dr3: It hasnt been swarmed since Phils refactor

ghost: A swarm was carried out on https://github.com/bbc/psammead/pull/480#issuecomment-493459076

dr3: Ahh, curse me being on leave :D haha thanks

packages/components/psammead-brand/src/index.jsx
```diff
@@ -105,4 +107,39 @@ Brand.propTypes = {
...
+const Brand = ({ brandName, svgHeight, minWidth, maxWidth, svg, url }) => (
+  <Banner svgHeight={svgHeight}>
+    {url ? (
+      <StyledLink href={url} maxWidth={maxWidth} minWidth={minWidth}>
+        <StyledBrand
+          brandName={brandName}
+          svg={svg}
+          maxWidth={maxWidth}
+          minWidth={minWidth}
+          url={url}
```
dr3: I dont think brand needs url

dr3: or min width by the looks of it, can you review this file for needed prop types please

packages/components/psammead-brand/src/index.jsx
dr3: This has been resolved but its not changed?

dr3 (dismissed):
LGTM, nice work

packages/components/psammead-brand/src/__snapshots__/index.test.jsx.snap
```diff
@@ -1,6 +1,86 @@
...
+}
+
+.c1 {
+  box-sizing: content-box;
+  fill: #FFFFFF;
+  padding-top: 1rem;
+  padding-bottom: 0.75rem;
+  width: 100%;
+  max-width: NaNrem;
+  min-width: NaNrem;
```
andrew-nowak: `max-width` and `min-width` are set to `NaNrem`

packages/components/psammead-brand/src/__snapshots__/index.test.jsx.snap
```diff
@@ -32,6 +112,8 @@ exports[`Brand should render correctly 1`] = `
   padding-top: 1rem;
   padding-bottom: 0.75rem;
   width: 100%;
+  max-width: 17.5rem;
+  min-width: NaNrem;
```
andrew-nowak: `min-width` set to `NaNrem` here

andrew-nowak (changes requested):
Looks mostly good to me, just some `NaNrem`s appearing in the css output, which should probably be avoided.

andrew-nowak (approved):

dr3 (approved):

jamesbrumpton:
LGTM. Happy for this to be merged.

PR #535, opened 2019-05-17
Fix broken story

andrew-nowak:
Resolves #NUMBER

**Overall change:** When entering the psammead storybook for the first time, you are presented with two knobs, `Image alt text` and `Image src`, which do not exist on the viewed story. These are because knob functions on the **Story Promo** story are called too early. Instead, only call the knob functions when the story is being rendered.

image
> These knobs don't affect this story!

Once this PR is merged, `Image alt text` and `Image src` knobs will be visible only on the Story Promo default story, and will control the image appearing on that story.

**Code changes:**

- Convert the static `ImageComponent` into a `buildImageComponent` function, which is called at the intended time.
- Use more representative example of alt text

---

- N/A Tests added for new features
- [ ] Test engineer approval

### DIFF
--- a/packages/components/psammead-story-promo/src/index.stories.jsx
+++ b/packages/components/psammead-story-promo/src/index.stories.jsx
@@ -7,9 +7,9 @@ import Timestamp from '@bbc/psammead-timestamp';
 import notes from '../README.md';
 import StoryPromo, { Headline, Summary } from './index';
 
-const ImageComponent = (
+const buildImageComponent = () => (
   <Image
-    alt={text('Image alt text', '2019-03-01T14:00+00:00')}
+    alt={text('Image alt text', 'An example of image alt text')}
     src={text(
       'Image src',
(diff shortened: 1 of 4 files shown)


DenisHdz (dismissed):
LGTM 👍

packages/components/psammead-story-promo/src/index.stories.jsx
```diff
@@ -7,7 +7,7 @@ import Timestamp from '@bbc/psammead-timestamp';
 import notes from '../README.md';
 import StoryPromo, { Headline, Summary, Link } from './index';
 
-const ImageComponent = (
+const buildImageComponent = () => (
```
ghost: Could we define `ImageComponent` close to where it is actually used in the `story` instead of creating an extra function at the top to build the component.

ghost (approved):
LGTM 👍

DenisHdz (approved):

david-boydell:
Cool, looks good

PR #547, opened 2019-05-22
Add Typography story in Storybook

### LINKED ISSUE #363: Typography in Storybook
We currently don't make the typography values from the `gel-foundations` package visible in Storybook. We should. This will enable UX designers, Product owners, Testers to have a view of the different types. 

To add a story to the `psammead-styles` package under typography that shows each of the Types (e.g. `GEL_LONG_PRIMER`)

These currently only handle the `latin` values.  

When the values are extended to have non-latin values (#342), we should be able to use a Storybook knob to change the script, from `latin` to `latin-with-diacritics` for example, to see the line-height and font-size to change accordingly by using the settings

N/A

**Testing notes**
[Tester to complete]

Dev insight: Review should be showing UX designers the Storybook story. 

Bopchy:
Resolves #363 

**Overall change:** _Create Typography story so that the typography values from the @bbc/gel-foundations package are visible in Storybook._

**Code changes:**

- _Create TypographyText component to render paragraph text with varying typography styles and scripts._
- _Create a Typography story in Storybook with inputProvider knob - allowing different scripts to be chosen and viewed._
- _Add test for TypographyText component._
- _Update CHANGELOG.md_
- _Update package version_
---

- [ ] Test engineer approval


packages/utilities/psammead-styles/src/typography.stories.jsx
```diff
@@ -0,0 +1,74 @@
...
+import TypographyText from '.';
+
+const script = latin;
+const sampleText =
+  'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.';
+
+storiesOf('Typography', module)
+  .addDecorator(withKnobs)
+  .add('Canon', () => (
+    <TypographyText script={script} typographyFunc={getCanon}>
```
sareh: Looks great so far. Adding the inputProvider to switch different scripts via the drop-down list would be great:
e.g. like the Paragraph story example: https://bbc.github.io/psammead/?path=/story/paragraph--default
https://github.com/bbc/psammead/blob/5700a49222fffef327bf73a157f686367b4064e0/packages/components/psammead-paragraph/src/index.stories.jsx#L12

Bopchy: 👍 Done

pjlee11 (dismissed):
Great work! 👍

packages/utilities/psammead-styles/src/typography.jsx
```diff
@@ -0,0 +1,17 @@
+import { shape, func } from 'prop-types';
```
ghost: Could we move this file to `typography.stories.jsx` since it is a helper component.

packages/utilities/psammead-styles/src/typography.jsx
```diff
@@ -0,0 +1,17 @@
...
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+
+const TypographyText = styled.p`
+  ${props =>
+    props.script && props.typographyFunc
+      ? props.typographyFunc(props.script)
+      : ''}
+`;
+
+TypographyText.propTypes = {
```
ghost: After moving it. Here we can safely ignore the proptype checks with
`/* eslint-disable react/prop-types */`

ghost (changes requested):
Looks great. Just a few things.

packages/utilities/psammead-styles/src/typography.jsx
Bopchy: My thinking behind having the TypographyText component in typography.jsx was to separate concerns. Personally, I think this file should remain as is

ghost: I saw that we already have test covering typography in https://github.com/bbc/psammead/blob/e42de7c76b31851de995a6823c34c7b0be90f636/packages/utilities/gel-foundations/tests/getTypeSizes.test.js#L5-L25

By having `typography.test.jsx` we would be duplicating the above test that ensures typography function works as expected.
This means we won't need `typography.test.jsx`. If we delete that file, now `typography.jsx` will only be needed in `typography.stories.jsx`. Thats why i made the recommendation. But feel free to ignore it.

Bopchy: 👌Changed it.

packages/utilities/psammead-styles/src/typography.stories.jsx
```diff
@@ -0,0 +1,51 @@
...
+const typographyStory = typographyFunc =>
+  inputProvider([{ name: 'sample text' }], ([text], script) => (
+    <TypographyText script={script} typographyFunc={typographyFunc}>
+      {text}
+    </TypographyText>
+  ));
+
+storiesOf('Typography', module)
+  .addDecorator(withKnobs)
+  .add('Canon', typographyStory(getCanon))
```
dr3: If you do 
```import * as typographies from '@bbc/gel-foundations/typography';```

You can then do 
```
const stories = storiesOf('Typography', module)
  .addDecorator(withKnobs)
```

then
```
Object.keys(typographies).forEach(typographyName => {
	stories.add(typographyName, typographyStory(typographies[typographyName]))
});
```

IMO its a bit nicer, and means when we add more, we dont have to manually add them here :)

**Comment by DenisHdz** [comment]
What about if instead of creating a story for each of the Typographies, we just create one and display all of them at once? Similar to what I have done with the [Colours](https://bbc.github.io/psammead/?path=/story/psammead-styles--colours). Thoughts?

**Comment by ghost** [comment]
@DenisBBC in this case, the approach in the colours would not be the best. It would mean we have a sample text and its identifier.
```
Canon: Sample text that is affected by the typography size
Pica : Another sample text
``` 
Which will become confusing if we want to add a knob to show the sizes in other scripts like `persian`
and switch the `sample text` at the same time.

Bopchy:
I have made changes to display all the typography stories at once - using the inputProvider though

j-pendlebury (dismissed):
LGTM 👍

oluoluoxenfree (dismissed):
LGTM 💯

ghost (changes requested):
Looks great but you forgot to add the `psammead-styles` notes on the typography story.

ghost (dismissed):

j-pendlebury (dismissed):

ghost (approved):
👍

oluoluoxenfree (approved):

jamesbrumpton:
LGTM. Happy for this to be merged.

PR #571, opened 2019-05-31
Add font families for services

ghost:
Resolves #NUMBER

**Code changes:**

---

- [ ] Test engineer approval


packages/utilities/psammead-styles/src/font-faces.js
```diff
@@ -0,0 +1,94 @@
...
+    font-display: optional;
+  }`,
+};
+
+export const news = {
+  sansBold: `@font-face {
+    font-family: "ReithSans";
+    font-weight: 700;
+    font-style: normal;
+    src: url('${baseFontUrl}ReithSans/latest/bold.woff2') format('woff2'), url('${baseFontUrl}ReithSans/latest/bold.woff') format('woff');
```
dr3: We should continue to use the gel base url for news, it has better cross over with other UK sites like sport

PR #573, opened 2019-06-04
Add functions to get font styles

### LINKED ISSUE #539: Update font config to be dynamic per service
BLOCKED ON UX MAPPING CONFIG

Update our [existing font families](https://github.com/bbc/psammead/blob/latest/packages/utilities/gel-foundations/src/typography.js#L4-L8) and manual weight and styling to be a function, dynamically created per service

Create 6 functions, in the format of `getSerifMedium(service)` which takes in service name for each service and exports css

Functions:
```
getSerifMedium('persian')
getSerifMediumItalic('persian')
getSansRegular('persian')
getSansItalic('persian')
getSansBold('persian')
getSansBoldItalic('persian')
```

Example behaviour
`getSerifMediumItalic('news')` returns
```
css`
font-family: ReithSerif, Helvetica, Arial, sans-serif;
font-weight: 500;
font-style: italic;
`
```
`getSerifMediumItalic('persian')` returns
```
css`
font-family: NassimPersian, Arial, Verdana, Geneva, Helvetica, sans serif;
font-weight: 700;
font-style: normal;
`
```

Depending on how complicated to config gets, we may decide that instead of passing in `service` we export `config` for each service which gets passed into the function. This is done to avoid bloating other services with weight. When implementing this, once you have a good idea of the size of the config, chat to the team and see what they think.

This can probably go in `@bbc/psammead styles` as it pairs closely with [the fontfaces used](https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-styles/src/fonts.js#L4). We can thus remove the font-family stuff from `@bbc/gel-foundations` [here](https://github.com/bbc/psammead/blob/latest/packages/utilities/gel-foundations/src/typography.js#L4-L8)

 
**Testing notes**
No testing

Dev insight: Unit tests

ghost:
Resolves #539 

**Overall change:** Add function that allow getting font family given a script name.
**Code changes:**

- Add font family config file that contains the font families
- Add function that enable getting of font family given a script name.
- Add tests for the function created.

---

- [ ] Test engineer approval


packages/utilities/psammead-styles/src/font-styles.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const getSansItalic = service => {
+  if (!fonts[service]) {
+    return undefined;
+  }
+  const { sansItalic, sansRegular } = fonts[service];
+  return sansItalic || sansRegular;
+};
+
+export const getSansBold = service => fonts[service] && fonts[service].sansBold;
```
dr3: this fails if sans bold isnt defined, should it fall back to regular?

ghost: yeah I believe it should.

packages/utilities/psammead-styles/src/font-styles.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+const getReithSerif = (weight, style) => `
+   font-family: ReithSerif, Helvetica, Arial, sans-serif; 
+   font-style: ${style};
+   font-weight: ${weight};
+  `;
+
+describe('fonts', () => {
+  describe('getSansRegular', () => {
+    it('should return regular sans font css', () => {
```
dr3: This file seems to only test some of the exports, is it possible to maybe make this a bit easier to read, maybe use config to auto generate the tests? I want to ensure full coverage, but as we get more and more services, verbose tests like this could become weildy. Could we maybe pass a looping test config?

dr3 (changes requested):

packages/utilities/psammead-styles/src/font-styles.test.js
ghost: Which export are not being tested? This file is supposed to test the exported functions which are all being tested. By all exports do you mean the values of the services?

dr3: e.g. getSansRegular on yoruba, news, pidgin etc

ghost: But this would be testing the values of the services not the functions. How do you feel about having a separate test file for `font-families`. And test that the services have the correct values. Because given the correct font values for the services, the functions would work as expected.

dr3 (commented):
(ignore this)

dr3 (changes requested):
I wrote a test that I think we should use, it allows for easy output testing to ensure outputs are maintained, for very little testing code

```
import * as fontStyles from './font-styles';

['news', 'persian', 'igbo', 'yoruba', 'pidgin'].forEach(service => {
	Object.keys(fontStyles).forEach(fontStyle => {
		it(`should render ${fontStyle} correctly for ${service}`, () => {
			expect(fontStyles[fontStyle](service)).toMatchSnapshot();
		})
	})
})
```

Thoughts? I think this could be in additon to the existing tests, as an assurance of output for the config

~There do appear to be some issues in the mapping however, like `getSansItalic` which appears to return undefined for persian~

Just realised this PR is missing persian, which we should add before merge

ghost:
@drew It is a good addition. 👍 . Persian is not a priority for v1.

packages/utilities/psammead-styles/src/__snapshots__/font-styles.test.js.snap
```diff
@@ -0,0 +1,193 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dr3, line 1: Rather love this snapshot file, soooo easy to read :P

packages/utilities/psammead-styles/src/font-styles.js
```diff
@@ -0,0 +1,44 @@
+import * as fonts from './font-families';
+
+export const getSansRegular = service =>
+  fonts[service] && fonts[service].sansRegular;
+
+export const getSansItalic = service => {
+  if (!fonts[service]) {
+    return undefined;
```
dr3: Usually i would prefer returning null, but im not too bothered

dr3 (dismissed):
LGTM, nice work

After this is approved, before merge can you get someone from UX to review `packages/utilities/psammead-styles/src/__snapshots__/font-styles.test.js.snap` please :)

packages/utilities/psammead-styles/src/font-styles.js
DenisHdz: Me too. @sadickisaac Can we replace them? Otherwise lgtm :)

ghost: I was thinking more of what would `fonts['service']` return given and invalid `service` name. It would be `undefined` not `null`. Changed it though.

DenisHdz (dismissed):
Awesome work 💯

j-pendlebury (dismissed):
looks awesome 👍

DenisHdz (dismissed):

j-pendlebury (dismissed):

packages/utilities/psammead-styles/src/font-families-config.js
```diff
@@ -0,0 +1,9 @@
+const helmet = 'Helmet, Freesans, Helvetica, Arial, sans serif';
```
dr3: Am I missing something? What happened to ReithSerif?

dr3 (changes requested):

packages/utilities/psammead-styles/src/font-families-config.js
ghost: Am looking at two ways of approaching this. 
1. We have `latin` return and object with `sans` and `serif`. But then that would be inconsistent because other scripts return a string.
2. We add `script` names `latinSans` and `latinSerif` and then later on in the simorgh service config add a script name key which will have an array of script names. Then let containers handle which script name is passed to a component

packages/utilities/psammead-styles/src/font-families-config.js
```diff
@@ -0,0 +1,12 @@
...
+const reithFallback = ', Helvetica, Arial, sans-serif';
+const reithSans = `ReithSans${reithFallback}`;
+const reithSerif = `ReithSerif${reithFallback}`;
+
+const fontFamilies = {
+  latinAndCyrillic: helmet,
```
dr3: latinSans and latinSerif arn't scripts, [the list is here](https://github.com/bbc/psammead/blob/latest/packages/utilities/gel-foundations/src/scripts.js)

Every script needs the ability of requesting a sans and serif font family, But also, I dont think it should be based on script, but service. 2 services could use the same script, but for data reasons we dont want to include a custom font.

dr3 (changes requested):

packages/utilities/psammead-styles/src/font-families-config.js
ghost: This PR just covers the definition of the font-families and fallbacks. The decision of wether to load custom fonts or not I think would be done in simorgh.

packages/utilities/psammead-styles/README.md
```diff
@@ -46,6 +49,16 @@ const SomeStyledComponent = css`
 `;
 ```
 
+### font-families
+
+```js
+import { getSansFontFamily } from '@bbc/psammead-styles/font-font-families';
+
+const SomeStyledComponent = css`
+  font-family: ${({ scriptName }) => getSansFontFamily(scritpName)};
```
dr3: What is `scriptName`? also `scritpName` spelling

packages/utilities/psammead-styles/src/font-families-config.js
```diff
@@ -0,0 +1,17 @@
+const helmet = 'Helmet, Freesans, Helvetica, Arial, sans serif';
+const reithFallback = ', Helvetica, Arial, sans-serif';
+const reithSans = `ReithSans${reithFallback}`;
+const reithSerif = `ReithSerif${reithFallback}`;
+
+const fontFamilies = {
+  latinAndCyrillic: {
```
dr3: Can you please give examples of this working IRL, latinAndCyrillic isnt [one of the script names exported](https://github.com/bbc/psammead/blob/latest/packages/utilities/gel-foundations/src/scripts.js)

dr3 (changes requested):

dr3:
@sadickisaac Can we sit down with UX at some point to review this? I much preferred your implementation before https://github.com/bbc/psammead/pull/573/files/6af32d6480eff7285f7803e5bfb8508700684b13 and now i'm insure this suits our needs, why did you redo the PR, did a criteria change?

You changed from mapping on service name to mapping on script, which simply wont work. News and Yoruba use the exact same script, but different font families, why did you change to using this? Was it a UX decision? This issue specifically says `Create 6 functions, in the format of getSerifMedium(service) which takes in service name for each service and exports css` because of this

dr3 (approved):
LGTM, can you make sure to go over `packages/utilities/psammead-styles/src/__snapshots__/font-styles.test.js.snap` with UX before merge to check the functions act as expected please :)

DenisHdz (approved):
LGTM 👍

WeimingFan:
Tom and Fan have reviewed the part of the code we understood (the font fallbacks), and it looks fine to us.

david-boydell:
Hi @sadickisaac, is there any representation of this in Storybook? I.e. is there anything for me to verify here? Or should I wait for it to be integrated into Simorgh?

ghost:
@david-boydell There is no representation for this in storybook. Unit tests and confirmation received from UX I think is sufficient. This functions would be used in psammead.

david-boydell:
Static check looks good, this is ready for merge.

PR #604, opened 2019-06-10
Remove nanoid

ryanmccombe:
Resolves (partially) https://github.com/bbc/simorgh/issues/1489

**Overall change:** Remove NanoID dependency used in the Sitewide Links 

**Code changes:**

- This component will now respect an `id` property passed to each link definition
- If not provided, the component will generate a key by stringifying the link, which includes the link text and URL

---

- [ ] Test engineer approval


packages/components/psammead-sitewide-links/src/List/index.jsx
```diff
@@ -68,7 +67,7 @@ const StyledListItem = styled.li`
 const List = ({ links }) => (
   <StyledList role="list">
     {links.map(link => (
-      <StyledListItem key={nanoid()} role="listitem">
+      <StyledListItem key={link.id || JSON.stringify(link)} role="listitem">
```
jroebu14: Based on the [discussion in this PR](https://github.com/bbc/simorgh/pull/1879#discussion_r292405198) I think we are going with just using the text as a key.

jamesbhobbs (dismissed):
I'm not very familiar with the package management in psammead, but I think this will need a semver update on this package. Best to check with the psammead code owners re this though.

dr3:
Yeah a package version bump and changelog update is needed :) Also can you reset `package-lock.json` please at the base of the repo, not sure whats happened there :P

jamesbhobbs (approved):

dr3 (approved):

PriyaKR:
The sitewide links breaks when you change language service ie) if i try changing the service it still has the links from the previous selected service showing along with the current ones.
sitewidelinks_issue

ryanmccombe:
eek - unfortunately one of the assumptions here is that the text of every link is unique, and that wasn't true of the storybook sample data

I've updated the sample data to make every link unique (just adding a number to the end of it)

PriyaKR:
@ryanmccombe Seeing an issue the copyright text doesn't seem to display for default and visually hidden text stuff too.Issue is not on the latest branch.

ryanmccombe:
> @ryanmccombe Seeing an issue the copyright text doesn't seem to display for default and visually hidden text stuff too.Issue is not on the latest branch.

@priya is it this one?  https://bbc.github.io/psammead/?path=/story/components-copyright--default

I can't reproduce that bug at all - the copyright looks the same on this branch as on latest for me - mind trying again in case it was some CI thing?

It would be very strange if this code changed the copyright component, as the only code changes here are in the sitewide links component

PriyaKR:
Regression looks good to me and also the link text is being used for key generation in a list ie)sitewide links.

PR #634, opened 2019-06-14
Fix the command injection vulnerability is getRegistryVersion.js

### LINKED ISSUE #526: Fix the command injection vulnerability in getRegistryVersion.js
**Describe the bug**
This is related to #525. It was noticed that the original [script](https://github.com/bbc/psammead/blob/12b27c2e3e1f1a42e2997bc653646184bade61a9/scripts/publish/src/getRegistryVersion.js#L6) uses `exec` to view the versions of the npm module. The problem is however that the script also concatenates with `name` string. This is an example of a serious command injection vulnerability. In this case an attacker can inject a custom package name to execute commands remotely.

Thankfully this is a build script and so this does not endanger are users directly. And the fix for this is actually quite simple. 

**First**: find out why we have custom scripts for publishing to NPM and if they are actually still required or could we simplify this.

If they **are** still required:

Solution:
The solution to this problem is actually quite simple: use `spawn` instead of exec. This is detailed here: https://nodejs.org/docs/latest-v10.x/api/child_process.html
The reason why `spawn` is better is that it enforces a split between the command and the parameters being passed in. This will **NOT** solve the vulnerability but it will mitigate its worst offenders. 

To solve this problem the script needs to pass the name input through some sort of whitelist. The psammead packages do have names that start with `psammead-*` and `@bbc` creating a very simple regex whitelist for package names would solve this vulnerability completely. This means that a whitelist would use the following rules:
- Name starts with `psammead`
- Name starts with `@bbc`
- Name starts with `gel`
- Name should not contain spaces
- Name should use `-` as a delimiter

A strict whitelist will enforce us developers to use specific naming conventions for our packages. [Which is a small price to pay for salvation](https://i.redd.it/utb2u8d8o9711.jpg). While annoying it will defend us from this type of vulnerabilities.

DemonusPC:
Resolves #526 

**Overall change:** 
Added a whitelist check for packages names that start either with
```
psammead-*
gel-*
```
And which contain only alphanumeric and the `-` character. 
If the `name` parameter does not pass the regex the function will immedietly return with a `-1` and the `exec` function will not be called. 

**Code changes:**

- Added a regex expression that locks down the available strings that can be passed in:
  - `/^psammead-[a-zA-Z0-9-]*$|^gel-[a-zA-Z0-9-]*$/`
- Added a function to verify the string against the regex
- Added an if statement to check if the name is `valid`. The function will immedietly return if this circuitbreaker is triggered.
---

- [ ] Test engineer approval


ryanmccombe (approved):

jamesbhobbs (approved):

jamesbhobbs:
Not testable

PR #643, opened 2019-06-17
Update timestamp container to accept locale as prop

### LINKED ISSUE #613: Psammead timestamp container - Accept and use moment locale
We need to use different moment locales for different services

Accept a 'locale' prop, and tell moment to use it. This makes the assumption that the locale is available to the package. 

Moment uses a `moment.locale('fa')` api to set the locale, which we can do once at the start of the component.

**Testing notes**
[Tester to complete]

Dev insight: Unit tests, storybook examples

**Additional context**

ghost:
Resolves #613

**Overall change:** Update timestamp container to accept locale as prop
**Code changes:**

- Add optional locale prop.
- Set locale using moment.

---

- [ ] Test engineer approval


packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -55,6 +62,7 @@ TimestampContainer.defaultProps = {
   padding: true,
   prefix: null,
   suffix: null,
+  locale: null,
```
j-pendlebury, line 65: what if the locale is null (if we don't pass it in), does it fail nicely without breaking the component or should we set the default to like `en-gb`?

j-pendlebury (changes requested):
Just one question :)

packages/containers/psammead-timestamp-container/src/index.jsx
dr3: Should do, due to the `if (locale) {`

packages/containers/psammead-timestamp-container/README.md
```diff
@@ -22,8 +26,10 @@
 | prefix | string | No | `null` | `Updated` |
 | suffix | string | No | `null` | `This is a suffix` |
 | script | object | Yes | N/A | { canon: { groupA: { fontSize: '28', lineHeight: '32',}, groupB: { fontSize: '32', lineHeight: '36', }, groupD: { fontSize: '44', lineHeight: '48', }, }, trafalgar: { groupA: { fontSize: '20', lineHeight: '24', }, groupB: { fontSize: '24', lineHeight: '28', }, groupD: { fontSize: '32', lineHeight: '36', }, }, } |
+| locale | string | no | `null` | `en` |
```
dr3, line 29: Can you add text somewhere that this is specifically the `moment.js` locale please

dr3 (changes requested):

packages/containers/psammead-timestamp-container/src/index.jsx
ghost: if the locale is null, we don't set it. I think that is handled with
```js
if(locale) {
}
```
Moment defaults to `en` if no locale is defined.

packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -18,11 +19,16 @@ const TimestampContainer = ({
...
   timezone,
   script,
+  locale,
 }) => {
   if (!isValidDateTime(new Date(timestamp))) {
     return null;
   }
 
+  if (locale) {
+    moment.locale(locale);
```
andrew-nowak, line 29: Have you considered passing the locale into `formatUnixTimestamp` and only setting the moment locale locally? http://momentjs.com/docs/#/i18n/instance-locale/

ghost: Thats a good idea. 👍

packages/containers/psammead-timestamp-container/src/index.stories.jsx
```diff
@@ -13,6 +13,7 @@ storiesOf('Containers|TimestampContainer', module)
       format="D MMMM YYYY"
       isRelative={boolean('isRelative', false)}
       script={latin}
+      locale="ig"
```
dr3: Can you use a locale like `fa`, and add a new story please. Would be good to have examples of different locales in storybook

packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -18,17 +18,21 @@ const TimestampContainer = ({
...
+
   return (
-    <Timestamp
-      datetime={formatUnixTimestamp(timestamp, dateTimeFormat, timezone)}
-      padding={padding}
-      script={script}
-    >
+    <Timestamp datetime={datetime} padding={padding} script={script}>
       {prefix ? `${prefix} ` : null}
       {showRelativeTime(timestamp, isRelative, format, timezone)}
```
dr3: Do you not need to pass locale to showRelativeTime?

dr3: TBH I prefer setting it once at the top of the container, avoids issues of having to pass it around everywhere

dr3 (changes requested):

packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -55,6 +60,7 @@ TimestampContainer.defaultProps = {
   padding: true,
   prefix: null,
   suffix: null,
+  locale: 'en',
```
dr3: I think it should default to null, as we use `en-gb` in psammead so it defaults to whatevers availible. i think if you give it `en-gb` but ask for `en`, it will use `en-gb` but throw a warning.

packages/containers/psammead-timestamp-container/src/index.stories.jsx
```diff
@@ -35,4 +35,14 @@ storiesOf('Containers|TimestampContainer', module)
...
     />
+  ))
+  .add('with locale', () => (
+    <Timestamp
+      timestamp={number('Unix timestamp', 1530947227000)}
+      dateTimeFormat="YYYY-MM-DD"
+      format="D MMMM YYYY, HH:mm z"
+      isRelative={boolean('isRelative', false)}
+      script={latin}
+      locale={text('Locale', 'en')}
```
dr3:
```suggestion
      locale={text('Locale', 'fa')}
```

packages/containers/psammead-timestamp-container/src/index.test.jsx
```diff
@@ -6,6 +6,7 @@ import Timestamp from '.';
 const defaultTimestamp = 1539969006000; // 19 October 2018
 const noLeadingZeroTimestamp = 1530947227000; // 07 July 2018
 const invalidData = '8640000000000001'; // A day holds 86,400,000 milliseconds - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#Description
+const igboLocale = 'ig';
```
dr3: Igbo isnt a standard moment locale, so it defaults to english, if you use one like `fa`, the snapshot will reflect the translation

dr3 (approved):

DenisHdz (approved):
Great job 👍🏻

david-boydell:
Tested with _fa_ and _yo_, this is ready for merge,

j-pendlebury (approved):

packages/containers/psammead-timestamp-container/src/__snapshots__/index.test.jsx.snap
```diff
@@ -30,10 +30,10 @@ exports[`Timestamp should add prefix and suffix 1`] = `
 
 <time
   className="c0"
-  dateTime="2018-10-19"
+  dateTime="۲۰۱۸-۱۰-۱۹"
```
sareh, line 33: `dateTime` should always be using these numerals `"2018-10-19"`. It's because this field is used by search engines. We should not localise this field for Persian and other services. Only the visible timestamp should be localised.

PR #666, opened 2019-06-20
Timestamp container datetime bug

### LINKED ISSUE #651: Psammead Timestamp Container - Dont localise datetime
**Describe the bug**
We shouldn't localise the time elements dateTime value since this is used by search engines in the standard latin format

Current
```
<time datetime="۲۰۱۸-۰۷-۰۷">۷ ژوئیه ۲۰۱۸، ۰۸:۰۷ BST</time>
```

**Expected behaviour**
When you translate a timestamp, the dateTime attribute remains in latin numerals

**Solution:**
For the dateTime attribute, add `moment.locale('en')` to the moment call. `en` is the default and fallback translation so will always be available, even when the only locale we include in a service is their own.

**Additional context**

j-pendlebury:
Resolves #651 

**Overall change:** 
Updated Timestamp Container to not localise the value of `datetime`.

**Code changes:**
- Updated the `formatUnixTimestamp` in `timestampUtilities` to set the locale to default to 'en' unless other value specified
- add `locale` to `showRelativeTime` function to localise the timestamp visible on the page
- updated snapshots

---

- [ ] Test engineer approval

### DIFF
--- a/packages/containers/psammead-timestamp-container/src/index.jsx
+++ b/packages/containers/psammead-timestamp-container/src/index.jsx
@@ -36,7 +36,7 @@ const TimestampContainer = ({
       script={script}
     >
       {prefix ? `${prefix} ` : null}
-      {showRelativeTime(timestamp, isRelative, format, timezone)}
+      {showRelativeTime(timestamp, isRelative, format, timezone, locale)}
       {suffix ? ` ${suffix}` : null}
     </Timestamp>
   );
--- a/packages/containers/psammead-timestamp-container/CHANGELOG.md
+++ b/packages/containers/psammead-timestamp-container/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 1.2.1 | [PR#TBC](https://github.com/bbc/psammead/pull/TBC) Do not localise `datetime` |
 | 1.2.0 | [PR#643](https://github.com/bbc/psammead/pull/643) Add support for passing locales as prop |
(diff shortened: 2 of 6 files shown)


andrew-nowak (approved):

ghost (approved):

DenisHdz (approved):

packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -36,7 +36,7 @@ const TimestampContainer = ({
       script={script}
     >
       {prefix ? `${prefix} ` : null}
-      {showRelativeTime(timestamp, isRelative, format, timezone)}
+      {showRelativeTime(timestamp, isRelative, format, timezone, locale)}
```
dr3, line 39: Sorry, gunna be a huge pain, can you flip this please? Pass 'en' on line 34 to override the setting on line 29.

This way, we set to the correct locale, then pass the same locale again, defaulting to en. Ide rather we set the locale once (like on line 29), then only override if we want it to be different, like 'en'

dr3 (changes requested):

packages/containers/psammead-timestamp-container/src/index.jsx
j-pendlebury: I might be going mad, but I'm not sure if I understand. Currently, `formatUnixTimestamp` defaults to `en` if the locale is not passed in which works for `datetime` value, but in that same function we need to be able to override the value of `locale` because of line 39 and 42 in `timestampUtilities`.

dr3: Its ok, its ok as it is :)

dr3 (approved):

staylos92:
LGTM 👍

PR #703, opened 2019-06-25
Add sizes support to psammead image

dr3:
Contributes to https://github.com/bbc/simorgh/issues/2022

**Overall change:** Add support to use `sizes` attribute in img and amp-img

**Code changes:**

- Add support to use `sizes` attribute in img and amp-img

---

- [x] Test engineer approval


packages/components/psammead-image/package-lock.json
```diff
@@ -46,7 +46,7 @@
       "dev": true,
       "requires": {
         "jest-styled-components": "^6.3.1",
-        "react-test-renderer": "^16.8.6"
+        "react-test-renderer": "^16.6.3"
```
andrew-nowak: Was this intended? I don't understand why this is getting downgraded here

dr3: Lol no idea, fixed :)

packages/components/psammead-image/src/index.jsx
```diff
@@ -25,8 +25,8 @@ const StyledImg = styled.img`
   ${props => props.fade && fadeIn};
 `;
 
-export const Img = ({ alt, fade, src, srcset, height, width }) => {
-  const props = { alt, fade, src, height, width };
+export const Img = ({ alt, fade, height, sizes, src, srcset, width }) => {
```
ghost: This component can be simplified to
```jsx
const Img = props => {
   const { srcset } = props;
   return <StyledImg {...props} srcSet={srcset} />
}
```

j-pendlebury (approved):

packages/components/psammead-image/src/index.amp.jsx
```diff
@@ -1,13 +1,23 @@
 import React from 'react';
 import { number, string } from 'prop-types';
 
-const AmpImg = ({ alt, attribution, layout, src, srcset, height, width }) => {
+const AmpImg = ({
```
ghost: Instead of `unpacking` props and then `packing` them again, Could we have
```jsx
const AmpImg = props => {
  const { alt /* the rest */ } = props
}
```
Looks much cleaner IMO.

ghost (changes requested):

packages/components/psammead-image/src/index.amp.jsx
ghost: Or simplify to 
```jsx
const AmpImg = props => {
   const { srcset } = props;
   return <amp-img {...props} srcSet={srcset} />
}
```

dr3: > Or simplify to
> 
> ```js-jsx
> const AmpImg = props => {
>    const { srcset } = props;
>    return <amp-img {...props} srcSet={srcset} />
> }
> ```

Probably shouldnt pass the `srcset` prop, could do
```jsx
const { srcset, ...otherProps } = props;
return <amp-img {...otherProps} srcSet={srcset} />
```

But also, being explicit is clearer for prop types and avoids accidentally passing something by mistake :/ not sure

ghost: `prop-types` doesn't guard against passing extra props that are not defined. You can pass in any extra prop that is not defined and it wouldn't even raise a warning.

ghost (approved):

david-boydell:
This is ready for merge.

PR #713, opened 2019-06-26
Update styled components in all packages

ryanmccombe:
Dependency for https://github.com/bbc/simorgh/issues/1910

**Overall change:** Update all the packages that use `styled-components` to use the same, latest, version

**Code changes:**

- Update the package.json, package-log and changelog of every package that used styled components

---

- [ ] Test engineer approval


packages/utilities/psammead-styles/src/font-families.js
```diff
@@ -65,7 +65,7 @@ const helmetFontStyles = {
   `,
 };
 
-const nessimPersian = `font-family: NassimPersian, Arial, Verdana, Geneva, Helvetica, sans serif`;
+const nessimPersian = `font-family: NassimPersian, Arial, Verdana, Geneva, Helvetica, sans serif;`;
```
pjlee11:
```suggestion
const nassimPersian = `font-family: NassimPersian, Arial, Verdana, Geneva, Helvetica, sans serif;`;
```

Also can you check where else this is using the wrong spelling please

pjlee11 (changes requested):
Previous typo in const that we can clear whilst making this change

packages/utilities/psammead-styles/src/font-families.js
ryanmccombe: 👍 searched a repo and "nessim" only appears in this file - updated it to nassim

jroebu14 (approved):

package.json
```diff
@@ -17,7 +17,7 @@
     "storybook": "NODE_ENV=development start-storybook -p 8180 -c .storybook",
     "test": "npm run test:lint && npm run test:lint:css && npm run test:unit && npm run test:packages",
     "test:packages": "lerna run test",
-    "test:lint": "eslint --ext .js,jsx,json ./packages ./scripts",
+    "test:lint": "eslint --ext .js,jsx,json ./packages ./scripts --fix",
```
pjlee11: Are we certain we want the `--fix` here? I'm not convinced everyone will be happy (or no confused) why the code they had saved locally then showed up different in github because the linter autofixed their code on commit because `husky` runs `test:lint` as part of the `pre-commit` hook

chris-hinds: I would rather be told there are linting problems and fix it how I like locally. That might be using `--fix` or that might be me manually fixing it.

So I'd be in favour of removing `--fix`

dr3: imo we shouldn't, --fix will fix issues on CI, leaving them wrong on latest. If we want a command to do --fix, it should be separate to the one CI uses to vet styling IMO

pjlee11: Think that confirms it, let's not `--fix` 

```suggestion
    "test:lint": "eslint --ext .js,jsx,json ./packages ./scripts",
```

pjlee11 (approved):
Nice work 😎

ryanmccombe:
This is ready for QA

Styled components are used in every component in Psammead

The update wasn't denoted as having any breaking changes, but may be worth a quick validation anyway

PriyaKR:
@ryanmccombe Does this PR only fixes only this PR https://github.com/bbc/simorgh/issues/1910 or does it update the styles for other services as well?

Looking at the `psammead-styles` change log i can only see addition of Nassim persian fonts.

PriyaKR:
I have done a regression on the storybook which looks good...all the components using `psammead-styles` are using the latest version `1.0.0` looking at the file changes but the Nassim fonts for persian are not seen for any component as there needs another issue to be fixed for the Persian services top pick up the Nassim fonts.

ryanmccombe:
@PriyaKR this PR doesn't add the Persian fonts - it was just an initial PR, as having the components using different versions of the the `styled-components` library was making work on the Persian ticket (and other tickets) a lot more difficult

https://github.com/bbc/simorgh/issues/1910 will be resolved in a separate PR once this is merged

PriyaKR:
Copying the comment from the dependency PR Note: the Mozart Headers have been deployed up to test (Mozart Routing Release 1240) but not yet to live. So any PR relating to this issue should not be merged until that has been deployed first - otherwise we'll have console errors on the Live environment.

sareh:
@PriyaKR The Mozart Routing work is only relevant for the Simorgh application, not for Psammead. This is because it adds Content Security Policy headers for the World Service font domain. So this PR is not dependent on the Mozart Routing work.

PriyaKR:
thanks @sareh

packages/components/psammead-story-promo-list/CHANGELOG.md
```diff
@@ -3,9 +3,9 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
-| 1.0.0 | [PR#741](https://github.com/BBC/psammead/pull/741) Remove alpha tag |
-| 1.0.0-alpha.1 | [PR#677](https://github.com/BBC/psammead/pull/677) Use `@bbc/gel-foundations@3.0.0` |
-| 1.0.0-alpha.0 | [PR#679](https://github.com/BBC/psammead/pull/679) Bump version number |
-| 0.1.0-alpha.3 | [PR#672](https://github.com/BBC/psammead/pull/672) Add margin to top of unordered list |
-| 0.1.0-alpha.2 | [PR#534](https://github.com/BBC/psammead/pull/534) Remove Timestamp padding. |
-| 0.1.0-alpha.1 | [PR#486](https://github.com/BBC/psammead/pull/486) Create initial package. |
+| 1.0.1   | [PR#713](https://github.com/bbc/psammead/pull/713) Update `styled-components` to 4.3.2 |
```
DenisBBC: This change has overwritten the 1.0.0 version, so we will need to bump again the `Story Promo List` to add another record in the `CHANGELOG`.

DenisHdz, line 6: This change has overwritten the 1.0.0 version, so we will need to bump again the `Story Promo List` to add another record in the `CHANGELOG`.

Issue #276, opened 2019-01-16
Update Psammead snapshots (react-testing-library)


radiocontrolled:
**Is your feature request related to a problem? Please describe.**
After https://github.com/bbc/psammead/issues/273 is complete, the Psammead snapshots should be updated. 

The package.json version numbers will have to be updated. 

- [x] Initially labelled with ["Refinement needed"](https://github.com/BBC-News/psammead/labels/Refinement%20Needed)

radiocontrolled:
Blocked on https://github.com/bbc/psammead/issues/273

PriyaKR:
Associated PR [Use react-testing-library for Simorgh bbc/simorgh#1182](url) is now closed and this mainly requires updating the snapshots using react-testing-library.

jroebu14:
`react-testing-library` has only been implemented in Simorgh. Before picking up this issue we need to complete https://github.com/bbc/psammead/issues/273 to implement `react-testing-library`* in Psammead.

*`react-testing-library` is now called `@testing-library/react`.

chris-hinds:
blocked on #273

PR #732, opened 2019-07-01
use @testing-library/react to generate snapshots

### LINKED ISSUE #273: Use react-testing-library in Psammead
We should update Psammead's [test helpers](https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-test-helpers/src/index.js) so that they use [react-testing-library](https://testing-library.com/docs/react-testing-library/)

1. Update psammead's test helpers to include react-testing-library

Related: [[POC] [Do not merge] testing library alternatives for Simorgh ](https://github.com/bbc/simorgh/pull/1154)

**Testing notes**
This will not require Cypress tests. From my understanding, this will require a run of the tests on Psammead just to confirm that the change hasn't broken the existing logic.

**Additional context**

jroebu14:
Resolves https://github.com/bbc/psammead/issues/273

**Overall change:** Updates snapshot test helper to use `@testing-library/react`.

**Code changes:**

- Uses `@testing-library/react` instead of  `react-test-renderer` for generating easier to read snapshots.

New snapshots will be generated in a PR for [this issue](https://github.com/bbc/psammead/issues/276)

---

- [ ] Test engineer approval


packages/utilities/psammead-test-helpers/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 1.0.1   | [PR#679](https://github.com/bbc/psammead/pull/679) Use @testing-library/react for generating snapshots |
```
ghost: Update PR number

ghost (dismissed):

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -1,11 +1,12 @@
 import renderer from 'react-test-renderer';
```
thekp, line 1: is it possible to use `render` instead of `renderer` here as well (the same as you done below) ? then we don't have to import `react-test-renderer`: https://github.com/bbc/psammead/blob/13307abf6cda16bf55203aab604e04c731ba5853/packages/utilities/psammead-test-helpers/src/index.js#L30

thekp: also would it be good to add `@testing-library/react` as a dependency of this package ?

https://github.com/bbc/psammead/blob/13307abf6cda16bf55203aab604e04c731ba5853/packages/utilities/psammead-test-helpers/package.json#L27-L29

jroebu14: @thekp `react-test-renderer` is used in a shallow render function and `@testing-library/react` doesn't do shallow rendering for [these reasons](https://kentcdodds.com/blog/why-i-never-use-shallow-rendering). The best thing to do would be to refactor tests using shallow rendering to remove the use of shallow rendering.

Oh I didn't realise it had to be a dependency of this package too. Should it not be a dev dependency though?

jroebu14: I just had a look to see which tests are using shallow rendering test helpers and there appears to be none. It's worth considering removing the shallow render test helpers.

thekp: oh yeah, making it a dev-dependency makes more sense 👍

thekp (approved):
looks good to me, maybe just link an issue regarding our discussion.

jroebu14:
Linking to issue regarding removal of `react-test-renderer` in favour of `@testing-library/react` and refactoring of shallow render tests.

https://github.com/bbc/psammead/issues/774

ryanmccombe (approved):

jroebu14:
@PriyaKR this doesn't require testing.

dr3 (approved):

PriyaKR:
This PR doesn't require a testing from a tester.

PR #769, opened 2019-07-03
Fix `install:packages:link`

### LINKED ISSUE #756: Many storybook stories do not work after `install:packages:link`
**Describe the bug**
When a developer runs the `install:packages:link` script, around half of the stories in storybook do not load

**To Reproduce**

1. Pull project
1. `npm run install:packages:link`
1. `npm run storybook` and navigate through the stories

**Expected behavior**
Stories load and show the components

**Screenshots**
Screen Shot 2019-07-02 at 16 34 30

**Additional context**
It was noted that the stories that error in this way all appear to be using `inputProvider` function - this is likely the cause of the issue

There is also a version mismatch in most of the packages, although Lerna's `--hoist` flag reportedly handles this scenario correctly, so `inputProvider` is likely the problem

ryanmccombe:
Resolves https://github.com/bbc/psammead/issues/756

**Overall change:** Update the usage of the `inputProvider` helper to match the API

**Code changes:**
`inputProvider` callback is called with an object - most of the consumers of this function (but not the unit tests) were assuming the callback was invoked with multiple arguments

It's surprising this ever worked at all

---

- [ ] Test engineer approval

### DIFF
--- a/packages/components/psammead-brand/package.json
+++ b/packages/components/psammead-brand/package.json
@@ -23,7 +23,7 @@
   },
   "devDependencies": {
     "@bbc/psammead-assets": "^0.1.9",
-    "@bbc/psammead-storybook-helpers": "^2.1.1",
+    "@bbc/psammead-storybook-helpers": "^3.0.0",
     "@bbc/psammead-test-helpers": "^0.3.3",
     "react": "^16.8.6",
     "styled-components": "^4.3.2"
--- a/packages/components/psammead-brand/package.json
+++ b/packages/components/psammead-brand/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-brand",
-  "version": "4.1.4",
+  "version": "4.1.5",
   "main": "dist/index.js",
   "description": "Provides the BBC News logo (as SVG), nested a hardcoded link to https://www.bbc.co.uk/news",
   "repository": {
--- a/packages/components/psammead-caption/src/index.stories.jsx
+++ b/packages/components/psammead-caption/src/index.stories.jsx
@@ -26,7 +26,7 @@ storiesOf('Components|Caption', module)
         { name: 'Visual hidden text', defaultText: 'visually hidden text' },
         { name: 'Caption', defaultText: 'caption' },
       ],
-      ([hiddenText, captionText], script) => (
+      ({ slotTexts: [hiddenText, captionText], script }) => (
         <Caption script={script}>
           <VisuallyHiddenText>{hiddenText}</VisuallyHiddenText>
           {captionText}
@@ -42,7 +42,7 @@ storiesOf('Components|Caption', module)
         { name: 'Inline link', defaultText: 'inline link' },
         { name: 'Caption', defaultText: 'caption' },
       ],
-      ([linkText, captionText], script) => (
+      ({ slotTexts: [linkText, captionText], script }) => (
         <Caption script={script}>
           {`${captionText} `}
           <InlineLink href="https://www.bbc.com">{linkText}</InlineLink>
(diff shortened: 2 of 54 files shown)


packages/components/psammead-caption/src/index.stories.jsx
```diff
@@ -11,9 +11,12 @@ storiesOf('Components|Caption', module)
   .addDecorator(withKnobs)
   .add(
     'default',
-    inputProvider([{ name: 'Caption' }], ([captionText], script) => (
-      <Caption script={script}>{captionText}</Caption>
-    )),
+    inputProvider(
```
dr3, line 14: This is the old API for this package, this is on `@bbc/psammead-storybook-helpers` `"^2.1.1"`, if you link them locally it uses v3, which is a breaking change.

We will need to bump them all to `3.0.0` to fix this API issue :P

dr3 (dismissed):

packages/components/psammead-caption/src/index.stories.jsx
dr3: We use `--force-local` which overrides semver breaking change

ryanmccombe: @dr3 ah, if v2 used that API, that would explain why they worked when not linked

So just to be clear - my code changes here are correct, I also just need to update the package.json of all the changed components to use the latest version of `storybook-helpers`?

dr3: Yep :) We should raise an issue to review our use of `--force-local` too, to avoid issues with breaking changes in future. I think we had issues in the past where it wouldnt link packages with different minor versions, but that might be fine now

packages/components/psammead-brand/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "devDependencies": {
```
pjlee11, line 24: The version of `psammead-brand` needs to be bumped and changelog updated.

packages/components/psammead-media-indicator/package.json
```diff
@@ -22,7 +22,7 @@
     "@bbc/psammead-visually-hidden-text": "^0.1.12"
   },
```
pjlee11, line 23: media indicator needs a version bump and changelog

packages/components/psammead-section-label/package.json
```diff
@@ -26,7 +26,7 @@
     "@bbc/psammead-styles": "^0.3.4"
   },
```
pjlee11, line 27: section-label needs a version update and changelog

packages/components/psammead-sitewide-links/package.json
```diff
@@ -27,7 +27,7 @@
     "@bbc/psammead-styles": "^0.3.2"
   },
```
pjlee11, line 28: sitewide-links needs version and changelog bump

packages/components/psammead-story-promo/package.json
```diff
@@ -23,7 +23,7 @@
   "devDependencies": {
     "@bbc/psammead-image": "^1.0.0",
```
pjlee11, line 24: story-promo needs a version and changelog bump

packages/components/psammead-timestamp/package.json
```diff
@@ -20,7 +20,7 @@
     "@bbc/gel-foundations": "^3.0.0"
   },
```
pjlee11, line 21: timestamp needs version and changelog bump

pjlee11 (changes requested):
Just some admin of version bumping and changelogs, otherwise looking good

pjlee11 (approved):
Worth noting I pushed some minor changes but LGTM 👍

jroebu14 (approved):
LGTM 👍

ryanmccombe:
No QA needed here, as it only impacts Psammead development

PR #788, opened 2019-07-05
Use rem instead of px in the Media Indicator SVGs

### LINKED ISSUE #784: Brand and Media Indicator not scaling on text resize.
**Describe the bug**
If the text is resized to 200%, the `Brand` and `Media indicator` SVGs are very small.

**To Reproduce**

1. Go to `Firefox preferences -> Fonts and Colours`
2. Change font size to `72px`
3. Look at the Brand and any story with Media Indicator

**Expected behaviour**
They should scale proportionally with the font-size.

**Screenshots**
text-resized

60676190-2ecc9600-9e76-11e9-8534-cfef51c7b5c6

DenisHdz:
Resolves #784

**Overall change:** 
 In order to make the Media Indicator scale as the text is resized, they should use `rem` instead of `px`.

**Code changes:**
- Use `VideoMediaIcon` and `AudioMediaIcon` styled components with the width and height in `rem`s.
- Update snapshots

Screenshot 2019-07-05 at 13 34 48
Screenshot 2019-07-05 at 13 35 40

---

- [ ] Test engineer approval

### DIFF
--- a/packages/components/psammead-media-indicator/src/mediaIcons.jsx
+++ b/packages/components/psammead-media-indicator/src/mediaIcons.jsx
@@ -9,17 +9,27 @@ const MediaIcon = styled.svg`
   fill: ${C_EBON};
 `;
 
+const VideoMediaIcon = styled(MediaIcon)`
+  width: 0.75rem;
+  height: 0.75rem;
+`;
+
+const AudioMediaIcon = styled(MediaIcon)`
+  width: 0.8125rem;
+  height: 0.75rem;
+`;
+
 const icons = {
   video: (
-    <MediaIcon viewBox="0 0 32 32" width="12px" height="12px" focusable="false">
+    <VideoMediaIcon viewBox="0 0 32 32" focusable="false">
       <polygon points="3,32 29,16 3,0" />
-    </MediaIcon>
+    </VideoMediaIcon>
   ),
   audio: (
-    <MediaIcon viewBox="0 0 13 12" width="13px" height="12px" focusable="false">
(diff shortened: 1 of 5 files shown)


packages/components/psammead-media-indicator/src/mediaIcons.jsx
```diff
@@ -9,17 +9,27 @@ const MediaIcon = styled.svg`
   fill: ${C_EBON};
 `;
 
+const VideoMediaIcon = styled(MediaIcon)`
+  width: 0.75rem;
+  height: 0.75rem;
+`;
+
+const AudioMediaIcon = styled(MediaIcon)`
```
dr3, line 17: I have a feeling the reason we did html attributes for width and height before, not CSS, was to ensure the icons dont go huge when css is disabled. This way, _i think_ they go full screen with you disable css. Havent checked though

DenisHdz: True, so then we will have to accept that they don't scale when the text is resized or find another workaround because as far as I know, we can't put `rem` directly in the HTML attributes.

dr3: Could we have both? width attribute for no css, width css for with css?

DenisHdz: That worked 💅🏼

dr3 (approved):

j-pendlebury (approved):

david-boydell:
Great, this is ready for merge

PR #1357, opened 2019-07-17
Use the npm client instead of exec in the version check script

### LINKED ISSUE #1081: Use the npm client instead of exec in the version check script
In the `getRegistryVersion.js` script we have the following line:
https://github.com/bbc/psammead/blob/ef8ea85e708ce26c8acf7f086ad0b8c294a390fc/scripts/publish/src/getRegistryVersion.js#L19
Now this line alone introduces a potential command injection vulnerability which is mitigated by enforcing a certain regex in order to check for package names.
Instead of using exec and running the npm command we could use the web client provided by npm themselves. 

- Replace the exec command with a call to the npm registry: https://www.npmjs.com/package/npm-registry-client
- Use the resulting version array to determine the package version (you will be able to most likely drop and replace the current returned object from the exec function) 
- The package regex should still stay

Democracy is non-negotiable

**Testing notes**
Doesn't require any testing from a tester.

**Additional context**

DemonusPC:
Resolves #1081

**Overall change:** 
The initial issue that this Pr resolves, only focused on the implamentation of the npm web client into the script. However, the npm client does not have a function that works in sync. This mean that I had to refactor the script to support asynchronous execution. The good news is that it made the script really quick. 

**Code changes:**
- Refactored the script using async/await
- Refactored the way we report back the script status
- Installed the npm api client
- Implemented the script to use the npm client

---

- [ ] Test engineer approval


dr3 (changes requested):
Personally I dont think security matters here, hence why I didnt do this stuff when i wrote this code. We dont run forks, and this code only runs on merge, after code reviews. If someone was going to abuse this script, they could just change the entire script to run what they want, it would equally have to get past 2 devs (one being a code owner) to run.

If someone malicious was to get a branch to run on jenkins, they would be much better off using `child_process` `exec` in a test file somewhere and running what ever they wanted. That wouldn't even need to be merged to run.

Anyway 🙈 If we do care about security for this script, Adding this to any package.json will create a file called `hey.txt`, or do anything it wants, so we should probably limit it too
```
"publishConfig": {
 "access": "restricted; echo 'hello' > hey.txt; "
},
```

jamesbhobbs:
> Personally I dont think security matters here, hence why I didnt do this stuff when i wrote this code. We dont run forks, and this code only runs on merge, after code reviews. If someone was going to abuse this script, they could just change the entire script to run what they want, it would equally have to get past 2 devs (one being a code owner) to run.
> 
> If someone malicious was to get a branch to run on jenkins, they would be much better off using `child_process` `exec` in a test file somewhere and running what ever they wanted. That wouldn't even need to be merged to run.
> 
> Anyway 🙈 If we do care about security for this script, Adding this to any package.json will create a file called `hey.txt`, or do anything it wants, so we should probably limit it too
> 
> ```
> "publishConfig": {
>  "access": "restricted; echo 'hello' > hey.txt; "
> },
> ```

My only response is to clarify that we will run forks in the future.

dr3:
>My only response is to clarify that we will run forks in the future.

@benjaminhobbs How will we stop someone from editing our publish script to do what they want? You can lock down a jenkins pipeline so if you change `withCredentials` the jenkins file wont run, to avoid easily accessing the credentials, but you cant stop it running if someone changes the publish scripts AFAIK.

jamesbhobbs:
> > My only response is to clarify that we will run forks in the future.
> 
> @benjaminhobbs How will we stop someone from editing our publish script to do what they want? You can lock down a jenkins pipeline so if you change `withCredentials` the jenkins file wont run, to avoid easily accessing the credentials, but you cant stop it running if someone changes the publish scripts AFAIK.

I'm not going to answer security questions on a public repo, will reply IRL.

DemonusPC:
> I dont think security matters here

Security should matter no matter what you are writing. The code should be secure by default, even if it is a dumb script that checks for packages. This is because someone smarter than all of us could come along still find a way to exploit it. 
On top of this, this refactoring improves the speed of this script significantly. The update is near instant.

dr3 (changes requested):
>Security should matter no matter what you are writing. The code should be secure by default, even if it is a dumb script that checks for packages.

Adding this to any package.json will create a file called `hey.txt`, or do anything it wants, so we should limit it too
```
"publishConfig": {
 "access": "restricted; echo 'hello' > hey.txt; "
},
```

DemonusPC:
Ok I don't care about this anymore. Ignore the security and performance improvements.

PR #1399, opened 2019-07-19
Module to update packages that depend on packages

### LINKED ISSUE #1354: Bumper bot - Find what packages use a dependancy and bump it
Summarised in #888 we need to create a bumping bot. We need to be able to bump dependancies in psammead

Create a script that, given a package name like `@bbc/psammead-assets`, finds all psammead packages that use it, and update their version of `@bbc/psammead-assets` to latest. Ideally this will use lerna and existing npm packages where possible.

it should be availible as a node script which can be called by "bumpDependancy(`@bbc/psammead-assets`)

**Testing notes**
[Tester to complete]

Dev insight: unit tests

**Additional context**

jroebu14:
Resolves https://github.com/bbc/psammead/issues/1354

**Overall change:**
Adds a module to update all packages that depend on specified packages to the latest version. 

After updating is completed the module then returns an array of the changed packages that can be passed along the CI pipeline for then bumping versions and updating changelogs.

**Code changes:**
- uses `lerna exec` and `npm-check-updates` for handling package updates.
- adds a function for updating all packages that depend on passed in package(s) to the latest version using `npm-check-updates`.
- adds a function for parsing the upgraded dependencies output from `npm-check-updates` to return an array.

**Dev notes:**
Take note that `updatePackagesDependents` will return packages with the scope `@bbc/` like `@bbc/psammead-image` and not `psammead-image`. This may cause confusion because `bumpPackages` and `bumpChangelogs` expect packages without the scope. Should consider refactoring `bumpPackages` and `bumpChangelogs` to accept packages with the scope.

Running `npx lerna exec --parallel --no-bail -- npx npm-check-updates` is quite slow. I think this is because `npm-check-updates` is installed in each package on the fly. It might be better for performance if `npm-check-updates` is a dev dependency of every package then running `npx lerna exec --parallel --no-bail -- npm run npm-check-updates  ${packageList} -u -a --jsonUpgraded` (note `npm run npm-check-updates` and not ` npx npm-check-updates`) wouldn't have to install `npm-check-updates` multiple times in parallel.

---

**Testing notes**
No testing required because it is a module for upgrading dependencies and isn't being used in the CI pipeline yet.

- [ ] Test engineer approval


scripts/utilities/getPackagesDependents/getAllPackagesDependencies.js
```diff
@@ -0,0 +1,26 @@
+const { exec } = require('shelljs');
+
+module.exports = () =>
+  new Promise((resolve, reject) => {
+    const callback = (code, output) => {
+      const getPackageDependencies = ({ name, dependencies }) => ({
+        name,
+        dependencies: Object.keys(dependencies),
```
dr3: I think we also need dev deps

jroebu14: this returns both deps and dev deps

ryanmccombe (dismissed):

dr3 (approved):
LGTM, tried locally with 
```
const get = require('./scripts/upgradeDependencies/index');
const stuff = get(['@bbc/gel-foundations', '@bbc/psammead-assets']);
stuff.then(thing => console.log(thing))
```

ryanmccombe (approved):

dr3:
Doesnt need testing, not prod

PR #1410, opened 2019-07-23
WIP - POC: Add Header and Footer SVG fallback to png support

### LINKED ISSUE #383: Header & Footer - fallback to PNG if browser does not support SVG
From what we've seen here, there are browsers that do not support SVGs https://caniuse.com/#search=svg. However these are very low in the stats

Therefore we would like to fall back to text link (BBC News) working with UX to ensure that the experience is OK

Bopchy:
Resolves #383 

**Overall change:** _Adds PNG fallback support for Header and Footer._

**Code changes:**

- _Check for SVG browser support._
- _Added static image support in Storybook config for testing._
- _Created getImage() to enable passing image paths as props._
- _Added tests._ 
- _Created stories._

**How do I see the fallback images on Storybook?**
- Uncheck `Use SVG instead of PNG fallback image` on brand story
---

- [ ] Test engineer approval


greenc05:
@Bopchy what mark up will a png output? You don't need a focusable="false" attribute on it... Will there be an image tag?

Bopchy:
Yes. There will be an image tag 

The mark up will be `<div>  </div>`

greenc05:
With an empty alt attribute? And then we don't need aria-hidden="true" either? I assume that we are still having the visually hidden text element that comes with the brand?

Bopchy:
Yes, the visually hidden text element will still be there. I will be adding an alt, I am not sure if the aria-hidden="true" should be removed though

greenc05:
aria-hidden="true" shouldn't been needed if there is an empty alt attribute on the image tag - this makes a screen reader ignore the image as it if wasn't there

greenc05:
I realise this is a POC but of course a developer accessibility swarm should be carried out on this when it's done for real

Bopchy:
This is no longer needed.

PR #1511, opened 2019-07-26
Setup chromaticqa

chris-hinds:
Resolves #NUMBER

**Overall change:**

Install chromaticqa as a dev dependancy, add a new npm script to run the chromaticqa tests and runs the tests on the CI pipeline

---

- [N/A] Tests added for new features
- [ ] Test engineer approval


package.json
```diff
@@ -7,12 +7,13 @@
     "build": "npm run build:cjs && npm run build:esm",
     "build:cjs": "lerna exec --parallel -- babel src -d dist --ignore $npm_package_config_babel_ignore --root-mode upward",
     "build:esm": "BABEL_TYPE=esm lerna exec --parallel -- babel src -d esm --configFile babel.config.js --ignore $npm_package_config_babel_ignore --root-mode upward",
+    "build:storybook": "build-storybook -o storybook_dist",
```
thekp, line 10: noticed we are `build:storybook`-ing differently to the way we do it on [Simorgh](https://github.com/bbc/simorgh/blob/latest/package.json#L12)

we include a `-c` option which specifies configuration to be used. [Storybook options].(https://storybook.js.org/docs/configurations/cli-options/)

not too sure what this config is for and what impact it will have if we specify it 👀 

probably one for a separate issue ^

dr3 (approved):

package.json
chris-hinds: Good spot, I'b been inclined to not add this for now, as it wasn't present before. But one we should keep an eye on.

thekp (approved):

PriyaKR:
LGTM.

package.json
```diff
@@ -22,6 +23,7 @@
     "publish": "npm run build && node scripts/publish",
     "storybook": "NODE_ENV=development start-storybook -p 8180 -c .storybook",
     "test": "npm run test:lint && npm run test:lint:css && npm run test:unit && npm run test:packages",
+    "test:chromatic": "chromatic test run --build-script-name build:storybook || true",
```
jamesdonoh, line 26: The `|| true` is to prevent it from ever failing the build, right? Is the plan to remove this eventually?

jamesdonoh (approved):

package.json
chris-hinds: Correct and I do not believe so,  `|| true` will stop chromaticqa failing the pipeline however the chromaticqa status check will still fail. and eventually that status check will be marked as required to prevent merging if these new tests do fail.

PR #1538, opened 2019-07-31
Revert "Moment timezone include"

dr3:
Reverts bbc/psammead#1521

Revert due to broken deployment


12 (approved):
Good work

pjlee11 (approved):
Yes no maybe I don’t know can you repeat the question

PR #1594, opened 2019-08-05
Consent banner - RTL

sareh:
Part of https://github.com/bbc/simorgh/issues/1903

**Overall change:** Updates Consent Banner component to take in a `dir` property, so it flips the layout for RTL services vs LTR services. 

In storybook, you just select the service from the drop-down menu and the layout should flip accordingly. 

**Code changes:**

- Added prop `dir`, which defaults to 'ltr', but also allows 'rtl'.
- Interpolated `margin-{x}` properties & `float` values using dir value. 
- Updated dependencies.
News service - has dir `ltr`
Screen Shot 2019-08-05 at 13 14 50

## Screenshots
**Note - now by pulling in `latest` the input provider is broken, so we do not see the Arabic/Persian text in the screenshots below**
Screen Shot 2019-08-05 at 18 13 22

Screen Shot 2019-08-05 at 18 12 34

## **Screenshots with the layout changes plus localised text**
Arabic service - has dir `rtl`
Screen Shot 2019-08-05 at 13 15 01

Persian service - has dir `rtl`
Screen Shot 2019-08-05 at 13 15 20

---

- [ ] Test engineer approval


packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -22,6 +22,11 @@ import {
 } from '@bbc/gel-foundations/spacings';
 import { getSansRegular } from '@bbc/psammead-styles/font-styles';
 
+const ltrRtl = (ltrValue, rtlValue) =>
```
sareh, line 25: Currently, this `ltrRtl` method is only required within this component. When it is needed in more than one component, we can pull it out into a shared utility package within Psammead.

packages/components/psammead-consent-banner/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 2.2.0 | [PR#1594](https://github.com/bbc/psammead/pull/1594) Update layout to flip correctly for right-to-left services |
```
12, line 6: If we're adding `dir` as a required prop, this should be a breaking change, right?

packages/components/psammead-consent-banner/README.md
```diff
@@ -21,6 +21,7 @@ The `psammead-consent-banner` component is a styled `div` that encapsulates info
 | reject | element | Yes | N/A | `<a href="https://www.bbc.co.uk/usingthebbc/your-data-matters">Find out what's changed</a>` |
 | id | string | No | Null | `ConsentBanner` |
 | hidden | bool | No | Null | `false` |
+| dir | string | Yes | N/A | One of `'rtl'` `'ltr'` |
```
12: Are we sure we don't want to default to `ltr`?

12 (changes requested):
This is really good, and I like that it's been made in a way such that we can spin it into its own utility in the future.

I think if we are to make this a required prop, we need to make this a breaking change.

However, I personally think we should default to `ltr`, what do you think?

packages/components/psammead-consent-banner/README.md
sareh: Updated to default to ltr.

12 (approved):
Looks good, checked out locally and is achieving issue requirements.

I've noticed that we're passing `dir` and `script` around quite a bit, have we considered using Styled Components theme providers? https://www.styled-components.com/docs/advanced#theming

sareh:
@rhell Thanks. Yes, Theme Providers are one way of handling this. It'd mean updating all of our components to use it. We could then replace the use of separate `dir` & `script` props with these:
`${props => props.theme.arabic.dir}` 
`${props => props.theme.arabic.script}`
This would be a tidier way of handling props in these components. 
One reason why we didn't use themes at the start of creating the Psammead component library was that we were not sure how many axes of variation we'd need, or the level of abstraction. e.g. a theme for different Article variants (like a dark theme for immersive stories and a light theme for standard stories), or whether we'd have multiple brand palettes and need to use the themes for that purpose.  It is worth opening up this discussion again. I'll raise an issue to consider the pros and cons of introducing the use of Theme Providers with Psammead.

dr3 (approved):

PR #1642, opened 2019-08-06
Lint only changed files in hooks

mbzain76:
Install lint-staged.

Resolves bbc/simorgh-infrastructure#636

Overall change: _A very high-level summary of easily-reproducible changes that can be
Install lint-staged to only lint changed files in Psammead.

Code changes:
Add new configuration to package.json and package-lock.json

---

- [ ] Test engineer approval


package.json
```diff
@@ -140,5 +141,15 @@
       "cp -af /usr/lib/PSAMMEAD-COMPONENT-LIBRARY/bake-scripts/. /etc/bake-scripts/PSAMMEAD-COMPONENT-LIBRARY",
       "chmod +x /etc/bake-scripts/PSAMMEAD-COMPONENT-LIBRARY/*"
     ]
+  },
+  "lint-staged": {
+    "*.js": [
```
chris-hinds: This will need to run against the same file extensions that we current lint i.e. `*.{js,jsx,json}` at the moment this only runs against `.js` files.

We also do not current use the auto-fix option so this should be removed.

Something like 
```
"lint-staged": {
    "*.{js,jsx,json}": "eslint"
  }
```

package.json
```diff
@@ -140,5 +141,15 @@
...
+      "eslint --fix",
+      "git add"
+    ],
+    "*.css": [
```
chris-hinds: Psammead uses css-in-js so we have no `.css` file to lint. Have a look at the `test:lint:css` script we already have and make sure stylelint is run against those same files.

package.json
```diff
@@ -140,5 +141,15 @@
       "cp -af /usr/lib/PSAMMEAD-COMPONENT-LIBRARY/bake-scripts/. /etc/bake-scripts/PSAMMEAD-COMPONENT-LIBRARY",
       "chmod +x /etc/bake-scripts/PSAMMEAD-COMPONENT-LIBRARY/*"
     ]
+  },
+  "lint-staged": {
+    "*.js": [
+      "eslint --fix",
+      "git add"
```
chris-hinds: Also not sure that we want to auto run `git add`

chris-hinds (changes requested):

package.json
mbzain76: Done

package.json
mbzain76: Done

package.json
mbzain76: Done

chris-hinds (approved):

dr3 (approved):

PR #1713, opened 2019-08-09
Adding new scripted pipeline running on single node

### LINKED ISSUE #1564: Lock to single node on Jenkins to improve pipeline stability
There has now been a number of occasions where builds are randomly failing, generally with the error `command storybook-build no available`.

I believe this is happening when a stage in the pipeline runs on a different agent to the first stage, meaning the second stage does not have access to the node_modules.

Locking to a single node would solve this however would potentially limit us in the future if we wanted to run some stages in parallel.

Restructure the Jenkins file to ensure that all stage run on a single node/Jenkins agent.

Creating and parsing archives of node modules to different stages.

**Testing notes**
[Tester to complete]

**Additional context**

chris-hinds:
Resolves #1564 

**Overall change:**

This PR refactors the pipeline to run as a scripted pipeline and on a single node.

The aim is to simply the pipeline as well as reducing the risk of running on multiple agents where the node_modules dir is then not available.

---

- [ ] Test engineer approval


Jenkinsfile
```diff
@@ -1,150 +1,100 @@
...
-def nodeImageVersion = "10.16.0"
-def nodeImage = "${dockerRegistry}/bbc-news/node-10-lts:${nodeImageVersion}"
+def nodeImage = "329802642264.dkr.ecr.eu-west-1.amazonaws.com/bbc-news/node-10-lts:10.16.0"
 
-def nodeName
-def slackChannel = "#si_repo-psammead"
-def stageName = "Unknown"
-def gitCommitAuthor = "Unknown"
-def gitCommitMessage = "Unknown"
+// def slackChannel = "#si_repo-psammead"
```
jamesdonoh: Presumably this will be deleted/changed after the new pipeline has been validated?

package.json
```diff
@@ -13,7 +13,7 @@
     "bumpPackages": "node scripts/bumpPackages/cli",
     "bumpChangelogs": "node scripts/bumpChangelogs/cli",
     "regeneratePackageLocks": "node scripts/regeneratePackageLocks",
-    "deploy-storybook": "npm run build:storybook && gh-pages -d storybook_dist",
+    "deploy-storybook": "gh-pages -d storybook_dist",
```
jamesdonoh, line 16: Just to clarify, why is the call to `build:storybook` no longer needed here? Because it's called via the Jenkinsfile?

Jenkinsfile
```diff
@@ -1,150 +1,100 @@
...
+
+      // get git commit info for notifications
+      gitCommitHash = sh(returnStdout: true, script: "git log -n 1 --pretty=format:'%h'").trim()
+      gitCommitAuthor = sh(returnStdout: true, script: "git --no-pager show -s --format='%an' ${gitCommitHash}").trim()
+      gitCommitMessage = sh(returnStdout: true, script: "git log -1 --pretty=%B").trim()
+
+      docker.image("${nodeImage}").inside {
+        try {
+          stage ('Setup & Install') {
+
```
jamesdonoh: Spare line?

Jenkinsfile
```diff
@@ -1,150 +1,100 @@
...
+            sh 'make setup-git'
+            sh 'npm run ci:packages'
+            sh 'npm run build:storybook'
```
jamesdonoh: Wondering about the mix-and-match of `npm` scripts and `Makefile` targets that are used here and in the following stage(s), is there an ongoing reason to need both or is there an opportunity to get more consistency/simplicity through this refactor?

Jenkinsfile
```diff
@@ -1,150 +1,100 @@
...
+        } catch (Throwable e) {
+          echo "Pipeline Failed: ${e}"
+
+          // send slack notification if building branch: latest
+          if (env.BRANCH_NAME == 'latest') {
+            // catch error in stages and notify slack
+            notifySlack('danger', 'Failed', gitCommitAuthor, e, gitCommitHash, gitCommitMessage, slackChannel)
+          }
+
+          // throw caught error to ensure pipeliune fails
```
jamesdonoh:
```suggestion
          // throw caught error to ensure pipeline fails
```

Jenkinsfile
```diff
@@ -1,150 +1,100 @@
...
+          throw e
+        } finally {
+          // send slack notification if building branch: latest
+          if (env.BRANCH_NAME == 'latest') {
+            switch (currentBuild.result) {
+              case 'SUCCESS':
+                notifySlack('good', 'Success', gitCommitAuthor, 'Successfully Deployed', gitCommitHash, gitCommitMessage, slackChannel)
+                break
+              case 'UNSTABLE':
+                notifySlack('DANGER', 'Ustable', gitCommitAuthor, 'Pipeline in an unstable state', gitCommitHash, gitCommitMessage, slackChannel)
```
jamesdonoh:
```suggestion
                notifySlack('DANGER', 'Unstable', gitCommitAuthor, 'Pipeline in an unstable state', gitCommitHash, gitCommitMessage, slackChannel)
```

Jenkinsfile
```diff
@@ -1,150 +1,100 @@
...
-      }
-      steps {
-        unstash 'psammead-publishes'
-        sh 'make bump-dependants'
-      }
-      post {
-        always {
-          script {
-            stageName = env.STAGE_NAME
+        } catch (Throwable e) {
```
jamesdonoh, line 73: Not familiar with Groovy exception handling but assuming it works like Java won't this result in a double Slack notification being sent as the `finally` block will also run?

Also is the `Throwable` here redundant?

Jenkinsfile
```diff
@@ -1,150 +1,100 @@
...
+          if (env.BRANCH_NAME == 'latest') {
+            switch (currentBuild.result) {
+              case 'SUCCESS':
+                notifySlack('good', 'Success', gitCommitAuthor, 'Successfully Deployed', gitCommitHash, gitCommitMessage, slackChannel)
+                break
+              case 'UNSTABLE':
+                notifySlack('DANGER', 'Ustable', gitCommitAuthor, 'Pipeline in an unstable state', gitCommitHash, gitCommitMessage, slackChannel)
+                break
+              default:
+                echo "Pipeline has failed with an unknow build result"
```
jamesdonoh:
```suggestion
                echo "Pipeline has failed with an unknown build result"
```

Jenkinsfile
chris-hinds: have changed this now. Just forgot to set it back.

package.json
chris-hinds: Story book is now built once after the node modules are installed. previously we would run the chromaticqa tests (which builds storybook to a tmp dir) then we would run `deploy-storybook` which would build storybook again.

This avoids the double storybook build as we can get chromaticqa to use the already built storybook.

Jenkinsfile
chris-hinds: Whats your preference here? Im happy with using just the makefile.

Jenkinsfile
chris-hinds: Was having having issues where the catch was not actually catching pipeline errors until I used `Throwable e`

It is true the the `finally` block will always run, this is why the default in the switch statement does not send a slack notification it simply echos the build status.

The switch case in the finally block also ensure we only send a slack notification if the build status is `SUCCESS` or `UNSTABLE` failed states will be caught in the `catch`

Jenkinsfile
chris-hinds: Fixed

Jenkinsfile
jamesdonoh: OK - though I still don't quite get it, if the build is `UNSTABLE` won't you get both a `Failed` and an `Unstable` Slack notification?

Jenkinsfile
jamesdonoh: If they are intended to be run only on CI inside a known Docker image personally I'd go for Makefile targets as they are more flexible than `npm` scripts but I don't have a very strong preference, just think it's not currently clear why we are using a combination of both.

Jenkinsfile
chris-hinds: hmmmmm I am not actually 100% sure. I need to see what triggers the `unstable` state.

if that is the case we can check the build status in the catch. And only have the finally block post to slack if the build status is a `success`?

chris-hinds: @jamesdonoh https://ci.news.tools.bbc.co.uk/blue/organizations/jenkins/psammead/detail/pipeline-improvemnets/45/pipeline/28

image


I forced this pipeline into an `UNSTABLE` state. This was not caught by the catch.

The notification was sent from the switch statement in the finally `block`

Jenkinsfile
chris-hinds: 👍 have moved to using the makefile for all pipeline steps

Jenkinsfile
```diff
@@ -1,153 +1,103 @@
...
+def notifySlack(colour, buildStatus, gitCommitAuthor, info, gitCommit, gitCommitMessage, slackChannel) {
+  // call the global slackSend method in Jenkins
+  slackSend(color: colour, message: "*${buildStatus}* on ${env.BRANCH_NAME} [build ${BUILD_DISPLAY_NAME}] \n*Author:* ${gitCommitAuthor} \n*Info:* ${info} \n*Commit Hash* \n${gitCommit} \n*Commit Message* \n${gitCommitMessage}", channel: slackChannel)
 }
 
-def notifySlack(colour, buildStatus, gitCommitAuthor, stageName, gitCommitMessage, slackChannel) {
-  // call the global slackSend method in Jenkins
-  slackSend(color: colour, message: "*${buildStatus}* on ${GIT_BRANCH} [build ${BUILD_DISPLAY_NAME}] \n*Author:* ${gitCommitAuthor} \n*Stage:* ${stageName} \n*Commit Hash* \n${GIT_COMMIT} \n*Commit Message* \n${gitCommitMessage}", channel: slackChannel)
+def cleanUp() {
+  sh 'chmod -R 777 .git'
```
andrew-nowak, line 13: i don't understand this method - why change the permissions? and im not sure how it's going to clean up?

Jenkinsfile
```diff
@@ -1,153 +1,103 @@
...
+      'CC_TEST_REPORTER_ID=06c1254d7c2ff48f763492791337193c8345ca8740c34263d68adcc449aff732'
+    ]) {
+      cleanWs()
+
+      // git checkout
+      checkout scm
+
+      // get git commit info for notifications
+      gitCommitHash = sh(returnStdout: true, script: "git log -n 1 --pretty=format:'%h'").trim()
+      gitCommitAuthor = sh(returnStdout: true, script: "git --no-pager show -s --format='%an' ${gitCommitHash}").trim()
```
andrew-nowak: nit picking - why not have the shell commands for gitCommitAuthor and gitCommitMessage match? They're doing roughly the same operation
```suggestion
      gitCommitAuthor = sh(returnStdout: true, script: "git log -1 --pretty=%an").trim()
```

Jenkinsfile
andrew-nowak: Perhaps the success/unstable slack notifications could be moved into the main 'try' block, or after the finally block? That way the failure execution flow shouldn't go anywhere near them

Makefile
```diff
@@ -13,26 +9,29 @@ code-coverage-before-build:
...
 setup-git:
 	git remote set-url origin "https://${GITHUB_TOKEN}@github.com/bbc/psammead.git"
 	git config remote.origin.fetch '+refs/heads/*:refs/remotes/origin/*'
 	git config user.email "bbc-news-frameworks@users.noreply.github.com"
 	git config user.name "BBC News Frameworks"
 
-storybook:
-	make setup-git;
-	npm run deploy-storybook;
+publush:
```
andrew-nowak:
```suggestion
publish:
```

Jenkinsfile
```diff
@@ -1,153 +1,103 @@
...
+          if (env.BRANCH_NAME == 'latest') {
+            switch (currentBuild.result) {
+              case 'SUCCESS':
+                notifySlack('good', 'Success', gitCommitAuthor, 'Successfully Deployed', gitCommitHash, gitCommitMessage, slackChannel)
+                break
+              case 'UNSTABLE':
+                notifySlack('DANGER', 'Unstable', gitCommitAuthor, 'Pipeline in an unstable state', gitCommitHash, gitCommitMessage, slackChannel)
+                break
+              default:
+                echo "Pipeline has failed with an unknown build result"
```
andrew-nowak: This feels like something that should be slacked? I can see that this shouldn't happen, but I'm somewhat concerned that if nothing gets slacked then no one will notice that the pipeline is in a really weird state.

Jenkinsfile
```diff
@@ -1,153 +1,103 @@
...
+
+      // git checkout
+      checkout scm
+
+      // get git commit info for notifications
+      gitCommitHash = sh(returnStdout: true, script: "git log -n 1 --pretty=format:'%h'").trim()
+      gitCommitAuthor = sh(returnStdout: true, script: "git --no-pager show -s --format='%an' ${gitCommitHash}").trim()
+      gitCommitMessage = sh(returnStdout: true, script: "git log -1 --pretty=%B").trim()
+
+      docker.image("${nodeImage}").inside {
```
andrew-nowak, line 32: Does this mean that we're losing the `args '-u root -v /etc/pki:/certs'` on the docker image? Is that likely to matter?

Jenkinsfile
andrew-nowak: Isn't this missing aborted? Or is that caught in the try/catch?

andrew-nowak (changes requested):
the only thing that "needs" "fixing" is the publish typo; everything else is questions :)

Jenkinsfile
chris-hinds: We are loosing those args yes, It doesn't matter. Those arguments are only needed when we need certs to interface with cosmos

Jenkinsfile
chris-hinds: The reason this isn't slacked is because an unstable build is caught in the try catch but this finally block will always run so would cause a double slack notification.

Jenkinsfile
chris-hinds: This is from the talos implementation. According to @dr3 777 permissions are required to jenkins can remove the .git folder ready for the next build. A new issues is being raised to see if we can move away from have to do this.

Jenkinsfile
chris-hinds: I dont think we can do this, as an unstable will be caught by the try catch no matter what

Jenkinsfile
andrew-nowak: sounds good!

Jenkinsfile
andrew-nowak: fair enough

Jenkinsfile
andrew-nowak: if unstable builds enter the catch block, then why don't we switch on `currentBuild.result` and slack failures/unstables from there?

Jenkinsfile
andrew-nowak: But `UNSTABLE` is in the above case statement?
(I think you mean `FAILURE` or whatever the code is in Jenkins, in which case:)
could we not 
```
case 'FAILURE':
  break
default:
  notifySlack('DANGER', 'Unknown', ....)
  break
```

chris-hinds: Discussed in a DM and changes pushed

Jenkinsfile
```diff
@@ -1,154 +1,103 @@
...
+                notifySlack('danger', 'Failure', gitCommitAuthor, 'Pipeline has failed', gitCommitHash, gitCommitMessage, slackChannel)
+                break
+              case 'UNSTABLE':
+                notifySlack('danger', 'Unstable', gitCommitAuthor, 'Pipeline in an unstable state', gitCommitHash, gitCommitMessage, slackChannel)
+                break
+              case 'ABORTED':
+                notifySlack('danger', 'Aborted', gitCommitAuthor, 'Pipeline was aborted', gitCommitHash, gitCommitMessage, slackChannel)
+                break
+              default:
+                notifySlack('danger', 'Unknow', gitCommitAuthor, 'Pipeline has failed', gitCommitHash, gitCommitMessage, slackChannel)
```
andrew-nowak:
```suggestion
                notifySlack('danger', 'Unknown', gitCommitAuthor, 'Pipeline has failed', gitCommitHash, gitCommitMessage, slackChannel)
```

andrew-nowak (approved):
logic lgtm, I don't know much about Groovy/Jenkinsfile syntax but no obvious problems

jamesdonoh (approved):
Nice, this seems like a positive step. 👍

PR #1714, opened 2019-08-09
Storybook helpers v5

### LINKED ISSUE #1668: Storybook Input Provider - Incorrect Service Names
**Describe the bug**
The input provider currently has the following issues:

- Incorrect service names - the `indonesia` service is called `indonesian` and `afaanoromoo` is called `afaanOromoo` (wrong case)
- Cannot configure default service - hardcoded to `news` - assumes every component supports this service

ryanmccombe:
Resolves https://github.com/bbc/psammead/issues/1668

**Overall change:** _Update service names for parity with Simorgh, and add option to set default service._

**Code changes:**

- _Renamed `indonesian` to `indonesia`._
- _Renamed `afaanOromoo` to `afaanoromoo`._
- _`inputProvider` now allows a default service to be configured, rather than always using `news`_

---

- [ ] Test engineer approval

### DIFF
--- a/packages/utilities/psammead-storybook-helpers/CHANGELOG.md
+++ b/packages/utilities/psammead-storybook-helpers/CHANGELOG.md
@@ -3,7 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
-| 5.0.0 | [PR#???](https://github.com/bbc/psammead/pull/???) Renamed `indonesian` to `indonesia` and `afaanOromoo` to `afaanoromoo`.  Input provider now has an optional 4th argument that allows the default service to be configured |
+| 5.0.0 | [PR#1714](https://github.com/bbc/psammead/pull/1714) Renamed `indonesian` to `indonesia` and `afaanOromoo` to `afaanoromoo`.  Input provider now has an optional 4th argument that allows the default service to be configured |
 | 4.0.0 | [PR#1679](https://github.com/bbc/psammead/pull/1679) Remove text knob from input-provider |
 | 3.3.4   | [PR#1685](https://github.com/bbc/psammead/pull/1685) Bump dependencies |
 | 3.3.3 | [PR#1682](https://github.com/bbc/psammead/pull/1682) Move all dev dependencies to top level package.json |
--- a/packages/utilities/psammead-storybook-helpers/CHANGELOG.md
+++ b/packages/utilities/psammead-storybook-helpers/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 5.0.0 | [PR#???](https://github.com/bbc/psammead/pull/???) Renamed `indonesian` to `indonesia` and `afaanOromoo` to `afaanoromoo`.  Input provider now has an optional 4th argument that allows the default service to be configured |
 | 4.0.0 | [PR#1679](https://github.com/bbc/psammead/pull/1679) Remove text knob from input-provider |
 | 3.3.4   | [PR#1685](https://github.com/bbc/psammead/pull/1685) Bump dependencies |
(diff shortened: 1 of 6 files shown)


packages/utilities/psammead-storybook-helpers/src/input-provider.jsx
```diff
@@ -4,14 +4,23 @@ import { Helmet } from 'react-helmet';
 import * as scripts from '@bbc/gel-foundations/scripts';
 import LANGUAGE_VARIANTS from './text-variants';
 
-const inputProvider = (slots, componentFunction, services) => () => {
+const inputProvider = (
+  slots,
+  componentFunction,
+  services,
+  options = {},
```
jroebu14, line 11: are we expecting more options? what if this arg was just `defaultService = 'news'`

dr3: Yeah, ide almost prefer 
```
(slots, componentFunction, services)
```
-> 
```
({ slots, componentFunction, services, defaultService })
```
or put `services` inside options, as its optional

ryanmccombe: I don't know if we'll need to add more arguments to this eventually, but I think the amount of arguments here is already problematic - you can see that in the test file where there are `null` parameters being passed in everywhere

Made the 4th an options object to stop the problem getting any further out of hand - https://github.com/ryanmcdermott/clean-code-javascript#function-arguments-2-or-fewer-ideally

ryanmccombe: @dr3 agree that all the arguments should just be one object eventually, but that's a massive breakign change that would require updating all stories in Simorgh and Psammead to a new API.  Not sure that's in scope for media pod?  We just want our page on storybook 😢

ryanmccombe: https://github.com/ryanmcdermott/clean-code-javascript#function-arguments-2-or-fewer-ideally

chris-hinds: @ryanmccombe definitely not in scope for the media pod or this PR really. Could you raise an issue about that refactor though so it can be picked up and looked into by the core pod at a later date.

ryanmccombe: https://github.com/bbc/simorgh/issues/3077

dr3 (approved):

packages/utilities/psammead-storybook-helpers/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 5.0.0 | [PR#1714](https://github.com/bbc/psammead/pull/1714) Renamed `indonesian` to `indonesia` and `afaanOromoo` to `afaanoromoo`.  Input provider now has an optional 4th argument that allows the default service to be configured |
```
pjlee11, line 6: Any reason you went with `5.0.0` rather than `4.1.0`, IMO this doesn't need to be a major change

ryanmccombe: I believe the service rename part of this is a breaking change - previously, "indonesian" appeared on the dropdown, and got passed to your component if someone selected it.  Now, it will be "indonesia"

At least I think that's true - the only thing that makes me doubt it is surely some components would have errored if you passed an invalid service name in the current version?

But maybe nobody noticed it as these services weren't used

pjlee11: Ahhh yeah, it is likely those services haven't been used/bugs haven't been spotted.

pjlee11 (approved):

ryanmccombe:
No testing needed here - just some behind-the-scenes work to add functionality that will be used for future PRs

PR #1758, opened 2019-08-14
Thai typography update

sareh:
No ticket - work raised as a result of the UX review. 

**Overall change:** Updated Minion and Brevier type sizes for Thai

**Code changes:**

- Change Thai Minion to be font-size 12, line height 16, for the media indicators since the text is with Western Arabic numerals (1, 2,..) they should match the Latin line height and font size.
- Change Thai Brevier line height to be 18px, for footer & timestamp legibility

---

- [ ] Test engineer approval


packages/utilities/gel-foundations/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 3.3.2 | [PR#1758](https://github.com/bbc/psammead/pull/1758) Update Thai script type sizes for Minion and Brevier. |
```
andrew-nowak: the package.json says 3.2.2 ... looks like a previous PR bumped the CHANGELOG higher than the package.json

DenisHdz: Said that, is this a breaking change? Should we be updating the minor version instead?

sareh: Good spot! It was a typo. Fixed it now. https://www.npmjs.com/package/@bbc/gel-foundations

sareh: @DenisHdz I'd say it'd count as a bug fix - since the previous values were incorrect. So that's why I classed it as a patch version bump.

DenisHdz: In that case, LGTM :)

DenisHdz (approved):

ghost (approved):

dr3 (approved):

PR #1763, opened 2019-08-15
Allow Talos to be manually invoked 

### LINKED ISSUE #1757: Talos - Allow manual invoking
Talos allows PRs to be generated to bump dependancies. If we need to generate a PR, maybe because 1 failed to be made, or the PR is out of date, to get a new one would require bumping the package again to trigger it.

Add a param to the jenkins job which would emulate the packages.txt file, allowing for manually triggering talos easily

**Additional context**

dr3:
Resolves #1757

**Overall change:** Allow Talos to be manually invoked with a TALOS_PACKAGES param which is passed to talos

**Code changes:**

- Move Install to its own pipeline stage, to allow for skipping tests when invoking talos manually
- Add param and pass to talos
- Remove make command due to inability to pass arguements nicely
- Add support in talos for using argument string if set, else use fs like normal

---

- [ ] Test engineer approval


Jenkinsfile
```diff
@@ -30,8 +31,33 @@ pipeline {
...
+        always {
+          script {
+            stageName = env.STAGE_NAME
+          }
+        }
+      }
+    }
     stage ('Run application tests') {
+      when {
+        expression { params.TALOS_PACKAGES == '' }
```
sareh: Why do you want to skip running tests if Talos is run manually?

Jenkinsfile
```diff
@@ -107,8 +135,12 @@ pipeline {
...
         sh 'make setup-git'
         sh 'git fetch --all'
-        unstash 'psammead-publishes'
-        sh 'make talos'
+        script {
+          if (publishesStashed) {
+            unstash 'psammead-publishes'
+          }
+        }
+        sh "npm run talos ${params.TALOS_PACKAGES}"
```
sareh: This is quite insecure since anyone can pass in values into the TALOS_PACKAGES field and then have them passed to run on Jenkins. eg `; npx randompackagehere`.

Jenkinsfile
dr3: Just doesn't seem necessary IMO, you would manually invoke it on latest, where its already merged, so dont need to test it again

Jenkinsfile
sareh: I'd suggest running a short input validation script that checks all values passed in are namespaced under `@bbc/` and that errors with a helpful message if anyone passes in something else.

dr3: Means we can’t use this to bump other deps though :( like styled components

sareh: Good point. Passing the params to talos with an argument `-packages` would be preferable.

dr3:
Ready for re-review, using new jenkins format. Example of manual running can be found on https://ci.news.tools.bbc.co.uk/blue/organizations/jenkins/psammead/detail/ManualTalos-try/3/pipeline which uses a branch were `latest` is replaced with `ManualTalos-try`

Jenkinsfile
```diff
@@ -93,6 +103,7 @@ node {
                 notifySlack('danger', 'Aborted', gitCommitAuthor, 'Pipeline was aborted', gitCommitHash, gitCommitMessage, slackChannel)
                 break
               default:
+                echo "Unknown build result: ${currentBuild.result}"
```
dr3:
```suggestion
```

dr3:
New working https://ci.news.tools.bbc.co.uk/blue/organizations/jenkins/psammead/detail/ManualTalos-try2/2/pipeline

sareh:
@dr3 On that successful build I see this in the output:
```
currentBuild.currentResult: SUCCESS
currentBuild.result: null
```
The result of `null` is confusing. 
This is also found on the `latest` branch when Talos is run automatically. Could you please open an issue to address this?

Also, adding documentation to the Psammead readme would be useful.

Other than these things, I'm happy with this PR.

sareh (approved):

chris-hinds (approved):

PR #1837, opened 2019-08-20
fix storybook on ie11

### LINKED ISSUE #1730: Storybook does not run on IE11
**Describe the bug**
When opening a storybook that is running locally (ie. not bbc.github.io/psammead) in IE11, you will see

image

image

The real issue is that IE11 will not allow you to have a trailing comma in function calls
eg.
image

🙃 

This blocks a11y testing

It seems that when building storybook to be deployed, the code is run through babel and this trailing comma is removed because the error does not exist on bbc.github.io/psammead

temporary workaround would be to use `npm run build:storybook` and serve that directory (using eg. `python3 -m http.server`), but this is not great dx

andrew-nowak:
Resolves #1730 

**Overall change:** Make babel transpile as if it's production mode

**Code changes:**

- Set babel to transpile as if in production mode 
- This makes HMR _very slow_, so only do it when FIX_STORYBOOK_IE11 is 'true'

---

- [-] Automated (jest and/or cypress) tests added (for new features) or updated (for existing features)
- [-] This PR requires manual testing

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -10,5 +10,7 @@ module.exports = async ({ config, mode }) => {
     ),
   };
 
+  config.mode = 'production';
+
   return config;
 };
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -10,6 +10,9 @@ module.exports = async ({ config, mode }) => {
     ),
   };
 
+  // Yes, this is a gross compromise. However, running storybook in production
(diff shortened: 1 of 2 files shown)


ghost (approved):
Nice. Works

j-pendlebury (approved):
at least it works ⭐️

If someone finds a better solution then this can always be removed/updated.

DenisHdz (approved):

andrew-nowak:
This doesn't require manual testing

PR #1839, opened 2019-08-20
PoC moving locales, services, and removing Moment.js

### LINKED ISSUE #2898: Animation for navigation dropdown not working on IOS
**Describe the bug**
Currently, animation of a dropdown menu does not work on IOS (clicking on the hamburger button does not show the menu).
To see the bug you need to check Simorgh integration of new navigation, but the bug is most probably a Psammead issue.

**To Reproduce**

1. Go to any service frontPage with integrated navigation
2. Click on the hamburger menu button
3. See error

**Expected behaviour**
A dropdown menu slides down with animation when you click on the menu button.

- Device: iPhone
- OS: 12 and 13
- Browser: Safari

**Testing notes**

gabriellorin:
Resolves #2898 (Simorgh, https://github.com/bbc/simorgh/issues/2898)

**Overall change:** 

To render timestamps and convert to various time zones, we are currently using `Moment.js` and `Moment Timezone`.

`Moment.js` is great but we're only using a fraction of all the features available. For example, we're not using any of Moment's parsing features (we display dates from timestamps which don't need parsing), we don't need to know which day of the week, quarter, year, etc, a date is, we don't need to manipulate dates (add days, subtract months, etc.), etc.

On the client, these libraries weight around 59KB (minified, no gzip, excluding locale data / timezone), but also `Moment.js` is very tightly integrated and doesn't currently match our setup really well, as it relies on a global config for locales, timezones, etc, adding complexity / fragility to our setup. 

`Moment.js` has fantastic data, which would be great to re-use, but preferably with a much smaller date library, which would integrate better with our current setup, and would be more immutable / pure / without side effects / tree shakable.

This PoC PR attempts to re-creates support for some of Moment's various date format tokens (`dddd`, `ddd`, `dd`, `DD`, `Do`, `D`, `MMMM`, `MMM`, `MM`, `M`, `YYYY`, `YY`, `HH`, `H`, `mm`, `ss`, `A`, `a`, `h`, `Z`), as well as long date tokens (contained in locales: `LLLL`, `llll`, `LLL`, `lll`, `LL`, `ll`, `LTS`, `LT`, `L`), allowing us to display dates using data exported from  `Moment.js` and `Moment Timezone` ( `locale` files and `timezones`).

The PoC also allows us to display relative dates, using the same data (`locales`), for example `3 minutes ago` (using the `en-gb` locale).

The new pure date functions are:

`displayLocale(localeData, timezoneData, format, timestamp)`
`displayRelativeLocale(localeData, timestampStart, timestampEnd)`

`format` and `timestamp` input would be same as when using Moment. In `displayRelativeLocale()`, the parameter `timestampEnd` would often be `Date.now()`, but having the flexibility would make it easier for testing. Also, because Unix timestamps are all UTC (if a user in the UK runs `Date.now()` at the same time as a user in Brazil, they will get the same timestamp), `displayRelativeLocale()` doesn't have a `timezoneData` parameter.

To export each `locale` data automatically, this PoC uses a quick custom script with 2 regexes to output part of each file and make it an ES6 module export. After exporting, a locale file looks like (example for `en-gb`, original file: https://github.com/moment/moment/blob/2.24.0/locale/en-gb.js):

```javascript
export default { 
  months : 'January_February_March_April_May_June_July_August_September_October_November_December'.split('_'),
  monthsShort : 'Jan_Feb_Mar_Apr_May_Jun_Jul_Aug_Sep_Oct_Nov_Dec'.split('_'),
  weekdays : 'Sunday_Monday_Tuesday_Wednesday_Thursday_Friday_Saturday'.split('_'),
  weekdaysShort : 'Sun_Mon_Tue_Wed_Thu_Fri_Sat'.split('_'),
  weekdaysMin : 'Su_Mo_Tu_We_Th_Fr_Sa'.split('_'),
  longDateFormat : {
      LT : 'HH:mm',
      LTS : 'HH:mm:ss',
      L : 'DD/MM/YYYY',
      LL : 'D MMMM YYYY',
      LLL : 'D MMMM YYYY HH:mm',
      LLLL : 'dddd, D MMMM YYYY HH:mm'
  },
  ...
  relativeTime : {
      future : 'in %s',
      past : '%s ago',
      s : 'a few seconds',
      ss : '%d seconds',
      m : 'a minute',
      mm : '%d minutes',
      h : 'an hour',
      hh : '%d hours',
      d : 'a day',
      dd : '%d days',
      M : 'a month',
      MM : '%d months',
      y : 'a year',
      yy : '%d years'
  },
  ...
  ordinal : function (number) {
      var b = number % 10,
          output = (~~(number % 100 / 10) === 1) ? 'th' :
          (b === 1) ? 'st' :
          (b === 2) ? 'nd' :
          (b === 3) ? 'rd' : 'th';
      return number + output;
  },
  ...
};
```

This is all the data we need (and more) to output the dates. In some cases, there is also some helper functions. For example for Russian (original file: https://github.com/moment/moment/blob/2.24.0/locale/ru.js):

```javascript
function plural(word, num) {
  var forms = word.split('_');
  return num % 10 === 1 && num % 100 !== 11 ? forms[0] : (num % 10 >= 2 && num % 10 <= 4 && (num % 100 < 10 || num % 100 >= 20) ? forms[1] : forms[2]);
}
function relativeTimeWithPlural(number, withoutSuffix, key) {
  ...
}
var monthsParse = [/^янв/i, /^фев/i, /^мар/i, /^апр/i, /^ма[йя]/i, /^июн/i, /^июл/i, /^авг/i, /^сен/i, /^окт/i, /^ноя/i, /^дек/i];
    
export default { 
  months : {
      format: 'января_февраля_марта_апреля_мая_июня_июля_августа_сентября_октября_ноября_декабря'.split('_'),
      standalone: 'январь_февраль_март_апрель_май_июнь_июль_август_сентябрь_октябрь_ноябрь_декабрь'.split('_')
  },
  ...
```

We can also export each `timezone` data from `Moment Timezone`  (simply unpacking the original data and making it an ES6 module export). For example the unpacked `Europe/London` timezone data for 60 years looks like (`offsets` in minutes, and `untils` timestamps in ms; original data also includes time zone abbreviations which is not included here but could be; more info: https://momentjs.com/timezone/docs/#/zone-object/):

```javascript
const Europe_London = { name: 'Europe/London', offsets: [ -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, -60, 0, ], untils: [ 57722400000, 69818400000, 89172000000, 101268000000, 120621600000, 132717600000, 152071200000, 164167200000, 183520800000, 196221600000, 214970400000, 227671200000, 246420000000, 259120800000, 278474400000, 290570400000, 309924000000, 322020000000, 341373600000, 354675600000, 372819600000, 386125200000, 404269200000, 417574800000, 435718800000, 449024400000, 467773200000, 481078800000, 499222800000, 512528400000, 530672400000, 543978000000, 562122000000, 575427600000, 593571600000, 606877200000, 625626000000, 638326800000, 657075600000, 670381200000, 688525200000, 701830800000, 719974800000, 733280400000, 751424400000, 764730000000, 782874000000, 796179600000, 814323600000, 828234000000, 846378000000, 859683600000, 877827600000, 891133200000, 909277200000, 922582800000, 941331600000, 954032400000, 972781200000, 985482000000, 1004230800000, 1017536400000, 1035680400000, 1048986000000, 1067130000000, 1080435600000, 1099184400000, 1111885200000, 1130634000000, 1143334800000, 1162083600000, 1174784400000, 1193533200000, 1206838800000, 1224982800000, 1238288400000, 1256432400000, 1269738000000, 1288486800000, 1301187600000, 1319936400000, 1332637200000, 1351386000000, 1364691600000, 1382835600000, 1396141200000, 1414285200000, 1427590800000, 1445734800000, 1459040400000, 1477789200000, 1490490000000, 1509238800000, 1521939600000, 1540688400000, 1553994000000, 1572138000000, 1585443600000, 1603587600000, 1616893200000, 1635642000000, 1648342800000, 1667091600000, 1679792400000, 1698541200000, 1711846800000, 1729990800000, 1743296400000, 1761440400000, 1774746000000, 1792890000000, 1806195600000, 1824944400000, 1837645200000, 1856394000000, 1869094800000, 1887843600000, 1901149200000, 1919293200000, null, ], };
```

Adding entries to the `Psammead Locales` Storybook stories with the new date functions and various `localeData` and `localeTimezone` inputs, and comparing the output with `Moment.js` and `Moment Timezone`, we get:

The new date functions weight around 4.2KB (minified, no gzip), so savings should be around 54.52KB (compared to using  `Moment.js` and `Moment Timezone`), although the new functions would be using the unpacked version of timezones (we could also use the packed version and re-use the unpack function from `Moment Timezone`) which would add around 1KB, so saving would be slightly more than 50KB.

Once the code has been loaded/parsed/executed by the browser, a quick benchmark between Moment vs the custom date functions, to output 20 dates:
```
5.25ms (for Moment) vs 9.11ms (new function displayLocale), to output date from tokens
6.09ms (for Moment) vs 0.48ms (new function displayRelativeLocale), to output a relative date.
```

The fact `displayLocale()` is currently slower than Moment is actually normal, as the current implementation is doing multiple passes to convert tokens to an output (each token from the format string is first converted to a temporary token, to avoid having unwanted replacements in escaped characters or subsequent replacements, etc). Also, to find the correct offset in the timezone sorted array, it's currently not using binary search, etc. If necessary, we could make it much faster.

**Code changes:**

- Moved service configs from Simorgh to Psammead `packages/utilities/psammead-services/services/`, to make sure service configs can be re-used more easily (for testing, Storybook stories, etc).

- Extracted each Moment locale and saved to `packages/utilities/psammead-services/locales/moment-extract/` (script: `packages/utilities/psammead-services/extract-moment-data.js`). While extracting, added an ES6 module export, so that the locale data can easily be imported and set as a new property `localeData` on the service config object.

- Extracted and unpacked all timezones, and saved to a single file with named ES6 module exports: `packages/utilities/psammead-services/timezones/timezones-data.js` (script is trivial, outside of repo; and ultimately would be better to have separate files).

- Imported (ES6 module import) extracted locale and timezone to a few selected services configs, set as new properties `localeData` and `timezoneData`.

- Created new functions `displayLocale` and `displayRelativeLocale` to display dates using `localeData` and `timezoneData`, supporting Moment tokens format `dddd`, `ddd`, `dd`, `DD`, `Do`, `D`, `MMMM`, `MMM`, `MM`, `M`, `YYYY`, `YY`, `HH`, `H`, `mm`, `ss`, `A`, `a`, `h`, `Z`, `LLLL`, `llll`, `LLL`, `lll`, `LL`, `ll`, `LTS`, `LT`, `L`: `packages/utilities/psammead-services/utils/convertToLocale.js`.

- Modified Storybook stories: `packages/utilities/psammead-locales/moment/index.stories.jsx` to include various locales/timezones and compare with output from Moment.

  Changed entries that were using Moment's `.fromNow()` method to use `.from(now)` instead, with `now` being defined as a timestamp (for example `const now = 1565787354967`), making sure the output is predictable (instead of relying on `Date.now()`, which can create issues). This would also make testing easier, as the input for the tests would be fixed instead of always changing.

- Added support of new functions to Timestamp container (`packages/containers/psammead-timestamp-container/src/index.jsx`), but later commented out as Storybook seems to currently be having some issues with Moment (unrelated to this PoC).

(This is just a PoC and there would too many changes to do all at once, so this is just to have an idea and see if it would be worth it 😊)

---


dr3 (commented):
All this data in this PR has been copied from moment, which means we instantly lose any updates to it that the community contributes, is it not possible to use moments data directly? even if it means stubbing moment a little and importing from files like `moment/locale/en-gb.js`?

gabriellorin:
Hey Drew! 😊
For the locales the data is actually being extracted locally from `../../../node_modules/moment/locale/${localeName}.js` (with a `require('moment');` at the top to make sure Moment is installed) when we run the script: `packages/utilities/psammead-services/extract-moment-data.js`, https://github.com/bbc/psammead/pull/1839/files#diff-ff27c19c8064d117c2c73723a1e8e25b 
So we can just run this script again when a new version comes out, it shouldn't be an issue (and probably the safest way!). The regex in the script is very simple so we should still be benefitting from changes made by the community! 😊

dr3:
@gabriellorin that’s a very manual process, would it not be possible to do this on the fly? Means we don’t have to update this psammead component every time moment release and can instead just bump the peer dep

gabriellorin:
@dr3 We sure could but last commit to Moment locales was 7 month ago (https://github.com/moment/moment/tree/develop/locale), so not sure it would be worth it 😁.

dr3:
Closing for the time being, will come back to this at a later date

PR #1841, opened 2019-08-21
Modify month names/numerals for Arabic locale

### LINKED ISSUE #1778: Extend arabic locale to psammead locales.
We have decided that the moment translations for Arabic aren't suitable for editorial and may need to extend it to suit our requirements

We need to add an arabic file to psammead-locales https://github.com/bbc/psammead/tree/latest/packages/utilities/psammead-locales/moment which updates the `ar` locale
The update should ensure we include:

1. The double month name (This can be seen on an existing Arabic page eg. https://www.bbc.com/arabic/middleeast-49365997)
2. Western Arabic numerals (i.e. 12345678...)
3. Be right-to-left
4. Remember to update relevant documentations

 

**Testing notes**
- Unit tests needed to cover locales in psammead

**Additional context**

RayNjeri:
Resolves #1778

**Overall change:** _Extends the `ar` locale to suit our desired requirements._

**Code changes:**

- _Updates the ar locale with:_

1. Double Month Names
2. Western numerals

**Manual Testing**
- _Navigate to http://localhost:8180/?path=/story/utilities-psammead-locales--moment-arabic-ar:_


packages/utilities/psammead-locales/moment/ar.js
```diff
@@ -0,0 +1,65 @@
...
+  'مايو/ أيار',
+  'يونيو/ حزيران',
+  'يوليو/ تموز',
+  'أغسطس/ آب',
+  'سبتمبر/ أيلول',
+  'أكتوبر/ تشرين الأول',
+  'نوفمبر/ تشرين الثاني',
+  'ديسمبر/ كانون الأول',
+];
+const symbolMap = {
```
12: Are we meant to be mapping numbers to themselves?

RayNjeri: yes i think we do otherwise eastern arabic numerals will be maintained which is not our desired behaviour

dr3:
```
.replace(/[1234567890]/g, function(match) {
      return numberMap[match];
})
```
Appears to be finding numbers, then replacing them with the value in this Map, but as far as I can tell, its replacing them with themselves e.g `'1': '1',` are the same character

RayNjeri: Actually we do not need this func, the `postformat` func ctaers for maintaining the western numerals

packages/utilities/psammead-locales/moment/ar.js
```diff
@@ -0,0 +1,46 @@
...
+  '4': '4',
+  '5': '5',
+  '6': '6',
+  '7': '7',
+  '8': '8',
+  '9': '9',
+  '0': '0',
+};
+
+moment.updateLocale('ar', {
```
EinsteinNjoroge, line 23: I'm not clear on what `postformat` does but this implementation produces the desired output without symbolMap

```
 /* eslint-disable func-names */
const moment = require('moment');
require('moment/locale/ar');

// these keys have been overridden to maintain western numerals
// some other keys have been left as set upstream
// see https://github.com/moment/moment/blob/develop/src/locale/ar.js

const months = [
  'يناير/ كانون الثاني',
  'فبراير/ شباط',
  'مارس/ آذار',
  'أبريل/ نيسان',
  'مايو/ أيار',
  'يونيو/ حزيران',
  'يوليو/ تموز',
  'أغسطس/ آب',
  'سبتمبر/ أيلول',
  'أكتوبر/ تشرين الأول',
  'نوفمبر/ تشرين الثاني',
  'ديسمبر/ كانون الأول',
];

moment.updateLocale('ar', {
  months,
  monthsShort: months,
  postformat(string) {
    return string.replace(/,/g, '،');
  },
});

```

RayNjeri: Had  to use symbolMap to replace the eastern arabic numerals with western numerals but tried this implementation seems to produce the desired effect too

jamesdonoh: I'm uneasy with this solution for numerals as a) it's cryptic and b) could have unintended effects since `postformat is _called after formatting on every string_ (https://momentjs.com/docs/).

packages/utilities/psammead-locales/moment/ar.test.js
```diff
@@ -0,0 +1,259 @@
...
+
+import moment from 'moment';
+import './ar';
+
+moment.locale('ar');
+
+// This asset overrides the gunit assertion done in the moment codebase.
+// Format and styling of this file has been keep consistent with the official moment tests.
+// An example of these tests can be seen at https://github.com/moment/moment/blob/develop/src/test/locale/en-gb.js
+const assert = { equal: (val1, val2) => expect(val1).toEqual(val2) };
```
jamesdonoh, line 11: Why is this needed? I know it's in `yo.js` but I want to be sure it is required rather than just blindly copying it, and if it is needed I think the comment should explain why.

packages/utilities/psammead-locales/moment/ar.test.js
```diff
@@ -0,0 +1,259 @@
...
+  'مايو/ أيار',
+  'يونيو/ حزيران',
+  'يوليو/ تموز',
+  'أغسطس/ آب',
+  'سبتمبر/ أيلول',
+  'أكتوبر/ تشرين الأول',
+  'نوفمبر/ تشرين الثاني',
+  'ديسمبر/ كانون الأول',
+];
+
```
jamesdonoh, line 31: Please could we have a comment here explaining the origin of the tests that follow? Similar to https://github.com/bbc/psammead/blob/60c0d36e3104507e1bfbb51b9df59e2d4f5692f8/packages/utilities/psammead-locales/moment/yo.test.js#L12-L14

packages/utilities/psammead-locales/moment/ar.js
```diff
@@ -0,0 +1,29 @@
+const moment = require('moment');
+require('moment/locale/ar');
+
+// these keys have been overridden to maintain western numerals
```
jamesdonoh: To be explicit: `...and to ensure that double month names are used`?

Maybe we could move this to L24 for consistency with `yo.js` too.

jamesdonoh (changes requested):

packages/utilities/psammead-locales/moment/ar.js
```diff
@@ -0,0 +1,29 @@
...
+// some other keys have been left as set upstream
+// see https://github.com/moment/moment/blob/develop/src/locale/ar.js
+
+const months = [
```
jamesdonoh: Suggest calling this `dualMonthNames` for readability.

packages/utilities/psammead-locales/moment/ar.test.js
```diff
@@ -0,0 +1,259 @@
...
+  'مارس/ آذار',
+  'أبريل/ نيسان',
+  'مايو/ أيار',
+  'يونيو/ حزيران',
+  'يوليو/ تموز',
+  'أغسطس/ آب',
+  'سبتمبر/ أيلول',
+  'أكتوبر/ تشرين الأول',
+  'نوفمبر/ تشرين الثاني',
+  'ديسمبر/ كانون الأول',
```
jamesdonoh, line 29: Have you confirmed that these render correctly (== like legacy PAL pages) inside a `dir=rtl` element?

packages/utilities/psammead-locales/moment/ar.test.js
RayNjeri: if not used, the tests fails with this error `assert.equal is not a function`

packages/utilities/psammead-locales/moment/ar.test.js
RayNjeri: yes i have

packages/utilities/psammead-locales/moment/ar.js
RayNjeri: IMO i think its better for the comment to be on L4 unlike L24 as it describes the changes that have been over ridden which happens on the months declaration. which is more or less the same concept that is on `yo.js`

jamesdonoh: It seems confusing to me because it mentions 'keys' here but there are no keys, whereas in the original scope there were.

packages/utilities/psammead-locales/moment/ar.test.js
jamesdonoh: Clarified with @aacn500 that this was done originally in order to make it potentially easier to contribute the tests back to Moment.js without rewriting them.

packages/utilities/psammead-locales/moment/ar.js
RayNjeri: updated the comment

packages/utilities/psammead-locales/moment/ar.test.js
RayNjeri: done

packages/utilities/psammead-locales/moment/ar.js
RayNjeri: for now the only effect `postformat` has is to replace the coma with the arabic coma,
I have counter checked with `ur` locale which has almost our intended output as for `ar` and seems i has no other effect https://github.com/moment/moment/blob/13a61b285c095bda7ea8e33156090ea5ccfeaef1/locale/ur.js#L83

packages/utilities/psammead-locales/moment/ar.js
```diff
@@ -0,0 +1,29 @@
...
+  'أكتوبر/ تشرين الأول',
+  'نوفمبر/ تشرين الثاني',
+  'ديسمبر/ كانون الأول',
+];
+
+moment.updateLocale('ar', {
+  months: dualMonthNames,
+  monthsShort: dualMonthNames,
+  postformat(string) {
+    return string.replace(/,/g, '،');
```
jamesdonoh, line 28:
```suggestion
    // Suppress conversion to Eastern numerals from base `ar` locale but retain use of Arabic comma
    return string.replace(/,/g, '،');
```

packages/utilities/psammead-locales/moment/ar.js
RayNjeri: `postformat` function returns the western numerals unless overridden as was the case initially in the `ar` locale
```
postformat: function (string) {
        return string.replace(/\d/g, function (match) {
            return symbolMap[match];
        }).replace(/,/g, '،');
    },
```
which in this case `symbol map` mapped the western numerals to arabic numerals.
So basically we are overriding  the postformat from the base locale to use the western numerals

EinsteinNjoroge (approved):

jamesdonoh (approved):

dr3 (approved):

staylos92:
LGTM :+1:

Issue #774, opened 2019-07-04
Remove use of react-test-renderer


jroebu14:
We currently have 2 test renderers in Psammead and Simorgh - `react-test-renderer` and `@testing-library/react`.

Making use of 2 test renderers will lead to confusion about which one to use and when, especially when they have different ideas about how to write tests.

## Solution
Remove `react-test-renderer` and all use of it. Refactor shallow render tests because `@testing-library/react` does not have or encourage shallow rendering.

---

Some of the benefits of using `@testing-library/react` include:
 - encourages better testing practices and writing more maintainable tests 
 - is designed to enable and encourage writing tests that use components as the end users use them
 - it supports querying for elements in ways that encourage accessibility in components

Find out more about [@testing-library/react](https://testing-library.com/docs/react-testing-library/intro).

This article briefly describes and compares React test renders https://codeburst.io/revisiting-react-testing-in-2019-ee72bb5346f4

Here is more on why [`@testing-library/react` doesn't do shallow rendering](https://kentcdodds.com/blog/why-i-never-use-shallow-rendering).

PR #1917, opened 2019-08-27
Remove react-test-renderer

ghost:
Contributes to #774

**Overall change:** Remove `react-test-renderer` and  shallow rendering for `psammead-test-helpers`

**Code changes:**

- Remove `shallowRender` and `shouldShallowMatchSnapshot`.
- Update `isNull` and `shouldMatchSnapshot`
- Use `render` from `@testing-library/react`

---


thekp:
Hi @sadickisaac 

I see you are removing the shallowRendering functionality. We currently have an on-going issue to make use of the `psammead-test-helper` in Simorgh, however the shallow rendering is heavily used in Simorgh for the page body tests (i.e `simorgh/src/app/containers/ArticleMain`: https://github.com/bbc/simorgh/blob/latest/src/app/containers/ArticleMain/index.test.jsx#L35-L54)

Do you know if  `@testing-library/react` package has an equivalent "shallow rendering functionality" we can use ?

thekp:
source of the CI issue btw: https://github.com/bbc/psammead/blob/latest/packages/components/psammead-media-player/src/Placeholder/index.test.jsx#L3

ghost:
> Do you know if @testing-library/react package has an equivalent "shallow rendering functionality" we can use ?

`@testing-library/react` does not support shallow rendering. We are trying to move from using shallow rendering. What will happen is, the test depending on `shallowRendering` will be refactored.

thekp:
> @testing-library/react does not support shallow rendering. We are trying to move from using shallow rendering. What will happen is, the test depending on shallowRendering will be refactored.

possibly major refactors to our Simorgh tests 👀(but I suppose it is inevitable since we want to move away from shallowRendering) I'm going to put my running issue on blocked until you get this one done: https://github.com/bbc/simorgh/issues/3264 👍

packages/utilities/psammead-test-helpers/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 2.0.0 | [PR#1917](https://github.com/bbc/psammead/pull/1917) Remove react-test-renderer and shallow rendering |
 | 1.3.2 | [PR#1919](https://github.com/bbc/psammead/pull/1919) Use @testing-library-react |
```
thekp:
```suggestion
| 1.3.2 | [PR#1919](https://github.com/bbc/psammead/pull/1919) Accidental version bump |
```

thekp (approved):
This is ok for psammead, as we don't use shallowRender anywhere here.

This issue is for using test-helper in Simorgh: https://github.com/bbc/simorgh/issues/3264

DenisHdz (approved):

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -11,25 +9,10 @@ export const shouldMatchSnapshot = (title, component) => {
...
-    expect(tree).toMatchSnapshot();
-  });
-};
-
 export const isNull = (title, component) => {
   it(title, () => {
-    const tree = renderer.create(component).toJSON();
-    expect(tree).toBeNull();
+    const { container } = render(component);
+    expect(container.firstChild).toBeNull();
```
dr3, line 15: Is this the same thing?  is first child always the top level thing? could you have a second child that was not null?

ghost: How would that look like? First child null and second not null. AFAIK `null` is not rendered which makes the second child that was not null to be the firstChild.

ghost: Maybe we could check if the container has children and assert that there are no children rendered.

dr3 (approved):

PR #1933, opened 2019-08-28
Changes to Timestamp Container stories, Moment locales stories

### LINKED ISSUE #1836: Use fixed timestamp for psammead storybook components
**Describe the bug**
`psammead-timestamp-container` is currently using a dynamic date, this is an issue on Chromatic QA as it will always say there is change on the snapshot.

`psammead-locales` also uses a dynamic date.

**To Reproduce**
Example of Psammead chromatic build: https://www.chromaticqa.com/build?appId=5d28eb5ee163f6002046d6fb&number=734

To trigger chromatic on local psammead branch: 

 - Push changes and make a PR
or 
- `npm run test:chromatic` from your local terminal

**Expected behaviour**
The date should not be dynamic so that it can pass on chromatic QA

**Testing notes**

Review Chromatic QA build

**Additional context**

gabriellorin:
Resolves #1836

**Overall change:** 

The Timestamp Container and Moment locales stories were using user's current timestamp, causing issues with Chromatic (new date/time for every build), so changed it to use a fixed timestamp by default (timestamp for 27 August 2019, 14:54 BST).

Also the Timestamp Container stories input buttons were not working (not updating date/time), so changed the code to use a `select` storybook input instead. And changed the UX a bit: now it doesn't show the `timestamp` `number` input unless the user selects `Custom timestamp` from the `select` input. 

(And added a `text` input for the date format, so that the user can try different formats.)

Before:

screenshots-before

After:

screenshots-after

(The Timestamp Container story had the `isRelative` set to true by default. Although we could get this default value to work with Chromatic, it would require an additional `Timestamp now` input: this would be a fixed timestamp, so that the difference between the fixed timestamp and the fixed now timestamp would always be identical. We would also need to add parameters `timestampNow` to the Timestamp Container, modify the  `showRelativeTime` function to use Moment's `.from(timestampNow)` instead of `.fromNow()`. For this PR, to keep it simple, I just changed the default value to `false` .)

**Code changes:**

- Changed the Timestamp Container stories to use a `select` instead of `button` (fixing issue with date/time not updating on button click), default timestamp input is now a fixed timestamp.
- Added `number` input if user selects entry `Custom timestamp` (it's the only entry with timestamp value `0`) from `select` input; by default the value for this input is set to the user's current timestamp.
- Added `text` input to allow user to change date format.
- Modified Moment locales stories to use a fixed timestamps: now using `moment(fixedTimestamp)` instead of `moment()`, `.from(fixedTimestamp)` instead of `.fromNow()`, and `.calendar(fixedTimestamp)` instead of  `.calendar()`.

---


dr3 (dismissed):
ignore this

dr3:
To get relative timestamps to work, i recommend using timemachine, like we do in simorgh https://github.com/bbc/simorgh/blob/latest/.storybook/config.js#L10-L13

dr3:
If you leave storybook open for a min, the timestamps go out of sync with what theyre meant to show
image

also rather than having a hardcoded date, you can use timemachine https://github.com/bbc/simorgh/blob/latest/.storybook/config.js#L10-L13 instead, and continue to use the date.now stuff

dr3 (changes requested):
If you leave storybook open for a min, the timestamps go out of sync with what theyre meant to show
image

also rather than having a hardcoded date, you can use timemachine https://github.com/bbc/simorgh/blob/latest/.storybook/config.js#L10-L13 instead, and continue to use the date.now stuff

gabriellorin:
Hi @dr3 thanks for the feedback! 😊
For the out of sync timestamps it should now be fixed. Because the Storybook `select` knob keeps track of the selected item by value (not by key, which is an issue when values get updated), had to wrap each value in an object to make sure the reference stays the same. And now at each render / input change it updates all timestamps in a mutable way to avoid changing object references.

For the use of `timemachine`, personally not a big fan of global objects mutations but could be a way 😁

Thanks!

thekp:
@dr3 I agree with @gabriellorin here, regarding using `timemachine`. 

My use case for it in Simorgh was that we were using a generator: https://github.com/bbc/simorgh/blob/latest/src/app/containers/ArticleTimestamp/index.stories.jsx#L4

and instead of updating the generator (and breaking other tests), I used timemachine to update the global instance of Date() for our storybook application.

In this case, the timestamps were coded to use current time within storybook.

packages/containers/psammead-timestamp-container/src/index.stories.jsx
```diff
@@ -15,110 +15,170 @@ import '@bbc/psammead-locales/moment/ig';
...
-        isRelative={boolean('isRelative', false)}
-        script={script}
-        locale={locale}
-        service={service}
-      />
-    )),
+    inputProvider([], ({ locale, script, service }) => {
+      updateAllTimestamps();
+
+      const storyTimestamp = select(
```
dr3: This is a lot of duplication of 
```
const storyTimestamp = select(
        'Timestamp',
        timestamps,
        timestamps['Fixed (27 Aug 2019)'],
      );
```
Can this be in a wider scope?

thekp (approved):

packages/containers/psammead-timestamp-container/src/index.stories.jsx
```diff
@@ -15,110 +15,161 @@ import '@bbc/psammead-locales/moment/ig';
...
-      .add(2, 'second');
-  });
-  button('9 months ago', () => {
-    timestamp = moment().subtract(9, 'months');
+const val = momentTimestamp => momentTimestamp.valueOf();
+
+const timestamps = {
+  'Fixed (27 Aug 2019)': { timestamp: undefined, update: () => fixedTimestamp },
+  Now: { timestamp: undefined, update: () => Date.now() },
+  '5 seconds ago': {
```
dr3, line 26: I cant see these anymore :(
image

dr3 (dismissed):

packages/containers/psammead-timestamp-container/src/index.stories.jsx
oluoluoxenfree: @drew does it work for you if you put the `const`s back where they were? It doesn't seem to for me :/

thekp (approved):
👍

EinsteinNjoroge (dismissed):
Just a few more changes...
You need to update version in `package.json` and `package-lock.json` for both `psammead-timestamp-container` and `psammead-locales`

oluoluoxenfree:
@EinsteinCarrey I can't see the problems you mean? Did someone already resolve them? (never mind, hadn't pulled!)

thekp (approved):

packages/containers/psammead-timestamp-container/src/index.stories.jsx
chris-hinds: These are showing in storybook for me.

thekp: @gabriellorin resolved the comment.

chris-hinds (approved):
LGTM ✅  works well in local storybook

dr3 (approved):

PR #1958, opened 2019-09-02
update psammead-test-helper

thekp:
related to: https://github.com/bbc/psammead/issues/774

**Overall change:** _Currently we are only checking the first child of the container. Sometimes components may have multiple children so we need to check the entire component tree._

**Code changes:**

- add logic that checks the rendered container if there are multiple children or not_

---


dr3 (approved):

FK78 (approved):

thekp:
if we remove `.firstChild` it will then wrap components being tested with an extra `<div>`

i.e without `.firstChild`: 
```
<div>
  <figcaption
    class="c0"
  >
    This is some Caption text
    <span
      class="c1"
    >
      Some offscreen text
    </span>
  </figcaption>
</div>
```

with `.firstChild`:
```
  <figcaption
    class="c0"
  >
    This is some Caption text
    <span
      class="c1"
    >
      Some offscreen text
    </span>
  </figcaption>
```

packages/utilities/psammead-test-helpers/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 2.0.1 | [PR#1917](https://github.com/bbc/psammead/pull/1917) Update `shouldMatchSnapshot` to match entire |
```
thekp:
```suggestion
| 2.0.1 | [PR#1958](https://github.com/bbc/psammead/pull/1958) Update `shouldMatchSnapshot` to match entire |
```

chris-hinds (approved):

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -5,7 +5,7 @@ import deepClone from 'ramda/src/clone';
 export const shouldMatchSnapshot = (title, component) => {
   it(title, () => {
     const { container } = render(component);
-    expect(container.firstChild).toMatchSnapshot();
```
jroebu14, line 8: What about something like

```js
const hasOneChild = container.children.length === 1
const removeWrappingDiv = container => container.firstChild

expect(hasOneChild ? removeWrappingDiv(container) : container).toMatchSnapshot();
```

thekp: Thanks for the help @jroebu14 with this logic I don't have to go updating every snapshot in psammead 😅 

Also we tested it against this PR: https://github.com/bbc/simorgh/pull/3375 by copying the `esm` and `dist` of `psammead-test-helper` directly into the Simorgh node_modules and it passes and fixes the snapshot issue we were having. 👍 

Thanks again 💯

FK78 (approved):

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -5,7 +5,11 @@ import deepClone from 'ramda/src/clone';
...
+    const hasOneChild = container.children.length === 1;
+    const removeWrappingDiv = containerTwo => containerTwo.firstChild;
```
dr3: Can you add  comment here explaining why please :)

jroebu14: why `containerTwo`?

FK78: @jroebu14 ESLint did not like that 'container' was already declared in the upper scope.

jroebu14: ah ok. it should be fine if you move the `removeWrappingDiv` function to the line before `const { container } = render(component);`

FK78 (approved):

dr3 (approved):

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -4,8 +4,13 @@ import deepClone from 'ramda/src/clone';
 
 export const shouldMatchSnapshot = (title, component) => {
   it(title, () => {
+    // Components with a single child do not have to be wrapped in a div when matching against a snapshot.
```
jroebu14, line 7: I think a more appropriate comment would be something like

`// select the first child to remove the pointless wrapping div from snapshots`

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -4,8 +4,13 @@ import deepClone from 'ramda/src/clone';
...
+    const removeWrappingDiv = container => container.firstChild;
     const { container } = render(component);
-    expect(container.firstChild).toMatchSnapshot();
+    const hasOneChild = container.children.length === 1;
```
jroebu14, line 10: and then here 

`// if the container has more than one child then it's a component that uses a fragment at the top level so we should not select the first child because it wouldn't snapshot the whole component`

I'm in a rush so couldn't think of a more concise comment!

thekp: I'll add it as a separate PR, but I don't think it needs to be any more concise as it is very precise

thekp: https://github.com/bbc/psammead/pull/1994

PR with the comment updates ^

PR #1964, opened 2019-09-03
Talos - Bump @bbc/psammead-image-placeholder, @bbc/psammead-media-player, @bbc/psammead-timestamp-container

bbc-news-frameworks:
@bbc/psammead-image-placeholder
@bbc/psammead-media-player
@bbc/psammead-timestamp-container

psammead


dr3:
replaced with https://github.com/bbc/psammead/pull/1993

PR #1987, opened 2019-09-04
Fix StoryPromo font size on IE11

### LINKED ISSUE #1988: IE11 does not support nested media queries
**Describe the bug**
IE11 does not support nested media queries

**To Reproduce**
This was highlighted in #1987 where the previously nested media query font-size and line-height were not being applied to the screen width specified in the nested media query in IE11 but was being applied in Chrome. 

**Expected behaviour**
All styling for the different screen widths should be applied similarly across supported browsers. 

Fixing this might mean refactoring code so that media queries do not need to be nested in the future (this is what was done in #1987 )

**Screenshots**

**Testing notes**

**Additional context**

Bopchy:
Resolves #1988 

**Overall change:** _Fixes StoryPromo font size appearing smaller in IE11._

**Code changes:**

- _Get `font-size` and `line-height` from script._
- _Update snapshots._

---


packages/components/psammead-story-promo/src/index.jsx
```diff
@@ -149,7 +148,16 @@ export const Headline = styled.h3`
 
   @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
     ${({ script, topStory }) =>
-      script && (topStory ? getParagon(script) : getGreatPrimer(script))}
+      script &&
+      (topStory
+        ? css`
+            font-size: ${script.paragon.groupD.fontSize / 16}rem;
```
dr3: Howcome this is needed? Do the `getGreatPrimer` and `getParagon` functions not work as expected? Or are we trying to use 1 group over 2 breakpoints?

Bopchy: It's needed for IE11 - because it wasn't picking up on the font-size and line-height specified here https://github.com/bbc/psammead/blob/f45a59d02dcd9f9120da9d9b6101dc7f2cdf37c2/packages/utilities/gel-foundations/src/getTypeSizes.js#L30 since what is returned by either getGreatPrimer() or getParagon() will have a nested media-query, and IE11 doesn't support nested media queries

dr3: Ohhhh hmmm thats annoying of IE :P

dr3: Ohhhh hmmm thats annoying of IE :P Any opinions @sareh @pjlee11 @jblaut? This PR works, but if this is gunna keep coming up, we may need to reevaluate our font size logic

pjlee11: Is it only nested media queries that IE11 doesn't support? I'm interested in how we currently do this on the PAL, maybe we only have a single level of media queries ...

dr3: It feels unique to story promo, where it switches typography at different breakpoints, which is  very uncommon, maybe were ok with it being in this 1 place?

packages/components/psammead-story-promo/src/index.jsx
```diff
@@ -149,7 +148,16 @@ export const Headline = styled.h3`
 
   @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
     ${({ script, topStory }) =>
-      script && (topStory ? getParagon(script) : getGreatPrimer(script))}
+      script &&
+      (topStory
+        ? css`
```
ghost: Could be pull this out to something like 
```js
const getHeadlineFontStyle = (script, topStory) => {
  const type = topStory ? 'paragon' : 'greatPrimer';
  
  const lineHeight = script[type].groupD.fontSize / 16
  const fontSize = script[type].groupD.fontSize / 16

  return css`
  font-size: ${lineHeight}rem;
  line-height: ${fontSize}rem; 
  `
}
```

ghost (changes requested):
Resolve conflicts

packages/components/psammead-story-promo/src/index.jsx
```diff
@@ -139,6 +138,16 @@ const InlineMediaIndicator = styled.div`
       `}
 `;
 
+const getHeadlineFontStyle = (script, type) => {
```
dr3: Above this line, can you write a comment about why this is needed please :)

dr3 (commented):
LGTM after the comment, not ideal to have an edge case, but a good solution seeing as this is needed :) nice work :D

packages/components/psammead-story-promo/src/index.jsx
```diff
@@ -149,7 +158,10 @@ export const Headline = styled.h3`
 
   @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
     ${({ script, topStory }) =>
-      script && (topStory ? getParagon(script) : getGreatPrimer(script))}
+      script &&
```
ghost: With this https://github.com/bbc/psammead/pull/1987#discussion_r321116701 suggestion you could simplify this to

```js
${({ script, topStory }) => script && getHeadlineFontStyle(script, topStory)}
```

ghost (approved):

dr3 (approved):

PR #1992, opened 2019-09-04
Update the psammead-locale package

RayNjeri:

Resolves chromaticqa failures on IE11 for https://github.com/bbc/simorgh/pull/3468 

**Overall change:** _updates the package.json._

**Code changes:**

- _Updates the package.json with a sideEffects flag of the module it is bundling.._

---


packages/utilities/psammead-locales/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@bbc/psammead-locales",
   "version": "2.1.0",
   "description": "A collection of locale configs, used in BBC World Service sites",
+  "main": "dist/index.js",
```
dr3: when psammead is built these files dont exist :( Im not sure what the IE11 issue is, maybe the `"sideEffects": false,` fixes it? Not sure why though :(  Sorry I cant be more helpful

dr3 (changes requested):

packages/utilities/psammead-locales/package.json
dr3: Happy to try merging just the `"sideEffects": false,` and see if that helps?

RayNjeri: yeah adding just this `"sideEffects": false` should fix it.
Testing it now, to be sure

thekp (approved):

packages/utilities/psammead-locales/package.json
jamesdonoh: Does anyone know if there is a reason why sideEffects: false should not be specified for all packages?

RayNjeri: @dr3 adding just the `sideEffects: false` flag does the trick,
Tested it out locally

thekp (commented):
You will have to bump the version and update the changelog 👀

EinsteinNjoroge (approved):

jamesdonoh (approved):
Thanks to all who helped figure this one out.

rhenshaw56 (approved):

dr3 (approved):

packages/utilities/psammead-locales/package.json
dr3: @jamesdonoh We only really added it where needed, alot of packages are accessed via a nested `package.json` so doesnt need it at the base `package.json`, we just missed it with moment

PR #2003, opened 2019-09-05
 Update psammead-locale package to remove "side-effect"

RayNjeri:
Reverts bbc/psammead#1992

### DIFF
--- a/packages/utilities/psammead-locales/CHANGELOG.md
+++ b/packages/utilities/psammead-locales/CHANGELOG.md
@@ -3,6 +3,8 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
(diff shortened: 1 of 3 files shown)


packages/utilities/psammead-locales/CHANGELOG.md
```diff
@@ -3,7 +3,6 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
-| 2.1.1 | [PR#1992](https://github.com/bbc/psammead/pull/1992) update the package with the side-effects flag |
```
dr3, line 6: Sadly because of how psammead publishes, we need to treat this a fix forward, not a revert, so we need a new version number (2.1.2), with a fresh changelog entry

dr3 (approved):

jamesdonoh (approved):

PR #2004, opened 2019-09-05
Talos - Bump @bbc/psammead-locales

bbc-news-frameworks:
@bbc/psammead-locales

psammead


dr3 (approved):

jamesdonoh (approved):

jamesdonoh:
Closing this as we need to fix psammead-locales for IE11 so this bump is pointless.

PR #2013, opened 2019-09-05
psammead-locale bug fix for IE11

RayNjeri:
Signed-off-by: Rachael Njeri <rachael.njeri@andela.com>

Resolves #2005

**Overall change:** _Disables eslint rules on multi-line split for IE11 support_

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Automated (jest and/or cypress) tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

Issue #2005, opened 2019-09-05
IE11 does not support multi-line split psammead-locale


RayNjeri:
**Describe the bug**
IE11 does not support just the multi-line split locale from psammead-locales

**To Reproduce**
This was highlighted in https://github.com/bbc/simorgh/pull/3468/ where stroybook failed to load in IE11

1. View simorgh's  storybook on IE11 or  Run chromaticqa locally 

**Expected behaviour**
Storybook should be able to evaluate and render properly in IE11 while using `@bbc/psammead-locales/moment/ar`

**Screenshots**
![Screenshot 2019-09-05 at 11 45 56 AM](https://user-images.githubusercontent.com/27012452/64326528-ce231c00-cfd2-11e9-81e2-757267ab210b.png)

**Testing notes**

- [ ] This bug fix is expected to need manual testing.

**Additional context**
Blocks https://github.com/bbc/simorgh/pull/3468

PR #2016, opened 2019-09-05
Avoid more trailing commas in function calls for IE11

jamesdonoh:
Resolves [NO ISSUE] but relates to https://github.com/bbc/psammead/issues/2005

Follows on from https://github.com/bbc/psammead/pull/2013

**Overall change:** _Fixes some more trailing commas in function args that would break on IE11 and adds linter rules to reduce chance of recurrence._

**Code changes:**

- Add .prettierrc suitable for Moment locales (as these are not transpiled)
- Remove remaining commas that would cause compatibility issues

---


packages/utilities/psammead-locales/moment/.eslintrc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
jamesbhobbs: Why not in base eslint file?

jamesdonoh: Because I assume we want to maintain our use of dangling commas everywhere else than here, the moment locales are not transpiled before being run on the client however so need special treatment.

We could move away from dangling commas everywhere if we want but would be a major change.

jamesbhobbs (approved):

packages/utilities/psammead-locales/moment/.eslintrc.js
```diff
@@ -0,0 +1,12 @@
...
+  rules: {
+    // IE11 gives syntax error for trailing commas on functions and Moment
+    // locales are not transpiled
+    'comma-dangle': ['error', {
```
dr3: Can we use the `es5` setting? https://prettier.io/docs/en/options.html#trailing-commas

jamesdonoh: That's a much better idea! Updated.

dr3 (approved):
LGTM :)

PR #2020, opened 2019-09-06
IE11 ES6 fixes

RayNjeri:

Resolves [NO ISSUE] but relates to #2016 

Follows on from #2016

Overall change: Reverts some es6 linting to ES5

### DIFF
--- a/packages/utilities/psammead-locales/moment/fa.js
+++ b/packages/utilities/psammead-locales/moment/fa.js
@@ -24,7 +24,7 @@ const jalaliFormats = ['D MMMM YYYY', 'LL'];
 moment().locale('fa');
 moment.updateLocale('fa', {
   // eslint-disable-next-line object-shorthand
-  postforma: function(string) {
+  postformat: function(string) {
     const str = jalaaliHelper.addJalaliDate(
       'fa',
       persianJalaliMonths,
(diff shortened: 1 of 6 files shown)


packages/utilities/psammead-locales/moment/fa.js
```diff
@@ -23,7 +23,8 @@ const jalaliFormats = ['D MMMM YYYY', 'LL'];
 
 moment().locale('fa');
 moment.updateLocale('fa', {
-  postformat(string) {
+  // eslint-disable-next-line object-shorthand
+  postforma: function(string) {
```
rhenshaw56: missing 't'

rhenshaw56 (changes requested):

packages/utilities/psammead-locales/moment/fa.js
RayNjeri: thanks

dr3 (commented):
LGTM, but is it possible to just make eslint global lint `/moment` to es5? would be easier to catch stuff in future

jamesdonoh (approved):
Good catch, thanks.

jamesdonoh:
@dr3 Incoming issue to improve linting which we'll do immediately after

RayNjeri:
@jamesdonoh i can just add the fix that @dr3 suggested here instead of creating an issue

dr3 (approved):
LGTM untill we add the es5 linting :)

rhenshaw56 (approved):

jamesdonoh:
Linting improvements: https://github.com/bbc/psammead/issues/2025

PR #2021, opened 2019-09-06
Add global styles to `psammead-styles`

### LINKED ISSUE #330: Add global styles to `psammead-styles`
We currently do not have an export of our global styles within the `psammead-styles` package. We should add these. 

We define them in Simorgh here https://github.com/bbc/simorgh/blob/67decf15656f785ae20caf0cef7e9e5657920766/src/app/lib/globalStyles.js#L5-L16

Export global styles, ensuring we use `styled-normalize` and box sizing.  Our font-faces should not be included initially, since they will vary depending on the service. 

N/A

**Testing notes**
[Tester to complete]

Dev insight: Unit tests should be sufficient. To test it works locally in an application, pull in the utility locally from Psammead into the Simorgh application. Check this package works when used within Storybook's style definitions too. 

EinsteinNjoroge:
Resolves #330 

**Overall change:** _Add global styles to `psammead-styles` and export it without fonts_

**Code changes:**

- _Add global styles to `psammead-styles` and export it without fonts_
- _Add tests and snapshot_

---


EinsteinNjoroge:
I'm not sure about this implementation, kindly confirm I'm on the right track

dr3 (commented):
Hey :) I see the issue says "Our font-faces should not be included initially, since they will vary depending on the service" but I would expect `${({ fonts }) => fonts && createFontStyles(fonts)}` to stay in here, so we can pass the fonts in in simorgh?

They will vary on service, but we can call it like `<GlobalStyles fonts={arrayOfServiceFonts} />` :)

Otherwise, this PR looks great :)

packages/utilities/psammead-styles/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 2.1.5 | [PR#2021](https://github.com/bbc/psammead/pull/2021) Export global styles |
```
sareh: Can you please update the package Readme with a description of how to use the `global-styles` import. A usage example would be great. One including fonts so that people can use them.

dr3: This should be a `minor` change as it adds functionality :)

packages/utilities/psammead-styles/src/__snapshots__/global-styles.test.jsx.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dr3, line 1: This file doesn't contain `Sans-Serif` which is passed in `it('should handle global-styles with fonts props', () => {`?

dr3 (changes requested):
Looks great, just a couple things :)

packages/utilities/psammead-styles/src/global-styles.js
```diff
@@ -0,0 +1,34 @@
...
+    *, *:before, *:after {
+      box-sizing: inherit;
+    }`;
+
+const GlobalStyles = createGlobalStyle`
+    ${styledNormalize}
+    
+    ${boxSizing}
+
+    ${({ fonts }) => fonts && createFontStyles(fonts)}
```
sareh: This creation of fonts in the `createFontStyles` method is very specific to our Simorgh application. It'd be better to decouple our implementation in Simorgh from this package, but retaining the ability to pass in fonts. 
```suggestion
    ${({ fonts }) => fonts ? fonts : null;}
```

sareh (changes requested):
Looks good so far! I've suggested some changes here.

packages/utilities/psammead-styles/README.md
```diff
@@ -84,6 +86,17 @@ const someGridUsingComponent = css`
...
 
+### global-styles
+
+We export a global styles component which uses `styled-normalize` and defines css rules for `box-sizing`. 
+This component accepts an optional `fonts` prop which is an array of font styles to be applied in the global styles.
+
+```js
+import GlobalStyles from '@bbc/psammead-styles/global-styles';
+
+<GlobalStyles fonts={['font-family: "Sans-Serif"']}/>
```
sareh: Thanks for the update. 

It'd be helpful to have example usage that uses our Psammead Styles fonts like this:

```
import GlobalStyles from '@bbc/psammead-styles/global-styles';
import {
  F_REITH_SANS_REGULAR,
  F_REITH_SERIF_MEDIUM,
} from '@bbc/psammead-styles/fonts';

<GlobalStyles fonts={[
      F_REITH_SANS_REGULAR,
      F_REITH_SERIF_MEDIUM,
    ]}
/>
```

Same for usage in the unit test & reference to the readme. 

A note for others code reviewing this PR: A Storybook example would be difficult to isolate, so shouldn't be necessary for this work. After this package is updated we'll use it in the global Psammead Storybook setup. So it'll be in use throughout the Storybook application.

sareh (changes requested):
Thanks for the changes. Suggested an update to the example.

packages/utilities/psammead-styles/src/global-styles.test.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+describe('global-styles', () => {
+  it('should inject normalize to global styles', () => {
+    render(<GlobalStyles />);
+    const styles = getGlobalStyles();
+    expect(styles).toMatchSnapshot();
+  });
+
+  it('should handle global-styles with fonts props', () => {
+    render(<GlobalStyles fonts={['font-family: "Sans-Serif"']} />);
```
dr3: I think this meant to add font faces, not font families. Just a test but good to be clear :)

sareh (approved):
👍 Looks great.

sareh:
@EinsteinCarrey After merging this in, we have two follow-up issues to use this package:  https://github.com/bbc/psammead/issues/2112 https://github.com/bbc/simorgh/issues/3718

dr3 (approved):
LGTM, great work :)

PR #2080, opened 2019-09-09
Configure eslint config overide for /moment to cater for IE11 compatibility issues

### LINKED ISSUE #2025: Eslint lint /moment to es5
Following up on #2013 #2016 && #2020  we've had to do a lot of manual fixes that disable es6 rules for /moment

Make Eslint global lint https://github.com/bbc/psammead/tree/latest/packages/utilities/psammead-locales/moment back es5 in order to catch more errors pertaining to IE11

**Testing notes**
1. Storybook on IE11 evaluates and renders correctly 

**Additional context**

RayNjeri:

Resolves #2025 

**Overall change:** _Introduces an eslint config for the moment folder in order to lint back to ES5._

**Code changes:**

- _Adds an .eslintrc config file._

---


packages/utilities/psammead-locales/moment/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  rules: {
+    // IE11 gives syntax error for trailing commas on functions and Moment
+    // locales are not transpiled
+    'comma-dangle': ['error', {
+      'functions': 'never',
+      'imports': 'only-multiline',
+      'objects': 'only-multiline',
+      'arrays': 'only-multiline',
+    }],
```
jamesdonoh: Do we need L5-10? I put this originally in https://github.com/bbc/psammead/pull/2016 but removed it as the suggestion of using the `es5` prettier option seemed to work better.

packages/utilities/psammead-locales/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-locales",
-  "version": "2.2.5",
+  "version": "2.2.6",
```
jamesdonoh: I'd make this a minor update as it's not really a bug fix.

jamesdonoh:
I tried adding a template literal to `fa.js` and this wasn't detected by the linter. Could you take another look at the rules and make sure that only IE11-compatible syntax is permitted?

jamesdonoh (dismissed):

packages/utilities/psammead-locales/moment/.eslintrc.js
RayNjeri: Yes, we still need this rule, otherwise prettier will try adding the commas and IE11 will fail

packages/utilities/psammead-locales/package.json
RayNjeri: My same thought, but Jenkins insisted on having this so idk

jamesdonoh: Where did it insist?

packages/utilities/psammead-locales/moment/.eslintrc.js
```diff
@@ -0,0 +1,27 @@
+module.exports = {
+  extends: [
+    "eslint:recommended",
+    "plugin:es5/no-es2015"
```
jamesdonoh, line 4: The documentation describes this as "Forbid ES2015 usage". That doesn't seem right?

packages/utilities/psammead-locales/moment/.eslintrc.js
jamesdonoh: I'm confused, when I worked on this before https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-locales/moment/.prettierrc#L3 was sufficient to stop prettier from adding dangling commas.

packages/utilities/psammead-locales/moment/.eslintrc.js
```diff
@@ -0,0 +1,27 @@
...
+    'no-var': 0,
+    'one-var':0,
+    'no-block-scoping': 0,
+    'no-template-literals': 0,
+    'no-nested-ternary':0,
+  },
+};
+
+
+
```
jamesdonoh: Extra blank lines.

packages/utilities/psammead-locales/moment/ar.js
```diff
@@ -1,12 +1,11 @@
-/* eslint-disable func-names */
-const moment = require('moment');
+var moment = require('moment');
```
jamesdonoh, line 1: As discussed on Slack, is there a way we can selectively allow `const` as IE11 doesn't seem to mind this?

packages/utilities/psammead-locales/moment/.eslintrc.js
RayNjeri: ES2015 Is ES6 so this is actually right

packages/utilities/psammead-locales/moment/ar.js
RayNjeri: i don't think there  is way we can achieve this whilst still using ES5 as this will bring about IIFE block scoping isssues

packages/utilities/psammead-locales/package.json
RayNjeri: here. https://ci.news.tools.bbc.co.uk/blue/organizations/jenkins/psammead/detail/IE11-fixes2/24/pipeline

Had to revert back the version to what it was, but jenkins fails

package.json
```diff
@@ -83,6 +83,7 @@
     "eslint": "^6.3.0",
     "eslint-config-airbnb": "^18.0.1",
     "eslint-config-prettier": "^6.2.0",
+    "eslint-plugin-es5": "^1.4.1",
```
dr3, line 86: This technically requires a base level psammead bump and changelog change 🙈

packages/utilities/psammead-locales/moment/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  extends: [
+    "eslint:recommended",
+    "plugin:es5/no-es2015"
+  ],
+  plugins: ['es5'],
+  rules: {
+    'object-shorthand': 0,
```
dr3, line 9: Howcome these specific rules? is `plugin:es5/no-es2015` not enough?

packages/utilities/psammead-locales/moment/index.stories.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
dr3: Can this disable be more specific? Maybe disable the es5 only stuff?

dr3 (changes requested):
LGTM, just a couple tweaks/questions :)

packages/utilities/psammead-locales/moment/.eslintrc.js
RayNjeri: Unfortunately yes

packages/utilities/psammead-locales/moment/index.stories.jsx
RayNjeri: had to disable eslint here for the entire file .i think its better if we maintained the ES6 features in here, otherwise it will force us into writing this file into ES5

thekp:
@RayNjeri 
I think we can add an override to your eslintrc in psammead-locales, to ignore the files we still want to use ES6 in. Then you can maybe remove `/* eslint-disable */`

```
  overrides: 
    [
      "excludedFiles": "*.jsx",
    ]
```

edit: this suggestion doesn't actually work. and I'm currently looking into a better example
result from findings: you can create a `. eslintignore` file in the same directory as your new eslintrc and then specify the files you want exclude. 

i.e : 
```
*.test.js
*.stories.jsx
```

EDIT2: I just tested it locally, and it doesn't work for me 👁 👁 

final edit: it works in the base `.eslintignore`

RayNjeri:
@thekp thanks for the suggestion, looking more into it

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 1.7.12 | [PR#2080](https://github.com/bbc/psammead/pull/2080) Bump @bbc/psammead-locales in use of es5 |
```
dr3:
```suggestion
| 1.7.12 | [PR#2080](https://github.com/bbc/psammead/pull/2080) Add 'eslint-plugin-es5' for use in @bbc/psammead-locales |
```

dr3 (approved):
LGTM :) great work

thekp (approved):
looks good to me 👍

PR #2100, opened 2019-09-11
Talos - Bump @bbc/psammead-timestamp-container, @bbc/psammead-locales

bbc-news-frameworks:
@bbc/psammead-timestamp-container
@bbc/psammead-locales

@bbc/psammead


sareh:
Closed since there is a newer PR: https://github.com/bbc/psammead/pull/2101/files

PR #2104, opened 2019-09-12
add function to detect and render helmet for use in snapshots

jroebu14:
**Overall change:** Adds functionality to `shouldMatchSnapshot` to include html and head tags if `react-helmet` was used inside of component. This enables us to see things like meta tags and scripts in snapshots.

**Code changes:**

- Adds new function called `renderWithHelmet` that is a wrapper around the `@testing-library/react` `render` method.

An example using the Chartbeat component in Simorgh

```jsx
const CanonicalChartbeatBeacon = ({ chartbeatConfig, chartbeatSource }) => {
  useEffect(() => {
    return () => {
      if (typeof window !== 'undefined' && window.pSUPERFLY) {
        /*
          This function is always called to update config values on page changes
          https://chartbeat.zendesk.com/hc/en-us/articles/210271287-Handling-virtual-page-changes
        */
        window.pSUPERFLY.virtualPage(chartbeatConfig);
      }
    };
  }, [chartbeatConfig]);

  return (
    <>
      <Helmet>
        <script async type="text/javascript">
          {`
        (function(){
          var _sf_async_config = window._sf_async_config = (window._sf_async_config || {});
          var config = ${JSON.stringify(chartbeatConfig)};
          for (var key in config) {
            _sf_async_config[key] = config[key];
          }
        })();
      `}
        </script>
        <script async type="text/javascript" src={chartbeatSource} />
      </Helmet>
      <h1>Hello I should be in the body</h1>
    </>
  );
};
```

would then render

```
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`CanonicalChartbeatAnalytics should return the helmet wrapper with the script snippet 1`] = `
<html>
  <head>
    <script
      async="true"
      type="text/javascript"
    >
      
        (function(){
          var _sf_async_config = window._sf_async_config = (window._sf_async_config || {});
          var config = {"domain":"test-domain","type":"article","sections":"section1 section2","chartbeatUID":1111,"virtualReferrer":null,"useCanonical":true,"title":"This is an article","uid":123};
          for (var key in config) {
            _sf_async_config[key] = config[key];
          }
        })();
      
    </script>
    <script
      async="true"
      src="//chartbeat.js"
      type="text/javascript"
    />
  </head>
  <body>
    <div>
      <h1>
        Hello I should be in the body
      </h1>
    </div>
  </body>
</html>
`;

```

---


packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -1,12 +1,60 @@
-import { render } from '@testing-library/react';
+import { render, waitForDomChange } from '@testing-library/react';
 import 'jest-styled-components';
 import deepClone from 'ramda/src/clone';
 
+const renderWithHelmet = async component => {
```
dr3: I know this file doesnt have tests, but as this is alot more complicated can you add some please?
Calling `shouldMatchSnapshot` a few times with various types of dummy components would be good. One with helmet, one with a fragment etc? We can use the snap it generates to assess its behavior?

dr3 (changes requested):
LGTM, just  some simple snapshot tests would be nice IMO as its more complicated

chris-hinds (approved):

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -1,12 +1,63 @@
...
+  /*
+ * Similar to this problem https://github.com/testing-library/react-testing-library/issues/402
+ * This helper was created to solve the problem of rendering helmet/head contents in snapshots.
+ * Pass in a component that uses helmet somewhere in the component tree.
+ * The full html tree is rendered and returned which you can then use to snapshot helmet/head contents.
+ * Example of use in a test file:
```
ryanmccombe: This example is confusing - `renderHelmet` isn't exported

Isn't this `renderWithHelmet` function an internal helper for `shouldMatchSnapshot`?  It doesn't seem like it should be used like described here, or that it would work if it was.

ryanmccombe (approved):

packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -1,12 +1,63 @@
...
+      }
+
+      return { container };
+    });
+};
+
 export const shouldMatchSnapshot = (title, component) => {
-  it(title, () => {
+  it(title, async () => {
     // select the first child to remove the pointless wrapping div from snapshots
```
dr3, line 8: This happens now :(

packages/utilities/psammead-test-helpers/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,72 @@
...
+    </title>
+    <meta
+      name="test name"
+      value="test value"
+    />
+    <script
+      src="test.js"
+    />
+  </head>
+  <body>
```
dr3, line 68: Any easy way to remove the body when not needed?

dr3 (changes requested):

packages/utilities/psammead-test-helpers/__snapshots__/index.test.jsx.snap
jroebu14: I'll look into it. Was worried about adding more complexity to an already complex looking function for not much value but it should be straightforward

packages/utilities/psammead-test-helpers/src/index.js
jroebu14: Oh yeh. Code has changed and I forgot to update the comments. A good example of how commenting code can be a bad idea!

packages/utilities/psammead-test-helpers/src/index.js
jroebu14: Are you worried about the async `it` callback? It won't affect the current test helper API.

packages/utilities/psammead-test-helpers/__snapshots__/index.test.jsx.snap
jroebu14: I tried removing the body but that stopped other tests from rendering things in the body. I guess it breaks the renderer because it can't find the body element's div to put stuff in. I also tried this with removing just the div and it had the same effect :(

packages/utilities/psammead-test-helpers/src/index.js
jroebu14: Is the comment looking ok now?

pjlee11:
Jenkins is failing because this updates `packages/utilities/moment-timezone-include/package-lock.json` but doesn't bump the package version and changelog

packages/utilities/psammead-test-helpers/src/index.js
dr3: Soz i meant `select the first child to remove the pointless wrapping div from snapshots` happens now. We no longer `remove the pointless wrapping div`

jroebu14: We still do when rendering components without helmet. I'll change the comments and var names.

dr3 (approved):
Soz for my confusion, looks great :)

PR #2113, opened 2019-09-12
Talos - Bump @bbc/psammead-story-promo

bbc-news-frameworks:
@bbc/psammead-story-promo

@bbc/psammead


thekp:
out of date

PR #2117, opened 2019-09-12
Optimise Navigation direction prop

DenisHdz:
Resolves #NaN

**Overall change:** 
Refactor Navigation to pass the `dir` prop to the main wrapper to avoid adding it to every link.

**Code changes:**
- Target the `StyledListItem` from the `StyledNav` and modify the CSS depending on the direction.
- Pass `dir` prop to the `Navigation` in the story.
- Update snapshots

---

### DIFF
--- a/packages/components/psammead-navigation/src/index.jsx
+++ b/packages/components/psammead-navigation/src/index.jsx
@@ -29,10 +29,6 @@ const SKIP_LINK_BORDER = '0.1875rem'; // 3px
 const SKIP_LINK_TOP_POSITION_LARGE = '-4rem'; // -64px
 const SKIP_LINK_TOP_POSITION_SMALL = '-3rem'; // -48px
 
-const StyledNav = styled.nav`
-  background-color: ${C_POSTBOX};
-`;
-
 const NavWrapper = styled.div`
   position: relative;
   max-width: ${GEL_GROUP_5_SCREEN_WIDTH_MIN};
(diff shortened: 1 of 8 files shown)


packages/components/psammead-navigation/src/index.stories.jsx
```diff
@@ -122,6 +122,7 @@ const navigationStory = (skipLinkText, currentPageText, navData, brand) =>
           script={script}
           skipLinkText={skipLinkText}
           service={service}
+          dir={dir}
```
sareh, line 134: Can you please add an example of a RTL service in the stories? It'll make it easier to spot issues to do with text direction within Psammead. You'll also want to add the `dir` attribute to `navStoriesData`

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -197,8 +183,27 @@ export const NavigationLi = ({
   </StyledListItem>
 );
 
-const Navigation = ({ children, script, skipLinkText, service }) => (
-  <StyledNav role="navigation">
+const StyledNav = styled.nav`
+  background-color: ${C_POSTBOX};
+
+  ${StyledListItem} {
```
ghost, line 189: This could be simplified to 
```jsx
${({ dir }) => css`
            &::after {
              ${dir === 'ltr' ? 'left' : 'right'}: 0;
            }`
```

DenisHdz: Done 🚀

ghost (approved):

packages/components/psammead-navigation/src/index.stories.jsx
DenisHdz: We can check that by selecting the service from the `knob`, but happy to add an `Arabic` story.

ryanmccombe (approved):

dr3 (approved):

paruchurisilpa:
Looks good to me..

Issue #1991, opened 2019-09-04
Jalaali calendar not shown when time is present


jamesdonoh:
**Describe the bug**
We added support for Jalaali calendar for Persian and Pashto in https://github.com/bbc/psammead/issues/1732. However this currently only works correctly for timestamps that only show the date without the time. When an absolute time is present the Jalaali calendar date does not display.

**To Reproduce**
Steps to reproduce the behaviour:

1. Look at Persian or Pashto page with time shown in timestamp
2. Observe Jalaali calendar not shown
3. Compare with a page where time is not shown, Jalaali calendar is displayed

**Expected behaviour**
The Jalaali calendar should always be shown for timestamps for the services that use it, whether the time is shown or not.

**Screenshots**

**Testing notes**
- [ ] We will need to add test coverage in this test [file](https://github.com/bbc/simorgh/blob/247d0f5be2916f0d145f18dc8e1c93fcc8cec61a/src/app/containers/ArticleTimestamp/helpers.test.js)
- [ ] This bug fix is expected to need manual testing.

**Additional context**
- [link to where the timestamp formatting happens](https://github.com/bbc/simorgh/blob/247d0f5be2916f0d145f18dc8e1c93fcc8cec61a/src/app/containers/ArticleTimestamp/helpers.js#L18-L28)
- [link to jalali formats](https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-locales/moment/fa.js#L23)
- Sareh has informed us that we can see the different timestamp scenarios through [storybook](https://bbc.github.io/simorgh/?path=/story/containers-article-article-timestamp--lastpublished-today-more-than-10-hours-ago-and-firstpublished-today-more-than-10-hours-ago), so  we can get a better understanding of how it works

PR #2142, opened 2019-09-16
Add format for jalaali calendar with time

HarveyPeachey:
POC of #1991 

**Overall change:** Allows the Jalali calendar to render within timestamps that include a time

**What it should do**

- If an article is `firstPublished` and `lastPublished` within a certain timeframe then a time within the timestamp will show along with the Jalali calendar, examples of these can be found in [storybook](https://bbc.github.io/simorgh/?path=/story/containers-article-article-timestamp--lastpublished-today-more-than-10-hours-ago-and-firstpublished-today-more-than-10-hours-ago)
- When the article timestamp has a time it should translate it to the jalali format with the time on the end

**Reasons why**
- Timezone 'BST' is being used is because all services using the Jalali calendar will follow the BST timezone as the timezones vary for people who use this calendar, so It was decided at product level to choose BST.
---


HarveyPeachey:
This work is going to be split between other PR's and a new psammead utility is going to be made to address this original issue.

PR #2143, opened 2019-09-16
Talos - Bump @bbc/psammead-locales

bbc-news-frameworks:
@bbc/psammead-locales

@bbc/psammead


thekp:
out dated by this: https://github.com/bbc/psammead/pull/2148/commits

PR #2190, opened 2019-09-19
Add moment calendar conversion utility

### LINKED ISSUE #1991: Jalaali calendar not shown when time is present

### LINKED ISSUE #2239: Implement Jalaali calendar formatting as a new utility package
To help solve #1991 and from discussions within the team, we should migrate the code for Jalaali calendar formatting from `psammead-locales` that is currently being done in a helper as a `postformat` of a moment object. I'm creating this as a separate issue for readability sake. This is going to be used for Article timestamps but can be used by other components.

- Create a utility that allows a moment to be passed into it, and formats it in the Jalaali calendar format.
- Should be designed and implemented in a way which would allow more calendar conversions in the future
- Make use of `jalaali-js` (a third party library) which handles the conversion of a gregorian date into its Jalaali equivalent
- Allow the output to be returned to the calling function as a single string.
- Stories should be created to show this functionality.

**Testing notes**
Unit tests will need to be created for this.

HarveyPeachey:
Resolves #2239 
Part of implementation for #1991 

**Overall change:** This PR adds a new moment calendar utility, the reason for creating this utility is because our Persian and Pashto services need support for the Jalaali calendar, as well as the Gregorian calendar for Article Timestamps. 

Currently, this logic is held within the `psammead-locales` utility. The logic to decide whether Jalaali should be added is done as a `postformat` on a string that has been generated by moment. It was decided that this check should be done at a higher level, as existing code had to check that the string that moment had generated had to be checked again which is counter-intuitive.

This utility is being created in a general sense, so that in the future if we need to support other calendars then we can make that process straight forward.

**Code changes:**

- Added a new utitility package called `psammead-calendars`
- Wrote code in a general sense so that the utility can be imported and used easily as shown from this [review](https://github.com/bbc/psammead/pull/2190#discussion_r327713642)
- Copied part of the logic for Jalaali conversion from `psammead-locales` into this utility
- Added a variety of unit tests
- Added new story on storybook to showcase the calendar conversion

**StoryBook Screenshots**
Screenshot 2019-10-01 at 11 10 52
Screenshot 2019-10-01 at 11 11 03

---


packages/utilities/psammead-calendars/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+This package provides a utility to convert calendar dates provided in a `Moment` object to other non-gregorian calendars. For now Jalaali is the only calendar which is being converted, more may be added in the future.
```
jamesdonoh:
```suggestion
This package provides a utility to format calendar dates provided in a `Moment`(https://momentjs.com/) object to other non-Gregorian calendars that Moment does not support. For now Jalaali is the only calendar which is supported, more may be added in the future.
```

packages/utilities/psammead-calendars/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```jsx
+npm install @bbc/psammead-calendars --save
+```
+
+## Calendars
+
+### Jalaali
+
+The functionality of this helper is to to convert a `Gregorian` calendar moment into it's `Jalaali` equivelant and outputting it as a string to be used by other components. This helper relies on `jalaali-js` which contains the logic to carry out the conversions from `Gregorian` to `Jalaali`. Documentation of this dependency can be found [here](https://www.npmjs.com/package/jalaali-js).
```
jamesdonoh:
```suggestion
The functionality of this helper is to to convert a `Gregorian` calendar moment into its `Jalaali` equivalent and outputting it as a string to be used by other components. This helper relies on `jalaali-js` which contains the logic to carry out the conversions from `Gregorian` to `Jalaali`. Documentation of this dependency can be found [here](https://www.npmjs.com/package/jalaali-js).
```

packages/utilities/psammead-calendars/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "bugs": {
+    "url": "https://github.com/bbc/psammead/issues"
+  },
+  "homepage": "https://github.com/BBC-News/psammead/blob/latest/packages/components/psammead-moment-calendars/README.md",
+  "dependencies": {
+    "jalaali-js": "1.1.0",
+    "moment": "^2.24.0"
+  },
+  "peerDependencies": {
+    "react": "^16.9.0"
```
jamesdonoh, line 28: I would expect this could be a `devDependency` as it's not needed at runtime, only when running tests.

packages/utilities/psammead-calendars/src/calendars/jalaali.js
```diff
@@ -0,0 +1,46 @@
...
+    'تله',
+    'لړم',
+    'لیندۍ',
+    'مرغومی',
+    'سلواغه',
+    'کب',
+  ],
+};
+
+function getJalaaliCalendar(gregorianMoment) {
```
jamesdonoh: I was thinking maybe something like this as an API to make the calling code readable and allow other calendars to be added later.

```javascript
import moment from 'moment';
import { jalaali } from '@bbc/psammead-calendars';

const gregorianMoment = moment(someTimestamp);
const gregorianStr = gregorianMoment.locale(locale).format('D MMMM YYYY');
const jalaaliStr = jalaali.formatDate(gregorianMoment);
```

packages/utilities/psammead-calendars/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Calendars Changelog
```
jamesdonoh:
```suggestion
# Psammead Calendars Changelog
```

packages/utilities/psammead-calendars/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
...
+
+| Version | Description |
+|---------|-------------|
+| 1.0.0-alpha.0 | [PR#2190](https://github.com/bbc/psammead/pull/2190) Initial creation of package. |
```
jamesdonoh:
```suggestion
| 1.0.0 | [PR#2190](https://github.com/bbc/psammead/pull/2190) Initial creation of package. |
```

jamesdonoh (changes requested):

packages/utilities/psammead-calendars/src/calendars/jalaali.js
dr3: SGTM :) them in `psammead-timestamp-container` we can do
```js
import moment from 'moment';

const gregorianMoment = moment(someTimestamp);
const gregorianStr = gregorianMoment.locale(locale).format('D MMMM YYYY');
const secondaryStr = secondaryCalender && secondaryCalender.formatDate(gregorianMoment);
```

and pass `secondaryCalender` from the service config to aid unwanted weight on other services :)

packages/utilities/psammead-calendars/src/calendars/jalaali.js
```diff
@@ -0,0 +1,56 @@
...
+    'لیندۍ',
+    'مرغومی',
+    'سلواغه',
+    'کب',
+  ],
+};
+
+const jalaali = {
+  formatDate: gregorianMoment => {
+    if (typeof gregorianMoment.locale !== 'function') {
```
simonsinclair: If `gregorianMoment` is null or undefined, does this line throw a TypeError? It might be safer to check `if (!gregorianMoment && ...` before trying to access `locale`.

packages/utilities/psammead-calendars/src/calendars/jalaali.test.js
```diff
@@ -0,0 +1,87 @@
...
+      locale: 'en',
+      expected: null,
+      summary: 'should return null if object passed is not a moment',
+    },
+  ];
+  testScenarios.forEach(({ nonMoment, date, locale, expected, summary }) => {
+    it(summary, () => {
+      let testMoment = moment(date);
+      testMoment.locale(locale);
+      if (nonMoment) {
```
simonsinclair: If you extract the logic on L69 and L80-82  into separate test, you might be able to make this one more pure.

The second test could look something like:

```javascript
it('should fail when a Moment instance is not passed', () => {
  // ...
  const jalaaliCalendar = jalaali.formatDate({ foo: 'bar' });
  // ...
});
```

simonsinclair (changes requested):
This is great, @HarveyPeachey. 👍 I've added a few thoughts.

packages/utilities/psammead-calendars/src/calendars/jalaali.test.js
HarveyPeachey: Yep makes sense, I'll create another set of tests that will check the passed values type

packages/utilities/psammead-calendars/src/calendars/jalaali.js
HarveyPeachey: Yeah you're right it doesn't throw an error, I'll amend

packages/utilities/psammead-calendars/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "email": "PsammeadMaintainers@bbc.co.uk"
+  },
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/bbc/psammead/issues"
+  },
+  "homepage": "https://github.com/BBC-News/psammead/blob/latest/packages/components/psammead-calendars/README.md",
+  "dependencies": {
+    "jalaali-js": "1.1.0",
+    "moment": "^2.24.0"
```
dr3, line 25: Moment can be a peerDependancy :)

packages/utilities/psammead-calendars/src/index.js
```diff
@@ -0,0 +1,3 @@
+import jalaali from './calendars/jalaali';
+
+export default jalaali;
```
dr3: Probably best to do 
`export jalaali from './calendars/jalaali';` 
This will mean doing 
```
import { jalaali } from '@bbc/psammead-calendars`
```
instead of
```
import jalaali from '@bbc/psammead-calendars`
```
But will avoid breaking changes if we add a second calender type

packages/utilities/psammead-calendars/package.json
HarveyPeachey: What's the difference between `peerDependancy` and a normal `dependancy`?

dr3: >Having a peer dependency means that your package needs a dependency that is the same exact dependency as the person installing your package.
https://yarnpkg.com/lang/en/docs/dependency-types/#toc-peerdependencies

(ignore thats Yarn, not npm, same meaning 🙈)

So its staying, "to use `psammead-calender` you need to install moment, and we use the same version"

packages/utilities/psammead-calendars/src/calendars/jalaali.test.js
```diff
@@ -0,0 +1,115 @@
...
+      summary: "should return null as date doesn't exist",
+    },
+    {
+      date: '31-2019-12',
+      locale: 'ps',
+      expected: null,
+      summary: 'should return null as this is an invalid moment',
+    },
+  ];
+  const nonMomentTestScenarios = [
```
simonsinclair: It's great to see you've chosen to separate these scenarios.

As they don't relate to Moment, would it be better to use a name that doesn't have "moment" in it and rename `dateTestScenarios` to `momentTestScenarios`? It might make things more readable.

HarveyPeachey: What do you suggest calling it?

simonsinclair: Would `momentTestScenarios` and `formatDateTestScenarios` make sense?

packages/utilities/psammead-calendars/src/index.js
jamesdonoh: I agree this is more flexible, will we need a comment to silence `prefer-default-export` as well?

packages/utilities/psammead-calendars/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+The functionality of this helper is to to convert a `Gregorian` calendar moment into its `Jalaali` equivalant and outputting it as a string to be used by other components. This helper relies on `jalaali-js` which contains the logic to carry out the conversions from `Gregorian` to `Jalaali`. Documentation of this dependency can be found [here](https://www.npmjs.com/package/jalaali-js).
+
+## Usage
+
+<!-- Description of the component usage -->
+
+```
+import Calendars from "@bbc/psammead-calendars"
+```
```
jamesdonoh, line 25: Please can you add a proper usage example here?

packages/utilities/psammead-calendars/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```jsx
+npm install @bbc/psammead-calendars --save
+```
+
+## Calendars
+
+### Jalaali
+
+The functionality of this helper is to to convert a `Gregorian` calendar moment into its `Jalaali` equivalant and outputting it as a string to be used by other components. This helper relies on `jalaali-js` which contains the logic to carry out the conversions from `Gregorian` to `Jalaali`. Documentation of this dependency can be found [here](https://www.npmjs.com/package/jalaali-js).
```
jamesdonoh:
```suggestion
The functionality of this helper is to to convert a `Gregorian` calendar moment into its `Jalaali` equivalent and outputting it as a string to be used by other components. This helper relies on `jalaali-js` which contains the logic to carry out the conversions from `Gregorian` to `Jalaali`. Documentation of this dependency can be found [here](https://www.npmjs.com/package/jalaali-js).
```

packages/utilities/psammead-calendars/src/calendars/jalaali.test.js
HarveyPeachey: They both use `formatDate()` so I'm not sure that's the best wording

packages/utilities/psammead-calendars/src/calendars/jalaali.js
```diff
@@ -0,0 +1,61 @@
...
+    const jalaaliDate = jalaaliJs.toJalaali(
+      gregorianMoment.year(),
+      gregorianMoment.month() + 1,
+      gregorianMoment.date(),
+    );
+    const localeJalaaliMonths = jalaaliMonths[gregorianMoment.locale()];
+    const jalaaliMonth = localeJalaaliMonths[jalaaliDate.jm - 1];
+    const output = `${jalaaliDate.jd} ${jalaaliMonth} ${jalaaliDate.jy}`;
+    return output;
+  },
```
jamesdonoh, line 61: This function is getting quite long, what about splitting out the validation into a separate internal `isSupportedMoment` helper function and leaving the rest of this function just carrying out the actual conversion/formatting? This would help make the code easier to follow.

jamesdonoh (changes requested):

jamesdonoh:
Sorry another thing, I'm getting this warning when running the tests locally, is there anything we can do about it?
```
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
```

HarveyPeachey:
> Sorry another thing, I'm getting this warning when running the tests locally, is there anything we can do about it?
> 
> ```
> Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
> ```

I'm not sure. It shows up because one of the test cases uses an invalid date format to test that scenario.

packages/utilities/psammead-calendars/src/calendars/jalaali.js
HarveyPeachey: Yeah that makes sense. Something like this? 

```
// Helper function to check if argument passed through is a supported moment
const isSupportedMoment = moment => {
  if (
    moment &&
    typeof moment.locale === 'function' &&
    moment.isValid() &&
    `${moment.locale()}` in jalaaliMonths
  ) {
    return true;
  }
  return false;
};
```
Which would make the formatting look like this:
```
const jalaali = {
  formatDate: gregorianMoment => {
    if (isSupportedMoment(gregorianMoment)) {
      const jalaaliDate = jalaaliJs.toJalaali(
        gregorianMoment.year(),
        gregorianMoment.month() + 1,
        gregorianMoment.date(),
      );
      const localeJalaaliMonths = jalaaliMonths[gregorianMoment.locale()];
      const jalaaliMonth = localeJalaaliMonths[jalaaliDate.jm - 1];
      const output = `${jalaaliDate.jd} ${jalaaliMonth} ${jalaaliDate.jy}`;
      return output;
    }
    return null;
  },
};
```

packages/utilities/psammead-calendars/src/index.js
HarveyPeachey: Looks like it passed the es-linting with @dr3's suggestion fine.

dr3: Haha I expected eslint to complain 🙈 odd 😛

HarveyPeachey: > Haha I expected eslint to complain 🙈 odd 😛

Maybe eslint is clever enough to detect the `export default` on the [jalaali.js](https://github.com/bbc/psammead/blob/a5e2214f6b4a318556fa879772a32a586a563812/packages/utilities/psammead-calendars/src/calendars/jalaali.js#L64) file?

dr3: `export thing from 'thing'` isnt actually a real thing (yet), we use `@babel/plugin-proposal-export-default-from` to get the syntax, so I have a feeling eslint doesnt know what it means and just nopes out 🙈

packages/utilities/psammead-calendars/src/calendars/jalaali.test.js
simonsinclair: That's probably a good argument _against_ separating the scenarios. Maybe this would be a better route to keeping the test pure? https://gist.github.com/simonsinclair/6f37753e74050339edfa63cf0ef408db

oluoluoxenfree:
I can't run storybook on this branch; do I need to do something different to `npm ci:packages`?

HarveyPeachey:
> I can't run storybook on this branch; do I need to do something different to `npm ci:packages`?

Try `npm run install:packages` @oluoluoxenfree

oluoluoxenfree (approved):
LGTM :)

packages/utilities/psammead-calendars/package-lock.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "@bbc/psammead-calendars",
+  "version": "1.0.0",
+  "lockfileVersion": 1
+}
```
simonsinclair, line 62: `psammead-calendars` depends on `jalaali-js`, but it doesn't appear in the lockfile. To update it:

```
cd packages⁩/utilities⁩/psammead-calendars⁩
npm install
```

simonsinclair (changes requested):
The tests are looking 👌.
See my comment on the lockfile.
Are we doing anything about the deprecation warning?

simonsinclair (approved):

HarveyPeachey:
> The tests are looking 👌.
> See my comment on the lockfile.
> Are we doing anything about the deprecation warning?

I've asked James hopefully he'll respond at the end of today, if not I can just remove the single test case that's causing the warning.

jamesdonoh:
@HarveyPeachey have you tried the workaround for the warning mentioned in the docs? https://momentjs.com/guides/#/warnings/js-date/

HarveyPeachey:
> @HarveyPeachey have you tried the workaround for the warning mentioned in the docs? https://momentjs.com/guides/#/warnings/js-date/

@jamesdonoh Yeah I saw that. I think I'm just going to remove the single test case. What it's testing is a format which isn't in ISO 8601 format, which doesn't really test what I've done, it's testing if the moment logic is working.

packages/utilities/psammead-calendars/README.md
jamesdonoh: Sorry I should have been clearer - can you include an example of how to actually use the function that is returned, and what is returned?

Also could you document the behaviour when an unsupported locale is specified?

jamesdonoh (changes requested):
Just needs a tiny bit more documentation. Also please can you add a screenshot of the new stories to the description for convenience?

jamesdonoh (approved):
This looks good - thanks for changes 👍

dr3 (approved):

FK78 (approved):

PR #2213, opened 2019-09-23
Grid

sareh:
Resolves #3684

**Overall change:** A new Psammead component `psammead-grid`. It has a single export `Grid` that has the flexibility to be a wrapper or a grid item. 

**Code changes:**

- wrapper prop to toggle on the `display: grid`
- enableGelMargins, enableGelGutters, enableGelMaxWidths are all boolean props Here are the GEL guidelines https://www.bbc.co.uk/gel/guidelines/grid#grid-sizes The flexibility is due to the fact that at the page level grid, we will want all of these values, but for nested components we might not need all of them. It allows us to choose the implementation. 
- columns prop takes an object of groups as keys and values being: the number of columns. 
- startOffset prop - allows you to start at an offsetted number of columns. These follow CSS Grid line numbering conventions, so use `2` to start at column 2. *Note*: the startOffset value only applies if you have `columns` defined for that group. 
- create a separate toggle for `enableGroupFourMaxWidth` (Currently only used by World Services Front Pages, but it is useful to have it at this level instead of forcing the team to extend the `<Grid>` component themselves)
- Fallbacks for 

Additional considerations (outside the scope of this PR)
- Fallbacks using % widths without gutters
- Using Feature query `import { grid } from '@bbc/psammead-styles/detection';` - seems to break the implementation. Do not know why. 

---


packages/components/psammead-grid/src/index.test.jsx
```diff
@@ -0,0 +1,93 @@
...
+describe('Grid component', () => {
+  shouldMatchSnapshot(
+    'should render Grid wrapper with Grid',
+    <Grid
+      columns={{ group3: 6, group4: 8, group5: 20 }}
+      startOffset={{ group3: 1, group4: 1, group5: 5 }}
+      enableGelMargins
+      enableGelGutters
+      enableGelMaxWidths
+      wrapper
```
dr3: Stylistic - IMO we should default to `wrapper` and use `leaf` on the opposite, to denote when a grid item is a leaf node and doesnt require a grid inside :)

packages/components/psammead-grid/src/index.test.jsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import { shouldMatchSnapshot } from '@bbc/psammead-test-helpers';
+import { Grid } from '.';
+import { ExampleParagraph, ExampleFigure } from './helpers';
+
+describe('Grid component', () => {
+  shouldMatchSnapshot(
+    'should render Grid wrapper with Grid',
+    <Grid
+      columns={{ group3: 6, group4: 8, group5: 20 }}
```
dr3, line 10: Stylistic - This may be because its an example, but imo we should have a file somewhere, idm psammead or simorgh that has something like like
```js
export const articlePageGridItem = { group3: 6, group4: 8, group5: 20 };
export const largeGridItem = { group3: 6, group4: 6, group5: 12 };
export const mediumGridItem = { group2: 6, group3: 5, group4: 5, group5: 10 };
```

sareh: Great suggestion. I'd suggest these not be within this Psammead component, but be in Simorgh, since they're very specific to the page layout of Simorgh articles/ front pages / etc.

packages/components/psammead-grid/src/index.test.jsx
sareh: I do like the naming of the `leaf` prop. However, since there are fewer wrapper nodes than leaf nodes, I think this current setup will be easier for users of the Grid component.  Happy to consider changing the name of `wrapper`, if you have another more clearer suggestion.

dr3: Thats fair, SGTM :)

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,235 @@
...
+  GEL_GROUP_2_SCREEN_WIDTH_MAX,
+  GEL_GROUP_3_SCREEN_WIDTH_MIN,
+  GEL_GROUP_3_SCREEN_WIDTH_MAX,
+  GEL_GROUP_4_SCREEN_WIDTH_MIN,
+  GEL_GROUP_4_SCREEN_WIDTH_MAX,
+  GEL_GROUP_5_SCREEN_WIDTH_MIN,
+} from '@bbc/gel-foundations/breakpoints';
+// import { grid } from '@bbc/psammead-styles/detection';
+
+const groups = {
```
DenisHdz, line 23: I was thinking.. could we add in this object, also the group respective `gutter` and `margin` instead of defining separately `gelGridGutters` and `gelMargins`? And then return the values inside the `gridMediaQueries()` function?

Otherwise, for cases like this one: 
`<Grid wrapper enableGelGutters enableGelMargins columns={{ group4: 8 }}>`
we will be adding the `grid-column-gap` and `padding` for other breakpoints, rather than just for the range defined by the `group4`.

sareh: Yes, that's a good point. We should still be able to toggle them on/off, just the implementation behind the scenes should be simpler, as you suggested. I'll see if I can make the change prior to the review session!

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,183 @@
...
+`;
+
+const GridComponent = styled.div`
+  ${gridFallbacks}
+
+  @supports (display: grid) {
+    ${({ enableGelGutters }) => enableGelGutters && gelGridGutters}
+    ${({ enableGelMargins }) => enableGelMargins && gelMargins}
+    ${({ enableGelMaxWidths }) => enableGelMaxWidths && gelMaxWidths}
+    ${({ enableGroupFourMaxWidth }) =>
```
dr3: I think `enableGroupFourMaxWidth` shouldnt be an option here, but instead set in a element that expands `Grid`

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,183 @@
...
+    max-width: ${group5WrapperMaxWidth};
+  }
+`;
+const group4MaxWidth = css`
+  @media (min-width: ${GEL_GROUP_4_SCREEN_WIDTH_MIN}) {
+    max-width: ${group4WrapperMaxWidth};
+  }
+`;
+
+const gridMediaQueries = ({ columns, startOffset }) => {
```
dr3: It almost feels like unit testing each of these functions on their own might be beneficial

packages/components/psammead-grid/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Psammead Grid Changelog
+
+<!-- prettier-ignore -->
+| Version | Description |
+|---------|-------------|
+| 1.0.0-alpha.0 | [PR#x](https://github.com/bbc/psammead/pull/x) Initial package creation - Grid and GridItem exports. |
```
jroebu14: PR number needs to be provided

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+       <p>Paragraph that spans 2 out of 8 columns through group4</p>
+    </Grid>
+    <Grid columns={{ group4: 2 }}>
+       <p>Paragraph that spans 2 out of 8 columns through group4</p>
+    </Grid>
+  </Grid>
+);
+```
+
+####<a name="multiple-groups">Setting values for multiple groups</a>
```
jroebu14: headings need formatted properly

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+####<a name="gutters-margins">Setting standard GEL gutters and GEL Margins</a>
+
+Using `enableGelGutters` and `enableGelMargins` on the wrapper Grid element.
+
+```jsx
+import Grid from '@bbc/psammead-grid';
+
+const MyComponent = () => (
+	<Grid
+		wrapper
```
jroebu14: formatting issues

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+      <p>Paragraph - for group 3 spans 6/6 columns, for group 4 spans 6/8 columns</p>
+    </Grid>
+    <Grid columns={{ group3: 6, group4: 2 }}>
+      <p>Paragraph - for group 3 spans 6/6 columns, for group 4 spans 2/8 columns</p>
+    </Grid>
+  </Grid>
+);
+```
+
+[Screenshot of this example without GEL Gutters and GEL Margins](./documentation/Screenshot\ example\ without\ gelMargins\ and\ gelGutters.png)
```
jroebu14: Needs fixed

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+    <Grid columns={{ group4: 2 }}>
+      <p>Paragraph that spans 2 out of 8 columns through group4</p>
+    </Grid>
+    <Grid columns={{ group4: 2 }}>
+       <p>Paragraph that spans 2 out of 8 columns through group4</p>
+    </Grid>
+    <Grid columns={{ group4: 2 }}>
+       <p>Paragraph that spans 2 out of 8 columns through group4</p>
+    </Grid>
+  </Grid>
```
jamesdonoh, line 77: As mentioned on the call, think it would be beneficial to have a simple schematic diagram to explain how this would 'look' and what the effect of the different values for `group4` has.

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+		columns={{ group3: 6, group4: 8 }}
+	>
+    <Grid columns={{ group3: 6, group4: 6 }}>
+      <p>Paragraph - for group 3 spans 6/6 columns, for group 4 spans 6/8 columns</p>
+    </Grid>
+    <Grid columns={{ group3: 6, group4: 2 }}>
+      <p>Paragraph - for group 3 spans 6/6 columns, for group 4 spans 2/8 columns</p>
+    </Grid>
+  </Grid>
+);
```
jamesdonoh, line 128: What happens if one group is omitted? Does it by default assume the full width? Would be good if we could document the default behaviour.

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+  - It can and should be used whenever you're defining a new grid. [See nested grid example.](#nested-grid). It should be used in any case where you're defining a new nested grid.
+- When should I use the `columns` prop?
+  - This should always be defined. For a Grid element that is not a `wrapper`, it's the number of columns it should span.
+- When should I use the `startOffset` prop?
+  - `startOffset` is an object just like the `columns` prop. You can set it on a Grid child element to start it at a column other than the first one. 
+  - If you don't pass it in, the value defaults to 1, the start of the grid.
+- Why is there no vertical spacing on the grid?
+    - The Grid implementation only has gutters/margins for columns, [according to the GEL specification](https://www.bbc.co.uk/gel/guidelines/grid#grid-sizes).  This is to allow flexibility for a variety of spacing. To add vertical spacing, you should add padding/margin/top/bottom to the contents.
+
+<!-- ### When not to use this component -->
```
jamesdonoh, line 180: Like, never? 😄

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,181 @@
...
+  ${({ enableGroupFourMaxWidth }) => enableGroupFourMaxWidth && group4MaxWidth}
+  ${gridFallbacks}
+  @supports (display: grid) {
+    ${gridMediaQueries}
+    ${({ wrapper }) =>
+      wrapper ? `display: grid; position: initial;` : `display: block;`}
+  }
+`;
+
+const Grid = ({ children, ...otherProps }) => {
```
dr3: Need logic to ensure that a child cant have more columns than its parent

packages/components/psammead-grid/package-lock.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "1.0.0-alpha.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@bbc/gel-foundations": {
+      "version": "3.4.0",
+      "resolved": "https://registry.npmjs.org/@bbc/gel-foundations/-/gel-foundations-3.4.0.tgz",
+      "integrity": "sha512-WuewciWzvSb/WKhUGSvIoighqCdXGN2PsVQbZbUo5/eWi5+o3O8uoojUpeQeRZMYWLsLVELPHr6qJDuqKghXKA=="
+    },
+    "@bbc/psammead-paragraph": {
```
jroebu14: should be dev dependency

packages/components/psammead-grid/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,699 @@
...
+    <p
+      class="c2"
+    >
+      3
+      This is a long paragraph that will wrap for several lines. This is a long paragraph that will wrap for several lines. This is a long paragraph that will wrap for several lines. This is a long paragraph that will wrap for several lines. This is a long paragraph that will wrap for several lines. This is a long paragraph that will wrap for several lines. This is a long paragraph that will wrap for several lines. This is a long paragraph that will wrap for several lines.
+    </p>
+  </div>
+  <div
+    class="c7"
+    startoffset="[object Object]"
```
jroebu14: possible reserved attribute name

packages/components/psammead-grid/src/helpers.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
jroebu14: rename file to `testHelpers`

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,181 @@
...
+  return selectedGroups.map(group =>
+    mediaQuery({
+      min: groups[group].min,
+      max: groups[group].max,
+      styles: `
+        width: initial;
+        grid-template-columns: repeat(${columns[group]}, 1fr);
+        grid-column-end: span ${columns[group]};
+      ${enableGelGutters ? `grid-column-gap: ${groups[group].gutterSize};` : ``}
+      ${enableGelMargins ? `padding: 0 ${groups[group].marginSize};` : ``}
```
jroebu14, line 103: && would be cleaner

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,181 @@
...
+      return `position: relative;`;
+    }
+    const selectedGroups = Object.keys(columns);
+    return selectedGroups.map(
+      group =>
+        `
+      ${mediaQuery({
+        min: groups[group].min,
+        max: groups[group].max,
+        styles: `
```
jroebu14, line 130: comments to explain please

packages/components/psammead-grid/CHANGELOG.md
jroebu14: mentions ` Grid and GridItem exports.` however only exports `Grid`

jroebu14:
Looks good to me. The thing I mentioned at the end of the meeting was considering having `Grid` for when it's a wrapper component and then `GridItem` as this might be a clearer API however if there's a good reason for just `Grid` then I'm all for it.

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,181 @@
...
+      return React.cloneElement(child, {
+        parentColumns: otherProps.columns,
+        parentStartOffset: otherProps.startOffset,
+      });
+    });
+
+  return <GridComponent {...otherProps}>{renderChildren()}</GridComponent>;
+};
+Grid.propTypes = {
+  children: node.isRequired,
```
dr3, line 175: Need a columns proptype with shape

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+## Installation
+
+`npm install @bbc/psammead-grid`
+
+## Props
+
+<!-- prettier-ignore -->
+| Argument  | Type | Required | Default | Example |
+| --------- | ---- | -------- | ------- | ------- |
+| columns | object | yes | N/A | { group1: 6, group3: 8 } |
```
dr3: Example should have all 6 groups

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+| startOffset | object | no | Sets all values as 1 for each of the groups defined in `columns`  | { group1: 2, group3: 3 } |
+| wrapper | boolean | no | false | `wrapper` |
```
dr3: Descriptions of what each prop does would be nice :)

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+You can enable it with the prop `enableGelMaxWidths`. 
+
+For some pages, we will want to constrain the page to 1008px. You can use `enableGroupFourMaxWidth` for this. 
+
+### When to use this component
+
+This `Grid` component can be used for page-level layouts as well as for layouts of individual components. You can nest Grids within one another for more complex layouts. Please see the Storybook Stories for examples.
+
+### Frequenty Asked Questions
+
```
AlistairGempf, line 166: I don't know if this is the best place for this question, but I think it'd be good to clarify what happens when you have grid items with a total number of columns that's greater than the wrapper they're in, eg:

```jsx
<Grid wrapper columns={{ group2: 4 }}>
  <Grid columns={{ group2: 3 }}>
    <Child />
  </Grid>
  <Grid columns={{ group2: 2 }}>
    <Child />
  </Grid>
</Grid>
```
I'm guessing the second non-wrapper Grid will move onto a second row and leave the 4th column of the first row empty. Is that correct?

packages/components/psammead-grid/src/index.jsx
AlistairGempf: This logic will presumably need to take into account the `startOffset` prop of the child (or what it gets renamed to).

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,176 @@
...
+- When should I use the `wrapper` prop?
+  - The `wrapper` prop should be used on a parent Grid element. It defines the grid, using the `columns` prop's values to define how many columns to define at each breakpoint. 
+  - It can and should be used whenever you're defining a new grid. [See nested grid example.](#nested-grid). It should be used in any case where you're defining a new nested grid.
```
AlistairGempf:
```suggestion
  - It can and should be used whenever you're defining a new grid. [See nested grid example](#nested-grid). It should be used in any case where you're defining a new nested grid.
```

packages/components/psammead-grid/package-lock.json
sareh: Has been removed as a dependency in the package.json since it is a devDependency at the repo level (can be used in tests and in Storybook).

packages/components/psammead-grid/src/helpers.jsx
sareh: Done

packages/components/psammead-grid/src/index.jsx
sareh: I tried this. However, since this is within a template literal, I get the undefined value returned in the CSS. So I'm sticking with the ternary here.  
Screenshot of CSS in browser with false false grid-column-start: undefined

packages/components/psammead-grid/src/index.jsx
sareh: Done

packages/components/psammead-grid/CHANGELOG.md
sareh: Done

packages/components/psammead-grid/README.md
sareh: Done

packages/components/psammead-grid/README.md
sareh: Fixed.

packages/components/psammead-grid/README.md
sareh: The image has had its syntax updated, but since I've used a relative path, it's pointing to the file location on the `latest` branch, so isn't visible on the PR just yet. Haven't yet found a work-around for this.

packages/components/psammead-grid/src/index.jsx
sareh: This is handled automatically by the current setup - a child cannot span more grid columns than the number of columns defined in its parent.

packages/components/psammead-grid/README.md
sareh: Yes that is correct. I've created this diagram to show what happens in that case.
The first item spans the three columns and the second item spans two on the next row down
four column grid with items three and two

packages/components/psammead-grid/src/index.jsx
sareh: Done

packages/components/psammead-grid/README.md
sareh: This has been updated

packages/components/psammead-grid/README.md
sareh: @dr3 I've added. description below the props table, since adding it in the table was difficult to read. Please let me know if you think the text should be updated.

packages/components/psammead-grid/README.md
sareh: Have added an image below.

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,198 @@
...
+    <Grid item columns={{ group4: 2 }}>
+       <p>Item 3 - Paragraph that spans 2 out of 8 columns through group4</p>
+    </Grid>
+    <Grid item columns={{ group4: 2 }}>
+       <p>Item 4 - Paragraph that spans 2 out of 8 columns through group4</p>
+    </Grid>
+  </Grid>
+);
+```
+Image of this example
```
sareh:
```suggestion
Image of this example when viewed at within `group4` (from 1008px to 1279px). When you look at this example at any other breakpoint, each of the four items fill up the full width of the row. 
```

packages/components/psammead-grid/src/index.test.jsx
sareh: @dr3 I've updated implementation to follow your suggested layout. This is due to the feedback that it makes more sense to have a grid and a grid item. So I've called the wrapper `<Grid>` and then the grid item is `<Grid item>` - so requires a boolean prop called 'item' to be added to it.

packages/components/psammead-grid/README.md
sareh: Updated the text entirely.

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,198 @@
...
+| startOffset | object | no | Sets all values as 1 for each of the groups defined in `columns` | { group1: 1, group2: 1, group3: 1, group4: 2, group5: 5 } |
+| item | boolean | no | false | `item` |
+| enableGelGutters | boolean | no | false | `enableGelGutter` |
+| enableGelMargins | boolean | no | false | `enableGelMargins` |
+| enableGelMaxWidths | boolean | no | false | `enableGelMaxWidths` |
+| enableGroupFourMaxWidth | boolean | no | false | `enableGroupFourMaxWidth` |
+
+- When should I use the `columns` prop?
+  - This should always be defined.
+  - For a `<Grid>` element, it's the number of columns the grid has.
```
dr3:
```suggestion
  - For a `<Grid>` element, it's the number of columns the grid has, and the number of columns it should span.
```

packages/components/psammead-grid/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "cssgrid",
+    "styled-components"
+  ],
+  "dependencies": {
+    "@bbc/gel-foundations": "^3.4.0",
+    "@bbc/psammead-styles": "^2.3.0"
+  },
+  "peerDependencies": {
+    "styled-components": "^4.3.2"
+  }
```
DenisHdz, line 34: We need to add the `alpha` tag.
```
  "publishConfig": {
    "tag": "alpha"
  }
```

jroebu14: From the readme:

> Currently this component is in alpha. This is because it has not yet been tested with various assistive technologies. After it has had an accessibility swarm, this will be published under a standard version.

packages/components/psammead-grid/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,714 @@
...
+@media (min-width:37.5rem) {
+  .c2 {
+    font-size: 1rem;
+    line-height: 1.375rem;
+  }
+}
+
+<div
+  class="c0"
+  startoffset="[object Object]"
```
dr3: Need to fix/rename this :)

packages/components/psammead-grid/README.md
jroebu14: @jamesdonoh 

Does this explain?

> Image of this example when viewed at within group4 (from 1008px to 1279px). When you look at this example at any other breakpoint, each of the four items fill up the full width of the row.

from line 78 of the readme.

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+        /* [1] */
+      })}
+    `,
+    );
+  }}
+`;
+
+const GridComponent = styled.div`
+  ${({ enableGelMaxWidths }) => enableGelMaxWidths && gelMaxWidths}
+  ${({ enableGroupFourMaxWidth }) => enableGroupFourMaxWidth && group4MaxWidth}
```
dr3: Ive lost the thread i said this in, but IMO `enableGroupFourMaxWidth` should be extracted into a wrapper component

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+      });
+    });
+
+  return <GridComponent {...otherProps}>{renderChildren()}</GridComponent>;
+};
+
+Grid.propTypes = {
+  children: node.isRequired,
+  columns: shape({
+    group1: number,
```
dr3, line 177: IMO for now, all these groups should be required

packages/components/psammead-grid/src/index.test.jsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import { shouldMatchSnapshot } from '@bbc/psammead-test-helpers';
+import Grid from '.';
+import { ExampleParagraph, ExampleFigure } from './testHelpers';
+
+describe('Grid component', () => {
```
dr3, line 6: All of these examples, IMO should have all 6 groups in their props

dr3 (changes requested):

packages/components/psammead-grid/src/__snapshots__/index.test.jsx.snap
jroebu14: I've renamed startOffset prop inside of the Grid component so that we can keep the prop name but prevent it from rendering an element attribute https://github.com/bbc/psammead/pull/2213/commits/cae3de92abda620ee4fd72d5d19aa5e36e371318

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+`npm install @bbc/psammead-grid`
+
+## Props
+
+<!-- prettier-ignore -->
+| Argument | Type | Required | Default | Example |
+| -------- | ---- | -------- | ------- | ------- |
+| columns | object | yes | N/A | { group1: 6, group2: 6, group3: 6, group4: 8, group5: 20 } |
+| startOffset | object | no | Sets all values as 1 for each of the groups defined in `columns` | { group1: 1, group2: 1, group3: 1, group4: 2, group5: 5 } |
```
AlistairGempf, line 18: I think this has been changed to `gridStartOffset`
```suggestion
| gridStartOffset | object | no | Sets all values as 1 for each of the groups defined in `columns` | { group1: 1, group2: 1, group3: 1, group4: 2, group5: 5 } |
```

jroebu14: @AlistairGempf the prop itself hasn't been renamed, it's just been given an alias. I've updated the comment to hopefully make that clearer https://github.com/bbc/psammead/pull/2213/commits/f6d9c05074fdf2d23b61debd5e0e7cc9363d757e

packages/components/psammead-grid/package.json
DenisHdz: But for now, it should be published with the alpha tag.

jroebu14: ah ok sorry. I misunderstood. I'll add that in

DenisHdz:
Getting these two errors about `parentColumns` and `parentStartOffset`:

grid-errors

packages/components/psammead-grid/src/index.jsx
DenisHdz: It does work for me using &&:
```
  ${enableGelGutters && `grid-column-gap: ${groups[group].gutterSize};`}
  ${enableGelMargins && `padding: 0 ${groups[group].marginSize};`}
```

jroebu14:
@DenisHdz I think this because we were cloning any children inside of Grid to pass those props down in case the children were nested Grid components. The error happened when we passed those props to e.g. paragraph elements. I've written a check to ensure we only pass those props to Grid components https://github.com/bbc/psammead/pull/2213/commits/299a62085e446785dfd0ba01f3b38fbd99e16ca2

I also don't see `parentStartOffset` being used anywhere so I'm going to remove that prop

packages/components/psammead-grid/src/index.jsx
jroebu14: @dr3 I'm struggling to understand this. Can you provide a quick code example?

packages/components/psammead-grid/src/index.jsx
jroebu14: It seems the grid component handles not being passed groups:

> In this example groups 1, 2, 3, & 5 are omitted from the columns prop - so at those breakpoints items fill the whole width of the page.

packages/components/psammead-grid/src/index.test.jsx
jroebu14: I *think* Sareh purposefully designed the grid component to handle not passing in groups. If you want items to span the full width of the page you can either not pass in those groups or you can configure the groups to span the full width. The latter would render a lot of CSS to do what the former does by default.

packages/components/psammead-grid/src/index.jsx
DenisHdz: I believe that he means to constrain the page max-width in a wrapper rather than in the Grid itself:

```
const Wrapper = styled.div`
   ${({ enableGroupFourMaxWidth }) => enableGroupFourMaxWidth && group4MaxWidth} 
`;

<Wrapper enableGroupFourMaxWidth>
   <Grid columns={{ group3: 6, group4: 8 }}>
        <Grid item columns={{ group3: 6, group4: 6 }}>
           ...
       </Grid>
   </Grid>
</Wrapper>
```

or

```
const WrapperFourMaxWidth = styled.div`
  margin: 0 auto;
  @media (min-width: ${GEL_GROUP_4_SCREEN_WIDTH_MIN}) {
    max-width: ${GROUP_4_MAX_WIDTH};
  }
`

<WrapperFourMaxWidth>
   <Grid columns={{ group3: 6, group4: 8 }}>
        <Grid item columns={{ group3: 6, group4: 6 }}>
           ...
       </Grid>
   </Grid>
</WrapperFourMaxWidth>

```

jroebu14: Ok thanks. So a separate exported component to constrain width and center the grid. What's the benefit of having this as a separate component? Also, if we do that for `enableGroupFourMaxWidth` then why wouldn't we apply the same thinking to `enableGelMaxWidths`?

dr3: I was more suggesting something like
```
const ArticlePageGridStyle = styled(Grid)`
  margin: 0 auto;
  @media (min-width: ${GEL_GROUP_4_SCREEN_WIDTH_MIN}) and (max-width: ${GEL_GROUP_4_SCREEN_WIDTH_MAX}) {
    max-width: ${GROUP_4_MAX_WIDTH};
  }
  @media (min-width: ${GEL_GROUP_5_SCREEN_WIDTH_MIN}) {
    max-width: ${GROUP_5_MAX_WIDTH};
  }
`;

export const ArticlePageGrid = ({ children, ...props }) => (
  <ArticlePageGridStyle
    columns={{ group5: 20 }}
    enableGelGutters
    enableGelMargins
    {...props}
  >
    {children}
  </ArticlePageGridStyle>
);
```

but likely not in this PR, we need to decide where to put them

DenisHdz: I don't really know if there would be a benefit on that, I guess it is more for clarity and cleanliness. If in the future we want to have another page type width, then we would have to update the `GridComponent` to allow to pass a new constraint, rather than create another export and use that wrapper.

I agree that if we change it, it should be applied to both `enableGroupFourMaxWidth` and `enableGelMaxWidths`, but I would like to hear @dr3  thoughts first.

dr3: @jroebu14 @DenisHdz I was more suggesting something like
```
const ArticlePageGridStyle = styled(Grid)`
  margin: 0 auto;
  @media (min-width: ${GEL_GROUP_4_SCREEN_WIDTH_MIN}) and (max-width: ${GEL_GROUP_4_SCREEN_WIDTH_MAX}) {
    max-width: ${GROUP_4_MAX_WIDTH};
  }
  @media (min-width: ${GEL_GROUP_5_SCREEN_WIDTH_MIN}) {
    max-width: ${GROUP_5_MAX_WIDTH};
  }
`;

export const ArticlePageGrid = ({ children, ...props }) => (
  <ArticlePageGridStyle
    columns={{ group5: 20 }}
    enableGelGutters
    enableGelMargins
    {...props}
  >
    {children}
  </ArticlePageGridStyle>
);
```

but likely not in this PR, we need to decide where to put them

jroebu14: @dr3 ok cool. I can create a separate issue to cover this work if that's ok with you

packages/components/psammead-grid/README.md
```diff
@@ -0,0 +1,208 @@
...
+    This spans 5 out of 6 columns
+  </Grid>
+</Grid>
+```
+
+- When should I use the `startOffset` prop?
+  - `startOffset` is an object structured just like the `columns` prop. You can set it on a `<Grid item>` element to start it at a column other than the first one.
+  - If you don't pass it in, the offset value defaults to 1, the start of the grid.
+- Why is there no vertical spacing on the grid?
+  - The Grid implementation only has gutters/margins for columns, [according to the GEL specification](https://www.bbc.co.uk/gel/guidelines/grid#grid-sizes). This is to allow flexibility for a variety of spacing. To add vertical spacing, you should add padding/margin/top/bottom to the contents.
```
AlistairGempf: This links to grid sizes, and the gutters and margin layout is actually in a section above.
```suggestion
  - The Grid implementation only has gutters/margins for columns, [according to the GEL specification](https://www.bbc.co.uk/gel/guidelines/grid#spacing-and-layout). This is to allow flexibility for a variety of spacing. To add vertical spacing, you should add padding/margin/top/bottom to the contents.
```

packages/components/psammead-grid/src/index.jsx
jroebu14: @dr3 does this look ok? https://github.com/bbc/simorgh/issues/4054

packages/components/psammead-grid/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "psammead-grid",
+  "version": "1.0.0-alpha.0",
+  "publishConfig": {
+    "tag": "alpha"
+  },
+  "description": "Grid setup for components",
+  "main": "dist/index.js",
+  "module": "esm/index.js",
+  "sideEffects": false,
```
dr3:
```suggestion
```
This isnt needed, as the files cant be 'pruned', e.g. `groups` object

packages/components/psammead-grid/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,658 @@
...
+    display: inline-block;
+    width: NaN%;
+    vertical-align: top;
+  }
+}
+
+@media (min-width:37.5rem) and (max-width:62.9375rem) {
+  .c5 {
+    display: inline-block;
+    width: 83.33333333333333%;
```
dr3, line 108: Ideally these should be rounded, but not hugely imporant

jroebu14: to 83% or 83.3%?

dr3 (approved):

packages/components/psammead-grid/src/__snapshots__/index.test.jsx.snap
dr3: 1 or 2 decimal places would be fine :)

packages/components/psammead-grid/src/index.jsx
```diff
@@ -0,0 +1,202 @@
...
+    const selectedGroups = Object.keys(columns);
+    return selectedGroups.map(
+      group =>
+        `
+      ${mediaQuery({
+        min: groups[group].min,
+        max: groups[group].max,
+        styles: `
+          display: inline-block;
+          width: ${(100 * columns[group]) / parentColumns[group]}%;
```
sareh, line 132:
```suggestion
          width: ${parentColumns[group] ? (100 * columns[group]) / parentColumns[group] : 100}%;
```

dr3:
IMO we should enforce the use of all 6 groups, until we have logic for them to default. However as @sareh is on leave, @jroebu14 is gunna make an issue to discuss it after

jroebu14:
@dr3 @sareh one to consider for the next release https://github.com/bbc/psammead/issues/2282

DenisHdz (approved):
IMO, this is ready to go. We can iterate afterward. Great work from everyone :)

PR #2244, opened 2019-09-25
Open PR in Simorgh when Talos is run successfully

### LINKED ISSUE #2174: Allow Talos to raise a PR in Simorgh updating `@bbc/psammead*` pacakges
Now that we have talos successfully keeping `@bbc/psammead*` packages up-to-date we should enable talos to open a PR in Simorgh so it can update the newly published psammead packages.

Once talos has merged a PR updating psammead dependencies it should also raise a PR in simorgh to update to the newly published dependencies.

Thus creating 3 PR when making changes in psammead.

1) Initial psammead change
2) Auto talos PR updating bbc deps
3) (New) Auto talos pr updating `@BBC/psammead*` deps in Simorgh

Not doing this..............

**Testing notes**
[Tester to complete]

**Additional context**

pharingee:
Resolves #2174

**Overall change:** 
Remotely create a PR in Simorgh when Talos is run successfully to update published packages

**Code changes:**
- Remotely update package.json in Simorgh

---


.gitignore
```diff
@@ -18,3 +18,6 @@ packages/**/dist
 packages/**/esm
 
 published.txt
+
+# Editor settings
+.vscode
```
jroebu14, line 23: was this intended?

scripts/talos/bumpSimorghPackages/getPublishedPackages/index.js
```diff
@@ -0,0 +1,16 @@
+const getPublishedPackages = bumpedPackages => {
+  const publishedPackages = {};
+  Object.values(bumpedPackages).forEach(bumpedPackageList => {
+    bumpedPackageList.forEach(bumpedPackage => {
+      const sections = bumpedPackage.replace(/  +/g, ' ').split(' ');
+      if (sections.length === 4) {
```
jroebu14: does this mean the package was upgraded? If so then `isUpgraded` would be a good variable name to put here. It also means the `bumpedPackage` isn't an accurate name

.gitignore
pharingee: Yes it was. There are no vscode files in the codebase and I needed a couple of settings set up which might not be the same for everyone

scripts/talos/bumpSimorghPackages/getPublishedPackages/index.js
pharingee: Yeah, I think `publishedPackage` will be a better name

jroebu14: Not a big deal but you can easily extract the things you need with regexes

```js
packageList.forEach(package => {
  const isUpgraded = package.inclues('  →  ');

  if (isUpgraded) {
    const [packageName] = package.match(/^@bbc\/psammead-\S+/)
    const [version] = package.match(/(?<=→\s\s).+$/)
    publishedPackages[packageName] = version;
  }
});
```

scripts/talos/bumpSimorghPackages/getUpdates/index.js
```diff
@@ -0,0 +1,13 @@
+const getUpdates = (dependencies, publishedPackages) => {
+  const updates = [];
+  Object.keys(publishedPackages).forEach(packageName => {
+    const version = dependencies[packageName];
+    if (version)
+      updates.push(
```
jroebu14: No big deal but I try to discourage the use of mutation where possible. I think `map` and `filter(Boolean)` would be cleaner than mutating a variable outside the scope of a `forEach` callback

pharingee: Yeah, initially updates needed to be an object thus the mutation. I'll refactor to array functions now.

scripts/talos/bumpSimorghPackages/getPublishedPackages/index.js
pharingee: I tend to find regexes not very straightforward to read. So I avoid them when I can. In this case it seems to make the code more readable so I'll use them.

jroebu14: @pharingee that's a fair point!

scripts/talos/bumpSimorghPackages/updatePackageJson/index.js
```diff
@@ -0,0 +1,30 @@
+const commitRemoteChanges = require('../../../utilities/commitRemoteChanges');
+
+const mergeUpdates = (packageJson, publishedPackages) => {
+  const newPackageJson = { ...packageJson };
+  Object.keys(publishedPackages).forEach(key => {
+    if (newPackageJson.dependencies[key])
```
jroebu14: it would be nice to see this stored as a well named variable like `hasNewDependency`

scripts/talos/bumpSimorghPackages/updatePackageJson/index.js
```diff
@@ -0,0 +1,30 @@
+const commitRemoteChanges = require('../../../utilities/commitRemoteChanges');
+
+const mergeUpdates = (packageJson, publishedPackages) => {
+  const newPackageJson = { ...packageJson };
+  Object.keys(publishedPackages).forEach(key => {
```
jroebu14, line 14: I think the `mergeUpdates` abstraction should be the callback function in the forEach here. It's also possible to build up a single object using a reduce function instead of creating an object to mutate. An example of this could be:

```js
const mergeUpdates = publishedPackages => (packageJson, key) => {
  const { dependencies, devDependencies } = packageJson;
  const newDep = dependencies && dependencies[key];
  const newDevDep = devDependencies && devDependencies[key];

  return mergeAll([
    packageJson,
    newDep && {
      dependencies: {
        ...dependencies,
        [key]: publishedPackages[key],
      },
    },
    newDevDep && {
      devDependencies: {
        ...devDependencies,
        [key]: publishedPackages[key],
      },
    },
  ]);
};
```

and used in `updatePackageJson` like:

```js
  const newPackageJson = Object.keys(publishedPackages).reduce(
    mergeUpdates(publishedPackages),
    packageJson,
  );
```

more lines required to the same thing but possibly more readable and doesn't include mutations.

jroebu14 (commented):
I haven't tested this out yet but overall the code seems good to me. Nice work. I'll approve after I test it out manually after lunch!

jroebu14 (approved):

scripts/talos/bumpSimorghPackages/updatePackageJson/index.js
pharingee: I agree on the avoidance on mutation. But I think this sacrifices some readability to achieve that. I feel your code is more elegant but significantly less readable because of multiple complex uses of object destruction and the reduce function. By renaming some of the variables I think the updates I made should make it easier to read.

thekp (commented):
I think it would be worth to update the documentation here: https://github.com/bbc/psammead/blob/latest/scripts/talos/README.md
just to briefly note that Talos also opens a PR in Simorgh now

FK78 (approved):
LGTM!

thekp (approved):
Thanks for updating the readme.

Do you have an example of Simorgh Talos PR?

scripts/talos/bumpSimorghPackages/updatePackageJson/index.js
```diff
@@ -0,0 +1,32 @@
...
+    const hasNewDevDependency = newPackageJson.devDependencies[key];
+    if (hasNewDependency)
+      newPackageJson.dependencies[key] = publishedPackages[key];
+    if (hasNewDevDependency)
+      newPackageJson.devDependencies[key] = publishedPackages[key];
+  });
+  return newPackageJson;
+};
+
+const updatePackageJson = async (
```
dr3, line 36: Why arnt you using non-check-packages for this like psammead?

dr3 (changes requested):
This PR introduces a second way of updating dependencies in a package.json file. IMO we should be consistent and share the logic between updating psammead and simorgh. On my phone ATM, will be able to give more detailed feedback after 4pm BST

pharingee:
@dr3 I agree. I think we should update Talos' run on psammead to drop the use of `npm-check-updates` since it's resource intensive and takes a really long time to complete (In the order of minutes). A sample run of the new method can be seen here on a fork: https://github.com/pharingee/simorgh/pull/4

dr3:
> @dr3 I agree. I think we should update Talos' run on psammead to drop the use of `npm-check-updates` since it's resource intensive and takes a really long time to complete (In the order of minutes). A sample run of the new method can be seen here on a fork: [pharingee/simorgh#4](https://github.com/pharingee/simorgh/pull/4)

@pharingee The fact its slow or resource intensive isnt a concern to me, as talos isnt something whose duration affects developers. I would much rather we use the script, in favour of writing (and having to maintain) our own code.

pharingee:
I think the speed and resources used are important because of 3 reasons:
1. There has been at least one recorded instance where Talos has been aborted by a member of the team.
2. A request was made I think by @amywalker1987 to not fail the Jenkins Pipeline when Talos fails.
3. There are multiple instances of Talos failing at the `npm-check-updates` level.

In respect to maintaining our own code, the amount of code here to maintain is 2 files 51 lines long in total. If we wrote Talos - at close to 1000 lines of maintenance - to fix dependabot, I think the  benefit is worth the amount of code we need to maintain.

scripts/talos/bumpSimorghPackages/getUpdates/index.js
```diff
@@ -0,0 +1,20 @@
...
+      const newVersion = publishedPackages[packageName];
+      return (
+        semver.validRange(oldVersion) &&
+        semver.validRange(newVersion) &&
+        semver.intersects(oldVersion, newVersion)
+      );
+    })
+    .map(
+      packageName =>
+        `${packageName}  ${dependencies[packageName]}  →  ${publishedPackages[packageName]}`,
```
dr3, line 16: Is this being done to mimic the output of `npm-check-updates`? why?

scripts/talos/bumpSimorghPackages/index.js
```diff
@@ -0,0 +1,55 @@
...
+const getUpdates = require('./getUpdates');
+const updatePackageLock = require('./updatePackageLock');
+const updatePackageJson = require('./updatePackageJson');
+
+const path = 'package.json';
+const repoName = 'simorgh';
+
+const getSimorghPullRequestBody = updates => {
+  return updates
+    .map(update => `${update}\n\n${getChangelogHead(update)}`)
```
dr3, line 16: Why have we got 2 ways of creating pull request bodies? (psammead and simorgh prs) when they can be the same?

dr3 (commented):
This appears to share almost no commonalities with how we bump psammead packages. Having two entirely different implementations isn't sustainable IMO, 900 lines of code (inc tests to maintain) to do pretty much the same thing, in a new repo seems alot

dr3:
> I think the speed and resources used are important because of 3 reasons:
> 
> 1. There has been at least one recorded instance where Talos has been aborted by a member of the team.

They could equally do that if it was shorter, they shouldnt

> 2. A request was made I think by @amywalker1987 to not fail the Jenkins Pipeline when Talos fails.

Thats not nothing to do with the contents of the talos script, but the way its run.

> 3. There are multiple instances of Talos failing at the `npm-check-updates` level.

Theres also risk in having our own code, if we have issues with the script, our first reaction should be to debug and report bugs to the maintainers, not (half) run away from it

pharingee:
I have open a draft PR that replaces npm-check-updates and completes the check for all packages in 2 minutes here: https://github.com/bbc/psammead/pull/2409

scripts/talos/bumpSimorghPackages/index.js
pharingee: The two repositories are different in structure. Psammead contains multiple packages, while Simorgh is a single repository. So we need to keep the bodies separate.

scripts/talos/bumpSimorghPackages/getUpdates/index.js
pharingee: Yes it is. Because there is a function written to take advantage of that structure. If we effectively take `npm-check-updates out`, we can do without the mimicking and have a unified structure across.

pharingee: Talos was written to update Psammead. With no intention of updating another repository. The current features of Talos like running commit messages and pushes from the command line will mean that we will have to clone the entire Simorgh repository and isolate it from the current git repository. Which seems like a lot of work which is prone to errors and inconsistency.

pharingee: A lot of the new functions written can replace the current way Talos runs by using Github API to make commits and pushes. That will unify both methods.

pharingee:
> They could equally do that if it was shorter, they shouldnt
> 
They do it because it hangs. If it were to throw an error, we could at least debug it. But by hanging for several minutes it's impossible to know what went wrong.


> Theres also risk in having our own code, if we have issues with the script, our first reaction should be to debug and report bugs to the maintainers, not (half) run away from it

`npm-check-updates` was designed to handle published articles to the npm for a package and has been optimised for that. In the case of psammead, having multiple packages updated isn't an ideal use-case for npm-check-updates, which is why we need to couple it with lerna to make it work. This isn't running away but taking a different route which is optimised for our workflow.

pharingee:
Picking this up again

PR #2253, opened 2019-09-27
Create psammead-play-button for use in Media Player placeholder

### LINKED ISSUE #2248: Add a play button to Media Player placeholder
We need to add a play button to the placeholder images of the media player, so that users are aware it is playable.

1. We should create `psammead-play-button` following the UX specification.
2. Overlay it on Media Player placeholder.

image

Designs, including screenreader UX:
- Default state for play button, (with duration displayed): https://zpl.io/aMK7WBd
- Hover/focus state for play button (with duration displayed): https://zpl.io/VOKM56l

According to the UX spec, the visually hidden text should follow this format for video:
- `Play video, "Video Title", 1 minute 40 seconds`

And this format for audio:
- `Play audio, "Audio Title", 1 minute 40 seconds`

We can get this information from the media block that we're embedding.

We will need to add a story into Storybook with this logic.

Accessibility notes:
```
HTML:
Button

Announced text:
Play video, "Video Title", 1 minute 40 seconds (Duration not possible with current version of SMP) 

Landmark: 
Sits within Primary (primary content)

NB:
- Sit within a 'figure'
- Image alt text should not be read out
```

n/a

**Testing notes**
[Tester to complete]

**Additional context**
- The play button will be the same for audio and video for the MVP.
- We can look at iterating on this at a later stage to introduce a 'speaker' icon for audio.
- The play button would be used in a Placeholder component, which might also house things like Guidance warnings etc.

simonsinclair:
Partially resolves https://github.com/bbc/psammead/issues/2248

**Overall change:** This is a new Psammead component called `PlayButton`, which renders a button with a 'video' or 'audio' icon and optional duration timestamp. It has been designed to be used within a media player placeholder.

**Code changes:**

- Create `src/index.js` -  exports PlayButton.
- Create `src/index.test.js`.
- Create `src/__snapshots__/index.test.jsx.snap`.
- Create `src/index.stories.js`.
- Create [README.md](https://github.com/bbc/psammead/blob/2248-av-media-asset-play-button/packages/components/psammead-play-button/README.md)
- Create [CHANGELOG.md](https://github.com/bbc/psammead/blob/2248-av-media-asset-play-button/packages/components/psammead-play-button/CHANGELOG.md)
- Create `package.json` and `package-lock.json` file.

**Within Media Player:**

insitu

---


packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+const Button = styled.button`
+  background-color: ${C_EBON};
+  border: none;
+  color: ${C_WHITE};
+  cursor: pointer;
+  display: block;
+  ${({ service }) => getSansBold(service)}
+  ${GEL_MINION};
+  height: 80px;
```
dr3: We generally use rems, for 80px could you please use `5rem` :) it can be declared as a constant like `GEL_SPACING_DEC` (for 10 gel spacings) :)

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+  > svg {
+    fill: ${C_WHITE};
+    height: 24px;
+    ${({ datetime, duration }) =>
+      datetime &&
+      duration &&
+      css`
+        margin-top: ${GEL_SPACING};
+      `}
+    width: 24px;
```
dr3: This can use https://github.com/bbc/psammead/blob/latest/packages/utilities/gel-foundations/src/spacings.js#L8 :)

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+  &:focus {
+    background-color: ${C_POSTBOX};
+    transition: background-color 300ms;
+  }
+`;
+
+const IconWrapper = styled.div`
+  > svg {
+    fill: ${C_WHITE};
+    height: 24px;
```
dr3: This can use https://github.com/bbc/psammead/blob/latest/packages/utilities/gel-foundations/src/spacings.js#L8 :)

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+  ${GEL_MINION};
+  height: 80px;
+  padding: 0;
+  transition: background-color 300ms;
+  width: 80px;
+
+  &:hover,
+  &:focus {
+    background-color: ${C_POSTBOX};
+    transition: background-color 300ms;
```
dr3: Can `300` be defined as a constant at the top of this file please :)

dr3 (changes requested):
Great work :) Just a few consistency things :)

packages/components/psammead-play-button/src/index.jsx
simonsinclair: Thanks for the heads-up.

packages/components/psammead-play-button/src/index.jsx
simonsinclair: Done. 👍  I kept it a string, so that it needn't be coerced and can keep it's unit like the other constants, but let me know if it should be a number.

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+const GEL_SPACING_DEC = '5rem';
+const BGC_TRANSITION_DURATION = '300ms';
+
+const Button = styled.button`
+  background-color: ${C_EBON};
+  border: none;
+  color: ${C_WHITE};
+  cursor: pointer;
+  display: block;
```
dr3, line 19: I think this may need
```
possition: absolute;
bottom: 0;
```
To position it at the bottom of the player placeholder :)

simonsinclair: Yeah, I'm glad you brought that up. In the 'in situ' GIF 👆, I used the parent to position it. The button knows which service it's rendering in, so _could_ position itself (e.g. for RTL), but was worried that might limit its use. What do you think?

dr3: I think if you only specify `bottom` and not `left` or `right`, the browsers default `rtl`/`ltr` behaviour will do that for us :) 

image

If you select `persian` in storybook and add these styles, it works with `rtl` automatically :D

simonsinclair: That's awesome. I'll do that. Thanks, Drew!

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+  &:hover,
+  &:focus {
+    background-color: ${C_POSTBOX};
+    transition: background-color ${BGC_TRANSITION_DURATION};
+  }
+`;
+
+const IconWrapper = styled.div`
+  > svg {
+    fill: ${C_WHITE};
```
greenc05: Can we make sure this works in a custom high contrast mode on PC and when the user changes colour preferences on FF please?

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+const TimeDuration = styled.time`
+  display: block;
+  margin-top: ${GEL_SPACING};
+`;
+
+const PlayButton = ({ datetime, duration, type, service, title, onClick }) => (
+  <Button service={service} onClick={onClick}>
+    <VisuallyHiddenText>{`Play ${type}, "${title}"`}</VisuallyHiddenText>
+    <IconWrapper datetime={datetime} duration={duration} aria-hidden="true">
+      {mediaIcons[type]}
```
greenc05: Can we check the inline SVG has an attribute of focusable false?

greenc05: Actually I see that in the test above....

packages/components/psammead-play-button/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,431 @@
...
+  width: 1.5rem;
+}
+
+<button
+  class="c0"
+>
+  <span
+    class="c1"
+  >
+    Play audio, "Dog barks at cat."
```
greenc05, line 63: Just checking.... we wouldn't actually have quotes around the tittle would we (unless the journalist has put them in)?

greenc05:
Lets make sure a developer accessibility swarm is carried out on this component before the alpha status is removed please

packages/components/psammead-play-button/src/index.jsx
simonsinclair: Hey @dr3, I've pushed a few changes that relinquish some style control to the parent following our discussion earlier.

packages/components/psammead-play-button/src/index.jsx
simonsinclair: Check. ✅

Screen Shot 2019-10-02 at 20 52 59

packages/components/psammead-play-button/src/__snapshots__/index.test.jsx.snap
simonsinclair: According to the UX a11y notes in Zeplin, the title should be wrapped in quotes. Is this an error or because we're using commas as delimiters and the title might include one?

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+  width: ${GEL_SPACING_DEC};
+
+  &:hover,
+  &:focus {
+    background-color: ${C_POSTBOX};
+    transition: background-color ${BGC_TRANSITION_DURATION};
+  }
+`;
+
+const IconWrapper = styled.div`
```
dr3, line 34: Did try to see if we can style the svg directly, avoiding the need for an extra div, but Im unable to do it with styled-components. This LGTM

dr3 (commented):
LGTM, however I think this will need some a11y tweaks to work on windows high contrast and firefox with colours set. Dont mind when this is done (as component is alpha) Happy to approve if this is gunna be fixed in a follow up PR :)

Great work :D

simonsinclair:
@greenc05, I'm organising a developer a11y swarm today.
Thanks @dr3, following the swarm, I'll submit any further tweaks to this PR.

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+}) => {
+  const hiddenText =
+    datetime && duration && durationSpoken
+      ? `Play ${type}, "${title}", ${durationSpoken}`
+      : `Play ${type}, "${title}"`;
+  return (
+    <Button className={className} service={service} onClick={onClick}>
+      <VisuallyHiddenText>{hiddenText}</VisuallyHiddenText>
+      <IconWrapper
+        datetime={datetime}
```
HarveyPeachey, line 73: Not sure it matters but shouldn't the attribute be camelCased to `dateTime = {..}`?

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+  datetime,
+  duration,
+  durationSpoken,
+  type,
+  service,
+  title,
+  onClick,
+}) => {
+  const hiddenText =
+    datetime && duration && durationSpoken
```
HarveyPeachey, line 66: Maybe change `datetime` to `dateTime` just to be consistent with the rest of the camelCasing?

packages/components/psammead-play-button/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Psammead Play Button Changelog
+
+<!-- prettier-ignore -->
+| Version | Description |
+| ------------- | ----------- |
+| 1.0.0-alpha.0 | [PR#2248](https://github.com/bbc/psammead/issues/2248) Create initial package. |
```
HarveyPeachey: I think this needs to link to the PR not the issue
```suggestion
| 1.0.0-alpha.0 | [PR#2253](https://github.com/bbc/psammead/pull/2253) Create initial package. |
```

HarveyPeachey (commented):
Looks really good just a few small things

packages/components/psammead-play-button/src/index.jsx
simonsinclair: It should - good spot! I wonder why the linter didn't pick that up.

packages/components/psammead-play-button/src/index.jsx
simonsinclair: I'm not sure JSX enforcing camelCase on attributes (that will end up being lowercase) is a good reason to change this, because `datetime` is actually the machine-readable value of "datetime", not "date time" or "date-time".

packages/components/psammead-play-button/CHANGELOG.md
simonsinclair: Another good spot. Thanks, Harvey.

simonsinclair:
@HarveyPeachey and I a11y swarmed on this yesterday. This final tweak means the SVG's `fill` colour is inherited from `color`, which adds better support for Windows' HCM.

hcm

packages/components/psammead-play-button/src/index.jsx
simonsinclair: Here we're actually passing the `datetime` value to the `datetime` prop, so that it can be used internally to set a conditional style. Whereas using `dateTime` would create a datetime="attribute" on the wrapping div. If you take a look at L39-45, hopefully this will make sense.

packages/components/psammead-play-button/src/index.jsx
simonsinclair: Yeah, it would be cool if you could pass a classname to `mediaIcons` in a similar way you can to Styled Components components. 😄

dr3 (approved):

HarveyPeachey (approved):
LGTM

packages/components/psammead-play-button/README.md
```diff
@@ -0,0 +1,75 @@
...
+`npm install @bbc/psammead-play-button`
+
+## Props
+
+<!-- prettier-ignore -->
+| Argument       | Type      | Required | Default   | Example                  |
+| -------------- | --------- | -------- | --------- | ------------------------ |
+| service        | string    | Yes      | N/A       | `'news'`                 |
+| title          | string    | Yes      | N/A       | `'Foo.'`                 |
+| onClick        | function  | Yes      | N/A       | `() => {}`               |
```
pjlee11, line 22: Would `onClick` be required for someone using this open source? Feels like this prop is only required internally.

packages/components/psammead-play-button/README.md
```diff
@@ -0,0 +1,75 @@
...
+| duration       | string    | No       | `null`    | `'2:30'`                 |
+| durationSpoken | string    | No       | `null`    | `'2 minutes 30 seconds'` |
+| datetime       | string    | No       | `null`    | `'PT2M30S'`              |
+| type           | string    | No       | `'video'` | `'audio'`                |
+| className      | string    | No       | `null`    | `'bar'`                  |
```
pjlee11, line 27: `className` is an ingrained part of React probably doesn't need to be mentioned as a prop here.

packages/components/psammead-play-button/README.md
simonsinclair: I deliberated over this one too and with any other element probably not, but for a `<button>` I felt it was safer and made more sense to require a handler. For the edge case where the button would do nothing, a noop could be used.

packages/components/psammead-play-button/src/index.jsx
```diff
@@ -0,0 +1,108 @@
...
+      )}
+    </Button>
+  );
+};
+
+PlayButton.propTypes = {
+  datetime: string,
+  duration: string,
+  durationSpoken: string,
+  type: oneOf(['video', 'audio']),
```
pjlee11, line 93:
```suggestion
  type: oneOf(['video', 'audio']).isRequired,
```

pjlee11: Ignore me this is optional as it defaults to video

packages/components/psammead-play-button/README.md
simonsinclair: That's a good point. The reason I mentioned it here is because this component is expecting to be styled by its parent and uses [this Styled Components functionality](https://www.styled-components.com/docs/basics#styling-any-component) to do that.

```
const PlayButton = ({ className, ... }) => {
  ...
  return (
    <Button className={className} service={service} onClick={onClick}>
    ...
```

pjlee11 (approved):
Couple notes on the README.md, but happy this can be merged and if tweaks to the README.md are done they can be done in a follow up PR - Great work 👍

packages/components/psammead-play-button/src/index.jsx
simonsinclair: A default `type` is set in `PlayButton.defaultProps`. This follows Media Indicator's behaviour.

packages/components/psammead-play-button/README.md
pjlee11: I'm not sure what you mean here, we just seem to be adding the `className` to the `<Button>`. I also can't see anywhere in https://github.com/bbc/psammead/blob/latest/packages/components/psammead-media-player/src/Placeholder/index.jsx or https://github.com/bbc/simorgh/blob/latest/src/app/containers/MediaPlayer/index.jsx where we are styling via the parent component. 🤔

simonsinclair: In order to style PlayButton from another component, like:
```
import PlayButton from '@bbc/psammead-play-button';

const StyledPlayButton = styled(PlayButton)`
  position: absolute;
  bottom: 0;
`;
```
It's required that PlayButton accept `className` as a prop and pass its value to a `className=""` attribute - confusing, I know, but take a look at the Styled Components link. It has a good example.

This isn't in Placeholder yet, but will be.

PR #2300, opened 2019-10-02
Add numeral system translator helper

### LINKED ISSUE #2299: Utility for numeral system conversion
We want to be able to easily convert between numeral systems without having to duplicate code/tests. This is currently being done in the Farsi and Pashto locales (via https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-locales/moment/helpers/stringHelper.js) and will need to be done in the new `psammead-calendars` utility too as well as within the Most Read component (for the item numbering).

Implement a helper function within `psammead-locales` (or similar) to provide a single consistent mechanism for translating numerals across our codebases. These numerals can be found here in this [file](https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-locales/src/numerals/index.js). 

Possible solution might look for the user something like (pseudocode):
```
import { EasternArabic, makeNumeralTranslator } from '@bbc/psammead-locales/numerals';
const translateNumerals = makeNumeralTranslator(EasternArabic);
const input = '123';
const output = translateNumerals(input); // output is '۱۲۳'
```

We should also update all existing cases where numeral conversion is performed by other means (e.g. `stringHelper.js` used in `fa.js` and `ps.js`) to use this approach.

We could look at standardised implementations such as https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl and assess if we can use them in SSR/support browser contexts.

**Testing notes**
A story should be created in psammead storybook to showcase the numerals, so we can check it with editorial to see if the numerals are correct. Look at the examples of stories we have at the moment, so we can easily add them to the new story.

Unit tests should be created to check the numerals are being translated correctly.

**Additional context**

jamesdonoh:
Resolves https://github.com/bbc/psammead/issues/2299

**Overall change:** _Adds a utility for converting numbers to an arbitrary specified numeral system_

**Code changes:**

- Added `makeNumeralTranslator` helper function to `/numerals`
- Added unit test coverage `makeNumeralTranslator`
- Updated README with description and example usage

---

- [NA] This PR requires manual testing


packages/utilities/psammead-locales/src/numerals/__snapshots__/index.test.js.snap
```diff
@@ -63,3 +63,5 @@ Array [
   "10",
 ]
 `;
+
+exports[`Numeral systems should return makeNumeralTranslator numerals 1`] = `[Function]`;
```
rhenshaw56: `makeNumeralTranslator` doesn't have numerals, I don't think it should be in this snap

packages/utilities/psammead-locales/src/numerals/index.test.js
```diff
@@ -1,10 +1,46 @@
...
+      ['The Magnificent 7', 'The Magnificent Ⅶ'],
+    ];
+    const translate = numerals.makeNumeralTranslator(RomanNumerals);
+
+    it.each(examples)('should translate %s', (input, expected) => {
+      expect(translate(input)).toBe(expected);
+    });
+  });
+
+  it('invalid numeral systems', () => {
```
rhenshaw56: maybe this should be a `describe` block having
```
it.each(invalidSystems)('should return ...
```
or still an `it` block with `invalid numeral systems should return something`

jamesdonoh: I did this originally but I thought it was unhelpful to list all the invalid systems in the test output so I removed it.

packages/utilities/psammead-locales/src/numerals/__snapshots__/index.test.js.snap
jamesdonoh: True, it's not used, but `jest -u` adds it automatically, not sure how to suppress it?

packages/utilities/psammead-locales/src/numerals/index.test.js
jamesdonoh: I've wrapped it in a describe block as you suggest so the output is a bit more descriptive, now it's:

```
  makeNumeralTranslator
    invalid numeral systems
      ✓ should return the input unchanged
```

Please take a look.

packages/utilities/psammead-locales/src/numerals/__snapshots__/index.test.js.snap
rhenshaw56: @jamesdonoh I think it has to do with how it's been imported
```
import * as numerals from '.'; // will import all exports in '.'
```

simonsinclair: We could suppress this by being more explicit about what's imported. https://gist.github.com/simonsinclair/ccc86b2ff9d900e3a08f1dd4be491fcb#file-index-test-js-L1-L10

simonsinclair (commented):
This is so simple - didn't take me long to understand. 👍 All tests pass.
It would be cool if we could verify the order of a `system`'s numerals, but they don't change that often and are covered by the snapshots.

packages/utilities/psammead-locales/src/numerals/__snapshots__/index.test.js.snap
jamesdonoh: Oh yes, good idea.

jamesdonoh: I've gone with @rhenshaw56's `Array` check actually as it reduces the need to keep updating the `import` list every time we add a new numeral system.

rhenshaw56 (approved):
Looks good.

simonsinclair (approved):

jamesdonoh:
@HarveyPeachey pointed out a problem with this, which is that the preexisting numeral systems exported from `/numerals` like `EasternArabic` actually all have 11 symbols not the 10 expected by this helper.

To address this I've pushed a fix https://github.com/bbc/psammead/pull/2300/commits/28341ceb7da2e5861f85b23ea882a5e72366a0c2 so that these systems are accepted, but that the translator only uses the first 10 symbols of any system.

dr3 (approved):

PR #2303, opened 2019-10-03
Add logic for adding alternative calendars in timestamps

### LINKED ISSUE #2305: Implement moment-calendar utility into timestamp container
Now that the moment-calendar utility has been added to psammead, we need to implement the functionality of it into psammead-timestamp-container.

In `psammead-timestamp-container` we are currently outputting the timestamp string in a `timestamp` component. We need to make use of an optional passed prop which contains a calendar from `psammead-calendars`. This will then be used to display the alternative calendar next to the original date and time if it's not in a relative format.

**Testing notes**
Unit tests will need to be created especiall

**Additional context**

HarveyPeachey:
Resolves #2305

**Overall change:** Added logic to `psammead-timestamp-container` which can render an alternative calendar if `altCalendar` prop is passed into it.

**Code changes:**

- Refactored some timestamp utility code to separate out Unix timestamp conversion logic so we can use the new `psammead-calendars` utility
- Applied alternative calendar logic based on if the `altCalendar` prop exists.
- Only applied alternative calendars to non-relative timestamps.
- Added proptype checking for `altCalendar`.
- Added snapshots for checking additional calendar functionality

---


packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -38,7 +52,8 @@ const TimestampContainer = ({
       service={service}
     >
       {prefix ? `${prefix} ` : null}
-      {showRelativeTime(timestamp, isRelative, format, timezone, locale)}
+      {getDateTime(timestamp, isRelative, format, timezone, locale)}
+      {secondaryString}
```
jamesdonoh: What about something like
```
{alternateDateTime ? `${alternateDateTime} - ` : null}
{mainDateTime}
```

lokh01:
Need to update `changelog.md` to in the containers folder as well as the `package.json` version in it too.

hotinglok (commented):
LGTM, just need to update `changelog.md` to in the containers folder as well as the `package.json` version in it too.

package-lock.json
```diff
@@ -1395,6 +1395,15 @@
         "@bbc/psammead-visually-hidden-text": "^1.2.2"
       }
     },
+    "@bbc/psammead-calendars": {
```
simonsinclair, line 1374: Should this be your Timestamp Container's `package-lock` rather than Psammead's? I believe Psammead's is updated automatically.

package.json
```diff
@@ -53,6 +53,7 @@
     "@bbc/psammead-assets": "^2.4.1",
     "@bbc/psammead-brand": "^5.0.4",
     "@bbc/psammead-caption": "^2.2.12",
+    "@bbc/psammead-calendars": "^1.0.0",
```
simonsinclair: Does this need to be a dependency of Timestamp Container, rather than Psammead? See above comment.

packages/containers/psammead-timestamp-container/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 2.5.0 | [PR#2303](https://github.com/bbc/psammead/pull/2303) Added logic for adding alternative calendars to timestamps |
```
simonsinclair: Would replacing `Added logic for adding...` with `Add option for alternative calendar timestamps` make this more readable?

packages/containers/psammead-timestamp-container/src/index.stories.jsx
```diff
@@ -138,6 +143,7 @@ storiesOf('Containers|TimestampContainer', module)
             script={script}
             locale={locale}
             service={service}
+            altCalendar={getAltCalendar() || null}
```
simonsinclair: Does this need `|| null` when `getAltCalendar()` otherwise returns `null`?

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
```diff
@@ -1,13 +1,15 @@
...
   formatUnixTimestamp,
   isValidDateTime,
-  showRelativeTime,
+  getDateTime,
 } from './timestampUtilities';
 import timestampGenerator from './helpers/testHelpers';
 
 const timezone = 'Europe/London';
 const timestamp = 1539969006000; // 19 October 2018
+const locale = 'en-gb';
```
simonsinclair, line 11: I'm curious why `formatUnixTimestamp()` sets the default locale to be `'en-gb'`, but `getDateTime()` doesn't - mentioning in case we can improve consistency.

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
```diff
@@ -25,6 +27,14 @@ describe('Timestamp utility functions', () => {
     });
   });
 
+  describe('unixTimestampToMoment', () => {
+    it("should a moment that's valid", () => {
```
simonsinclair: Did you mean `"should be a valid moment"`? 😄

simonsinclair (changes requested):
Great work, @HarveyPeachey! I've left a few comments - the first two should help with this error:
```
ERROR in ./packages/containers/psammead-timestamp-container/src/index.stories.jsx
Module not found: Error: Can't resolve '@bbc/psammead-calendars' in '/Users/simon/Documents/BBC/psammead/packages/containers/psammead-timestamp-container/src'
```

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
HarveyPeachey: I was as well but I assumed it's for a reason because it was explicitly declared before I refactored [(previous code)](https://github.com/bbc/psammead/blob/0eac2370cf1200bcdfdddc70f1a11f2c9b2b58c3/packages/containers/psammead-timestamp-container/src/timestampUtilities.js#L38). So to keep it consistent and not break code elsewhere I kept it in.

@jblaut What's the reason behind not localising datetime?

package-lock.json
HarveyPeachey: Yeah I asked @sareh about this. Because `@bbc/psammead-calendars` is only being used in storybook for this component, all storybook dependencies should be added at the the top level rather than component level.

packages/containers/psammead-timestamp-container/src/index.stories.jsx
HarveyPeachey: You're right left it in by accident I'll remove

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
HarveyPeachey: Oopsie I'll amend

package-lock.json
HarveyPeachey: As an example, the [`index.stories.jsx`](https://github.com/bbc/psammead/blob/0eac2370cf1200bcdfdddc70f1a11f2c9b2b58c3/packages/containers/psammead-timestamp-container/src/index.stories.jsx#L14) for timestamp containers has `moment-locales` imported in but it isn't added as a [dependency](https://github.com/bbc/psammead/blob/latest/packages/containers/psammead-timestamp-container/package.json) at component level

packages/containers/psammead-timestamp-container/src/index.stories.jsx
```diff
@@ -122,13 +123,17 @@ storiesOf('Containers|TimestampContainer', module)
...
       componentFunction: ({ locale, script, service }) => {
         updateAllTimestamps();
-
         const storyTimestamp = select(
           'Timestamp',
           timestamps,
           timestamps['Fixed (27 Aug 2019)'],
         );
-
+        const getAltCalendar = () => {
```
ryanmccombe: If this is going to be extended, I'd usually recommend writing these mapper-type functions using objects where possible:

```
const getAltCalendar = () => ({
  fa: jalaali,
  ps: jalaali
}[locale]);
```

ryanmccombe (dismissed):

packages/containers/psammead-timestamp-container/src/index.stories.jsx
HarveyPeachey: That's a good idea as more calendars could be added in the future

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
j-pendlebury: https://github.com/bbc/psammead/blob/0eac2370cf1200bcdfdddc70f1a11f2c9b2b58c3/packages/containers/psammead-timestamp-container/src/timestampUtilities.js#L38

what the line above does is it sets the locale to `en-gb`, unless another locale is provided when the function is called. Basically instead of having to pass in the locale and having to set it to `en-gb` every single time, the locale only has to be passed in when we want it to be anything other than that.

simonsinclair: Would it make sense for `getDateTime()` to set a default locale too?

packages/containers/psammead-timestamp-container/src/timestampUtilities.js
```diff
@@ -30,19 +30,23 @@ export const isValidDateTime = dateTime => {
   return !isNaN(new Date(dateTime)); // eslint-disable-line no-restricted-globals
 };
 
-// when using this function, we recommend using webpack configuration to only load in the relevant timezone, rather than all of moment-timezone
+// when using the following 2 functions, we recommend using webpack configuration to only load in the relevant timezone, rather than all of moment-timezone
+export const unixTimestampToMoment = (timestamp, locale) => {
+  moment.locale(locale);
+  return moment(timestamp);
+};
```
ghost: I find the function name to be a little miss-leading. Instead of only converting unix timestamp to moment, it also sets the locale.

HarveyPeachey: The reason it sets the locale here is because the moment needs the current locale of the service to figure out if it needs an alternative calendar: https://github.com/bbc/psammead/blob/530e70f7ec06cadefaa2161eda99b7f17fd461f6/packages/containers/psammead-timestamp-container/src/index.jsx#L44

ghost: Ok, But `moment(timestamp)` does not depend on the `locale` to be set, therefore should be moved outside the function.

HarveyPeachey: I don't think we need a function just for setting the locale though. How about changing the name to `unixToLocaleMoment`?

ghost: We don't need another function to set locale. Why do we need to set `locale` though? I can see we are already setting the locale here https://github.com/bbc/psammead/pull/2303/files#diff-9167111d8797dcd3c0e7444464a1d1c7R31-R33. Is there something am missing?

HarveyPeachey: Ahh completely missed that you're right. I'll remove the unnecessary locale setter thanks 😃

packages/containers/psammead-timestamp-container/src/index.test.jsx
```diff
@@ -63,4 +64,32 @@ describe('Timestamp', () => {
...
+      timestamp={defaultTimestamp}
+      dateTimeFormat="YYYY-MM-DD"
+      format="D MMMM YYYY"
+      isRelative
+      script={latin}
+      locale="fa"
+      service="persian"
+      altCalendar={jalaali}
+    />,
+  );
```
jamesdonoh: Can we nest these inside an appropriate `describe` block to give the suite a bit of structure? Also please could you make the description more descriptive, e.g. "should not add alternative calendar _if timestamp is relative_"?

packages/containers/psammead-timestamp-container/src/index.test.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { isNull, shouldMatchSnapshot } from '@bbc/psammead-test-helpers';
 import { latin } from '@bbc/gel-foundations/scripts';
+import { jalaali } from '@bbc/psammead-calendars';
```
jamesdonoh: The tests here do not really need to know anything about the Jalaali calendar. Rather than use the real implementation, for greater test isolation I would create a simple `mockCalendar` object here that provides the `formatDate` function with a predictable result.

packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -56,6 +73,9 @@ TimestampContainer.propTypes = {
   script: shape(scriptPropType).isRequired,
   locale: string,
   service: string.isRequired,
+  altCalendar: shape({
+    formatDate: func,
```
jamesdonoh: Could this be `func.isRequired` to be stricter?

Also we should add documentation for the new prop.

packages/containers/psammead-timestamp-container/src/timestampUtilities.js
jamesdonoh: Good spot @sadickisaac but I'd say the lines you link to are problematic, as they set the default moment locale 'globally' (https://momentjs.com/docs/#/i18n/changing-locale/) which can have unwanted side-effects on code that makes assumption about the default locale. Instead we should normally set the locale 'locally' on moment objects we create as described at https://momentjs.com/docs/#/i18n/instance-locale/

Unless anyone is aware of any reason why we can't?

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
jamesdonoh: @jblaut Thanks for the context, are you aware of any code that is relying on that locale defaulting behaviour? It may have made sense when we were only building articles for News, but I'd have expected that now the locale should be explicitly supplied everywhere it's needed, so I'd be interested to know if we can remove this default (and fix any bugs that arise due to imprecise assumptions elsewhere in the code). What do others think?

Also see my comment above at https://github.com/bbc/psammead/pull/2303#discussion_r332470068

jamesdonoh (changes requested):

packages/containers/psammead-timestamp-container/src/index.jsx
HarveyPeachey: I originally had this, but then I thought if a new calendar is added, something other than `formatDate` could be used as a function for formatting. 

Which made me think that we might have to refactor `psammead-calendars` so that all future calendars will make use of a `formatDate` function so we can use `func.isRequired`.

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
j-pendlebury: So the requirement came from this issue https://github.com/bbc/psammead/issues/651. We wanted to keep the `datetime` attribute in English locale, so at that point `formatUnixTimestamp` was called to get the timestamp, and `showRelativeTime` was called to get the timestamp in the correct locale. `showRelativeTime` would then call `formatUnixTimestamp` with the locale passed in...

So basically, initially we needed `formatUnixTimestamp` to always return `en-gb` locale, unless called from `showRelativeTimestamp`.

packages/containers/psammead-timestamp-container/src/index.jsx
jamesdonoh: At the moment the only way an alt calendar can be used is by calling `formatDate` like on line https://github.com/bbc/psammead/pull/2303/files#diff-9167111d8797dcd3c0e7444464a1d1c7R44 so I think it's fine to add `isRequired` for some extra help debugging. If we expand the calendar API later it's easy to change.

HarveyPeachey: Okay I'll change it to required

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
jamesdonoh: If the requirement is to ensure that the `datetime` attribute output is in en-gb can we not explicitly set that when generating the value for that attribute, rather than relying on a default locale?

packages/containers/psammead-timestamp-container/src/timestampUtilities.test.js
```diff
@@ -25,6 +27,12 @@ describe('Timestamp utility functions', () => {
     });
   });
 
+  describe('unixTimestampToMoment', () => {
+    it('should be a valid moment', () => {
+      expect(unixTimestampToMoment(timestamp)).toHaveProperty('isValid');
```
rhenshaw56, line 31: question, the test reads as `should be a valid moment` does this mean we expect that all valid "moment" have an "isValid" property ? why not `instanceof moment` instead ?

jamesdonoh: I don't think we can rely on the object having been created with the `moment` constructor as this isn't described in Moment's public API, whereas the presence of `isValid` should be reliable.

jamesdonoh: Could clarify the description maybe to say not "be a valid moment" but "be a moment object" (as we aren't actually calling `isValid`) but I don't think it's essential.

hotinglok (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 1.8.20 | [PR#2303](https://github.com/bbc/psammead/pull/2303) Added psammead-calendars as a dependency |
```
jamesdonoh: Doesn't this need changing in `package.json` and `package-lock.json` too?

rhenshaw56 (approved):

RayNjeri (approved):

simonsinclair (approved):

jamesdonoh (approved):

packages/containers/psammead-timestamp-container/package.json
```diff
@@ -29,6 +29,9 @@
     "@bbc/psammead-timestamp": "^2.2.12",
     "moment-timezone": "^0.5.26"
   },
+  "devDependencies": {
+    "@bbc/psammead-calendars": "^2.0.0"
```
dr3: This is at the top level, so doesnt need to be here

dr3 (changes requested):

dr3 (approved):

PR #2304, opened 2019-10-03
Add utility package rich-text-transforms

### LINKED ISSUE #2301: Create rich-text-transforms utility package
For the MAP work we need to find a way to render the `candy_xml` format we get from Ares. The WebCore team have built an XML parser that we can make use of however it is not publicly available on `npm`. Therefore we should host it on psammead so that it can be used within Simorgh. 

We should use the version of this library that has `ramada` and `xml2js` as it is smaller and safer - `fabl-modules/pull/126`

Copy the `rich-text-transforms` library from the `fabl-modules` repo and deploy it via psammead.

Rather than copying the library we could make the other repo open source but this would be a slow process that would block the use of this library in Simorgh. 

**Testing notes**
No testing required 

- ~[ ] This feature is expected to need manual testing.~

**Additional context**

pjlee11:
Resolves #2301

**Overall change:** Creates the utility package `rich-text-transforms` which is a fork of the FABL module (github `fabl-modules/libraries/rich-text-transforms`).

**Code changes:**

---

### DIFF
--- a/packages/utilities/psammead-rich-text-transforms/CHANGELOG.md
+++ b/packages/utilities/psammead-rich-text-transforms/CHANGELOG.md
@@ -1,0 +1,6 @@
+# @bbc/psammead-rich-text-transforms Changelog
+
+<!-- prettier-ignore -->
+| Version | Description |
+| ------- | ----------- |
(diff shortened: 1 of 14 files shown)


dr3 (approved):

chris-hinds (approved):
LGTM

PR #2317, opened 2019-10-07
Add psammead-play-button to Media Player placeholder

### LINKED ISSUE #2248: Add a play button to Media Player placeholder

simonsinclair:
Resolves https://github.com/bbc/psammead/issues/2248.

**Overall change:** This change adds `psammead-play-button` to the bottom corner of the `psammead-media-player` placeholder.

**Code changes:**

- Import `PlayButton` into `Placeholder`.
- Add parent-specific styles to `PlayButton` - position, etc.
- Pass `service` and `mediaInfo` (containing all required data for PlayButton) from `CanonicalMediaPlayer` to `Placeholder`.
- Update `Placeholder` and `CanonicalMediaPlayer` tests.
- Update `CanonicalMediaPlayer` stories.

**A11y Swarm**
_Checklists (Simon):_
- [x] https://bbc.github.io/accessibility-news-and-you/accessibility-news-and-testers
  - [x] Axe
  - [x] Design
  - [x] Colour Contrast
  - [x] Colour blindness
  - [x] Windows HCM
  - [x] FF Colour Preferences
  - [x] Text 200%, Page 200%
  - [x] Icons
  - [x] Heading Order
  - [x] Keyboard Navigation
  - [x] No JS (n/a)
- [x] Screen readers
  - [x] ZoomText Magnifier/Reader
  - [x] Dragon NaturallySpeaking
  - [x] JAWS
  - [x] Read&Write (n/a)

_Mac, iOS, and Android (Ray):_ 
- [x] VoiceOver iOS (iPhone/iPad)
- [x] VoiceOver OS (Mac)
- [x] TalkBack (Chrome on Android)

**Notes:**


packages/components/psammead-media-player/src/Placeholder/index.jsx
```diff
@@ -12,15 +14,48 @@ const StyledPlaceholder = styled.div`
...
 Placeholder.propTypes = {
   onClick: func.isRequired,
+  service: string.isRequired,
   src: string.isRequired,
+  mediaInfo: shape({
+    title: string.isRequired,
+    datetime: string,
+    duration: string,
+    durationSpoken: string,
+    type: oneOf(['video', 'audio']),
```
HarveyPeachey: Does this need a `.isRequired`? I know that in your play button component it defaults to video. But what happens if someone adds audio content and for some reason, the `type` isn't set.

packages/components/psammead-media-player/src/Placeholder/index.test.jsx
```diff
@@ -1,28 +1,54 @@
 import React from 'react';
-import { render, fireEvent } from '@testing-library/react';
+import { render, fireEvent, getByText } from '@testing-library/react';
 import { shouldMatchSnapshot } from '@bbc/psammead-test-helpers';
 import Placeholder from '.';
 
 describe('Media Player: Placeholder', () => {
   const mockOnClick = jest.fn();
+  const mediaInfo = {
```
HarveyPeachey: Might be beneficial to add a test/snapshot which doesn't pass through `duration`, `durationSpoken` and `datetime`?

HarveyPeachey (commented):
Looks good! Just a couple of questions. I know you might not be finished with the PR yet because I think you mentioned an a11y swarm for this as well.

packages/components/psammead-media-player/src/Placeholder/index.jsx
simonsinclair: This is a good point - someone will need to look into this while I'm away. My concerns are:
- Requiring `type` in a component that doesn't use it directly
- Introducing inconsistency (requiring it in one and not the other)
- The relationship `type` has with `skin`. Could we infer the value of `type` from this and drop it altogether?

cc @jamesdonoh

packages/components/psammead-media-player/src/Placeholder/index.jsx
```diff
@@ -12,15 +14,48 @@ const StyledPlaceholder = styled.div`
...
+
+const Placeholder = ({ onClick, service, src, mediaInfo }) => {
+  const { title, datetime, duration, durationSpoken, type } = mediaInfo;
+  return (
+    <StyledPlaceholder onClick={onClick}>
+      <StyledPlayButton
+        title={title}
+        service={service}
+        onClick={event => {
+          event.stopPropagation();
```
jamesdonoh: What is the effect of not having `stopPropagation` here?

packages/components/psammead-media-player/src/Placeholder/index.jsx
jamesdonoh: If it's not required in [psammead-play-button](https://github.com/bbc/psammead/tree/latest/packages/components/psammead-play-button) and defaults to `video`, I can't see why we'd need to make it required here. If the calling code fails to pass a `type` then AFAICS showing the standard play button (which visually at least could equally well apply to video or audio) is like a reasonable default.

And I'd be wary of coupling it to the `skin`, we shouldn't assume the type of the content and the skin being used will always go hand-in-hand.

packages/components/psammead-media-player/src/index.stories.jsx
```diff
@@ -3,10 +3,18 @@ import { storiesOf } from '@storybook/react';
...
+  durationSpoken: '2 minutes 30 seconds',
+  datetime: 'PT2M30S',
+};
+
 storiesOf('Components|Media Player', module).add('default', () => (
   <CanonicalMediaPlayer
     src="https://www.test.bbc.co.uk/ws/av-embeds/articles/c3wmq4d1y3wo/p01k6msp/en"
     placeholderSrc="https://ichef.bbci.co.uk/news/640/cpsdevpb/4eb7/test/ba7482d0-cca8-11e8-b0bf-f33155223fc4.jpg"
+    service="news"
+    mediaInfo={{ title: 'Dog chases cat.', ...withDuration }}
```
jamesdonoh, line 17: Might be nice to add a story for audio as well as video, what does anyone else think?

packages/components/psammead-media-player/src/Placeholder/index.jsx
RayNjeri: Am guessing what Simon was aiming for here with event.stopPropagation(); is to keep the onClick event from bubbling any further up into the DOM Avoiding more click events from happening even after the play button is already clicked.

RayNjeri: but on second thought we do not need it as we only require the onClick to be called once

HarveyPeachey (commented):
LGTM but are we not adding an audio story from James's suggestion?: https://github.com/bbc/psammead/pull/2317#discussion_r334436145

packages/components/psammead-media-player/src/index.stories.jsx
RayNjeri: Don't we already have audio stories from L31 unless am missing something? 🤔 
cc: @HarveyPeachey

packages/components/psammead-media-player/src/Placeholder/index.jsx
```diff
@@ -12,15 +14,57 @@ const StyledPlaceholder = styled.div`
...
+  ${StyledPlaceholder}:focus & {
+    background-color: ${C_POSTBOX};
+  }
+  /* stylelint-enable */
+`;
+
+const Placeholder = ({ onClick, service, src, mediaInfo }) => {
+  const { title, datetime, duration, durationSpoken, type } = mediaInfo;
+  return (
+    <StyledPlaceholder onClick={() => {}}>
```
simonsinclair: Why is this a no-op? Both the placeholder (and play button) should trigger playback.

packages/components/psammead-media-player/src/Placeholder/index.jsx
simonsinclair: Please see https://github.com/bbc/psammead/pull/2317#discussion_r335337769.

dr3: Not sure I understand @simonsinclair, even from https://github.com/bbc/psammead/pull/2317#discussion_r335337769, can you explain more please?

packages/components/psammead-media-player/src/index.jsx
```diff
@@ -35,7 +37,12 @@ export const CanonicalMediaPlayer = ({
   return (
     <StyledContainer portrait={portrait}>
       {placeholderActive ? (
-        <Placeholder onClick={handlePlaceholderClick} src={placeholderSrc} />
+        <Placeholder
+          onClick={handlePlaceholderClick}
```
dr3: Not sure this onClick works atm, the player doesnt appear for me

packages/components/psammead-media-player/src/Placeholder/index.jsx
```diff
@@ -12,15 +14,59 @@ const StyledPlaceholder = styled.div`
...
+  mediaInfo: shape({
+    title: string.isRequired,
+    datetime: string,
+    duration: string,
+    durationSpoken: string,
+    type: oneOf(['video', 'audio']),
+  }),
+};
+Placeholder.defaultProps = {
+  mediaInfo: shape({
```
dr3, line 61: `defaultProps` are the values it falls back to using :) an example would be like default strings. Most of these should be `.isRequired` above, and not in the default props :)

packages/components/psammead-media-player/src/Placeholder/index.jsx
```diff
@@ -12,15 +14,59 @@ const StyledPlaceholder = styled.div`
...
+    title: string.isRequired,
+    datetime: string,
+    duration: string,
+    durationSpoken: string,
+    type: oneOf(['video', 'audio']),
```
dr3:
```suggestion
    type: 'video',
```

packages/components/psammead-media-player/src/index.jsx
```diff
@@ -60,6 +67,14 @@ CanonicalMediaPlayer.propTypes = {
   showPlaceholder: bool,
   src: string.isRequired,
   skin: oneOf(['classic', 'audio']),
+  service: string.isRequired,
+  mediaInfo: shape({
+    title: string.isRequired,
+    datetime: string,
```
dr3, line 81: lines 73-75 should be required :)

dr3 (changes requested):
Looks great :) just a few tweaks

packages/components/psammead-media-player/src/Placeholder/index.jsx
simonsinclair: Hey @dr3, in the current PR, the play button's onClick calls the same handler as the parent (which needs to remain clickable), so using stopPropagation prevents the handler from being called twice.
Perhaps it would make more sense to use a no-op in the play button and let it bubble to the parent.

packages/components/psammead-media-player/src/index.stories.jsx
simonsinclair: Yes, I think we could add one for Audio with a placeholder. This is not the same as  'Audio Skin' from L31.

packages/components/psammead-media-player/src/index.jsx
dr3: > "Default props are only shallowly merged."
https://stackoverflow.com/questions/38123498/how-do-you-provide-default-props-for-nested-shape-in-react

Ah, if these arnt infact required then this is fine as is :)

packages/components/psammead-media-player/src/Placeholder/index.jsx
RayNjeri: actually these props are not required

packages/components/psammead-media-player/src/Placeholder/index.test.jsx
```diff
@@ -1,28 +1,76 @@
...
   );
 
-  it('should call onClick when clicked', () => {
+  shouldMatchSnapshot(
+    'should render a video placeholder without duration',
+    <Placeholder
+      onClick={mockOnClick}
+      src="http://foo.bar/placeholder.png"
+      service="news"
+      mediaInfo={{ title: 'Dog chases cat.', ...withDuration }}
```
AlistairGempf: Why does the title of this test say `without duration` but the `mediaInfo` includes `...withDuration`? I may be missing something but I don't see any difference between this test and the one above.

12:
Worth noting for the future that when we pass `ui.cta.direction: 'rtl'` to SMP for right-to-left services, the play button will be aligned to the bottom-right of the container. We should mirror this in the placeholder play icon placement.

It shouldn't be a requirement for now though.

packages/components/psammead-media-player/src/Placeholder/index.test.jsx
RayNjeri: Aaah I see you what you mean, the second test was meant to test for when the `duration` prop is not provided.

RayNjeri: addressed and updated in f2026f6

packages/components/psammead-media-player/src/Placeholder/index.jsx
```diff
@@ -12,15 +14,57 @@ const StyledPlaceholder = styled.div`
...
 Placeholder.propTypes = {
   onClick: func.isRequired,
+  service: string.isRequired,
   src: string.isRequired,
+  mediaInfo: shape({
+    title: string.isRequired,
+    datetime: string,
+    duration: string,
+    durationSpoken: string,
+    type: 'video',
```
simonsinclair: I think you copied the default prop value here by mistake. This should be a valid prop type: `oneOf(['audio', 'video'])`.

packages/components/psammead-media-player/src/Placeholder/index.jsx
```diff
@@ -12,15 +14,57 @@ const StyledPlaceholder = styled.div`
...
+  }),
+};
+Placeholder.defaultProps = {
+  mediaInfo: shape({
+    title: null,
+    datetime: null,
+    duration: null,
+    durationSpoken: null,
+    type: 'video',
```
simonsinclair, line 65: 👍

simonsinclair (commented):
Nice work, @RayNjeri. Please see my comment below.

jamesdonoh (approved):

packages/components/psammead-media-player/src/Placeholder/index.jsx
```diff
@@ -12,15 +14,57 @@ const StyledPlaceholder = styled.div`
...
+    title: string.isRequired,
+    datetime: string,
+    duration: string,
+    durationSpoken: string,
+    type: oneOf(['audio', 'video']),
+  }),
+};
+Placeholder.defaultProps = {
+  mediaInfo: shape({
+    title: null,
```
dr3: Doesn't need a default as its required :)
```suggestion
```

dr3 (approved):
Great work :D

simonsinclair:
@RayNjeri and I have completed the dev a11y swarm - please see issue body for details.

jamesbrumpton:
lgtm

PR #2342, opened 2019-10-07
Fix Moment defineLocale warning

### LINKED ISSUE #2272: Moment defineLocale warning in Storybook
**Describe the bug**
In Storybook, if you look in the console, you'll see the deprecation warning:

```
Deprecation warning: use moment.updateLocale(localeName, config) to change an existing locale. moment.defineLocale(localeName, config) should only be used for creating a new locale
```

Attempted to fix in https://github.com/bbc/psammead/pull/2240, but no luck. Seems like it's an issue with the locales imported in the Timestamp Container storybook file.

DenisHdz:
Resolves #2272

**Overall change:** 
It seems that `webpack` doesn't differ between for example ` import '@bbc/psammead-locales/moment/am'; ` and `import ./am`, so both locales are imported in Storybook, which provokes a warning as trying to define the locales for a second time.

For this reason, we can remove the imports from `TimestampContainer` to avoid the warning, since they are already imported and accessible.

**Code changes:**
- Remove imports from `TimestampContainer`
- Bump package

---


ghost (approved):

thekp (approved):

packages/containers/psammead-timestamp-container/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 2.4.22 | [PR#2342](https://github.com/bbc/psammead/pull/2289) Remove psammead-locales imports from story |
```
ryanmccombe:
```suggestion
| 2.4.22 | [PR#2342](https://github.com/bbc/psammead/pull/2342) Remove psammead-locales imports from story |
```

ryanmccombe (approved):

j-pendlebury (approved):

HarveyPeachey:
@DenisHdz I'm not sure where to post about this but the story for TimeStamp containers is now broken after this merge. 

If you go to [storybook](https://bbc.github.io/psammead/?path=/story/containers-timestampcontainer--default) and set the service knob to `korean` for example and then set it to a service which doesn't have a default locale like `pashto` it keeps the locale of the previous timestamp which is in korean.

DenisHdz:
@HarveyPeachey That was already happening before, this was about fixing the warning that was caused due to importing the locales in this container when they were already imported in the `psammead-locales` story.

DenisHdz:
> then set it to a service which doesn't have a default locale like pashto it keeps the locale of the previous timestamp which is in korean.

@HarveyPeachey How so? All the services should have a locale defined in `text-variants`. If doesn't work, you should link the packages:

`cd` to `psammead-storybook-helpers`, run `npm link` and then `cd` to `psammead-timestamp-container` and run `npm link @bbc/psammead-storybook-helpers`.

PR #2313, opened 2019-10-04
Fix grid css for Useful Links

OlgaLyubin:
Resolves #2311 

**Overall change:** 
Changed grid items alignment and columns gap.
Changed order of elements displayed in columns (from _row by row_ to _column by column_) so that the screen reader announces them correctly.

**Code changes:**
Align items to the top instead of the centre. Add column grid gap of 30px.
Change grid to include rows amount so that the order of the elements is correct

---

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Automated jest tests added (for new features) or updated (for existing features)
- [x] This PR requires manual testing

PR #2351, opened 2019-10-08
Sitewide Links fix for IE11

### LINKED ISSUE #2316: Sitewide Links display incorrectly in IE11
**Describe the bug**
Sitewide links display vertically instead of horizontally in the grid in IE11. This is because we use `@supports not grid`, but this doesn't work in IE11. https://github.com/bbc/psammead/blob/latest/packages/components/psammead-sitewide-links/src/List/index.jsx#L84

**To Reproduce**
Open up any site or Psammead storybook in IE11 and look at the sitewide links in the footer.

**Expected behaviour**
They should behave like the other browsers and display in horizontal grid.

**Screenshots**
How it looks in IE11:

igbo footer
sitewidelinks

How it should look:
Screen Shot 2019-10-07 at 09 24 16
Screen Shot 2019-10-07 at 09 24 34

- Browser: IE11

**Testing notes**
Need to check the component in all browsers we support to make sure it looks consistent.

**Additional context**

j-pendlebury:
Resolves https://github.com/bbc/psammead/issues/2316

**Overall change:** Add column support to IE11 without `grid`

**Code changes:**

- Remove `@supports not` because they don't work on IE11 👎 
- Use `column-count` to select how many columns there should be per breakpoint, same as on Chrome
- Use `column-span` for the first `li` to make sure the first row spans all columns
- Add `column-gap` for the spacing

## Screenshots:

Sitewide Links on Chrome [TOP is current LIVE component, BOTTOM is component with changes in this PR]

Screen Shot 2019-10-08 at 11 02 10

---
Sitewide Links on IE11 with fix

Screen Shot 2019-10-08 at 11 05 59

---
Sitewide Links on Chrome with `display: grid` disabled

image

---


packages/components/psammead-sitewide-links/src/List/index.jsx
```diff
@@ -73,17 +73,13 @@ const StyledList = styled.ul`
     padding: ${GEL_SPACING} 0;
     margin-bottom: ${GEL_SPACING};
     grid-column: 1/-1;
-    @supports not (${grid}) {
-      width: 100%;
-    }
+    width: 100%;
```
DenisHdz, line 82: Wondering if we would also need this:
```
@supports (${grid}) {
    width: initial;
}
```

packages/components/psammead-sitewide-links/src/List/index.jsx
```diff
@@ -73,17 +73,13 @@ const StyledList = styled.ul`
...
   }
 `;
 
 const StyledListItem = styled.li`
+  display: inline-block;
   min-width: 50%;
```
DenisHdz, line 88: Same here:
```
@supports (${grid}) {
    display: block;
    min-width: initial;
 }
```

DenisHdz: @OlgaLyubin is doing something similar in her PR https://github.com/bbc/psammead/blob/67db16d966d8122333c2cbb5dcebd278dc7f1995/packages/components/psammead-useful-links/src/index.jsx

packages/components/psammead-sitewide-links/src/List/index.jsx
j-pendlebury: I could add this just to be safe but it doesn't look like the component is affected by this change if you look at the screenshots above

DenisHdz: Yep, just tested and it works fine without this, as when supporting grid the first list item will take the whole column 👍🏻

DenisHdz (approved):

j-pendlebury:
image

Chromatic comparison. No changes in Chrome which supports CSS Grid

jroebu14 (approved):

j-pendlebury:
image
PR went back to In Progress because the issue above was found during testing ^

packages/components/psammead-sitewide-links/src/List/index.jsx
```diff
@@ -73,16 +73,16 @@ const StyledList = styled.ul`
...
-  @supports not (${grid}) {
-    display: inline-block;
+
+  @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
```
DenisHdz: Is this to detect we are on IE? What about using `column-gap`, as we did in the `UsefulLinks`?
https://github.com/bbc/psammead/pull/2313/files#diff-72dd5167119d67cc68065f80ca0f141fR39

j-pendlebury: do you mean use `column-gap` inside of the `@media` block or instead of it?

DenisHdz: Instead of it and adding padding, sorry.

j-pendlebury: but then how do we make sure this is only added to IE? that was my issue with this change. The media query looks horrible but it was the only way (that I know of) to make sure the padding is only added to IE

j-pendlebury: Moved this discussion to a private chat and we used a solution similar to the Useful Links IE fix [here](https://github.com/bbc/psammead/blob/latest/packages/components/psammead-useful-links/src/index.jsx#L38-L39).

DenisHdz (approved):
LGTM, thanks for improving it 🚀

jamesbrumpton:
lgtm

PR #2364, opened 2019-10-10
use xmldoc instead of xml-js, with refactor to accomadate new package

### LINKED ISSUE #2363: Use a small XML parser
As described in this comment https://github.com/bbc/simorgh/pull/4184#issuecomment-540079051 the package `xml-js` in `psammead-rich-text-transforms` isn't big as a published size but the install size is significantly bigger than expected. 

Find and use a smaller parser for `psammead-rich-text-transforms`

Accept the 90KB+ increase and split the bundles so `xml-js` is only included on page that require it to render. 

**Testing notes**
[Tester to complete]

- ~[ ] This feature is expected to need manual testing.~

### Additional context
Tried and not suitable: 

**xml2json-light**
Ignores text between nodes 
`</paragraph><bold>this is okay</body>this is lost</paragraph>`

**fast-xml-parser**
Combines matching nodes and doesn't keep the order when multiple matched

Now trying
**xmldoc** 
Looks promising with the complicated example.

pjlee11:
Resolves #2363

**Overall change:** Swaps out `xml-js` to `xmldoc` as the dependency of this package and updates the code based on the new packages format.

**Code changes:**

- Swaps the checking the `node.type as text` to `node.constructor.name === 'XmlTextNode`
- Removes `findEl` and `findAtr` as the new format is less nested
- Has to loop over each child of the urlLink to build out the `text`, `locator` and `blocks`
- Swaps `elements` for `children` due to the new format 
- Checks if the array is empty with (`childNode.children.length !== 0`) before recursively calling `convertToBlocks()`

---

### DIFF
--- a/packages/utilities/psammead-rich-text-transforms/src/candy-xml.js
+++ b/packages/utilities/psammead-rich-text-transforms/src/candy-xml.js
@@ -9,7 +9,7 @@ const attributeTags = ['bold', 'italic'];
 const supportedXmlNodeNames = ['paragraph', 'link', 'url', ...attributeTags];
 
 const isXmlNodeSupported = node => {
-  if (node.constructor.name === 'XmlTextNode') {
+  if (path(['constructor', 'name'], node) === 'XmlTextNode') {
     return true;
   }
 
--- a/packages/utilities/psammead-rich-text-transforms/src/candy-xml.js
+++ b/packages/utilities/psammead-rich-text-transforms/src/candy-xml.js
@@ -1,5 +1,5 @@
 /* eslint-disable no-use-before-define */
-const { xml2json } = require('xml-js');
+const xmldoc = require('xmldoc');
 const path = require('ramda/src/path');
 const pathOr = require('ramda/src/pathOr');
 const is = require('ramda/src/is');
@@ -9,8 +9,8 @@ const attributeTags = ['bold', 'italic'];
 const supportedXmlNodeNames = ['paragraph', 'link', 'url', ...attributeTags];
 
 const isXmlNodeSupported = node => {
-  if (path(['type'], node) === 'text') {
-    return Boolean(pathOr('', ['text'], node).trim());
+  if (node.constructor.name === 'XmlTextNode') {
+    return true;
   }
 
   return supportedXmlNodeNames.includes(path(['name'], node));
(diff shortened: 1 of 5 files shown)


packages/utilities/psammead-rich-text-transforms/src/candy-xml.js
```diff
@@ -9,8 +9,8 @@ const attributeTags = ['bold', 'italic'];
 const supportedXmlNodeNames = ['paragraph', 'link', 'url', ...attributeTags];
 
 const isXmlNodeSupported = node => {
-  if (path(['type'], node) === 'text') {
-    return Boolean(pathOr('', ['text'], node).trim());
+  if (node.constructor.name === 'XmlTextNode') {
```
ryanmccombe:
```suggestion
return node.constructor.name === 'XmlTextNode';
```

ryanmccombe (approved):

ibMadbouly (approved):

packages/utilities/psammead-rich-text-transforms/test/candy-xml.test.js
```diff
@@ -418,9 +413,9 @@ test('returns plain text if wrapped in an unsupport xml node', () => {
 
 test('can handle chaos', () => {
   const richText = candyXmlToRichText(
-    createBody(`
-      <paragraph><foo><bar><bold><meh>Bold text in unsupported nodes</meh></bold></bar></foo> followed by normal text then <italic><this><is>some <bold>carnage<carnage> </carnage></bold></is></this></italic>.</paragraph>
```
dr3, line 422: How come you removed the space in carnage? Was it because you had to?

pjlee11: I realised the previous logic didn't work with an empty XML tag `<carnage></carange>` so removing this space actually makes this test even more complicated

pjlee11: I'll add another explicit test for an empty XML tag

pjlee11: This pointed out a problem ... will push shortly

pjlee11: The fragment carnage gets lost in the current implementation...
image

😞

pjlee11: I don't think this should hold up this PR. I've created https://github.com/bbc/psammead/issues/2378 to address this, but I'm not convinced we will ever receive input this complicated so might not be required.

packages/utilities/psammead-rich-text-transforms/src/candy-xml.js
dr3: should prob use `path` here

jroebu14 (approved):

PR #2369, opened 2019-10-10
Fix Timestamp Container service locale 

### LINKED ISSUE #2368: Timestamp Container - service not applying right locale
**Describe the bug**
The service knob is not working properly in the Timestamp Container.

**To Reproduce**

1. Go to `TimestampContainer` story
2. Select the `korean` service
3. Select the `punjabi` service
4. It keeps the locale of the previous timestamp which is in `korean`.

**Expected behaviour**

**Screenshots**

**Testing notes**

**Additional context**

DenisHdz:
Resolves #2368

**Overall change:** 
Apply the right service locale to the Timestamp container.

**Code changes:**
- Remove `moment.locale` from the container as it is already set in the `formatUnixTimestamp ` function.
- Add `psammead-locales` as a dev dependency.
- Import custom moment locales from `psammead-locales` in the `TimestampContainer` story
- Add two new assertions to test that locales are used correctly for both normal and relative timestamps
- Update snapshot

---


packages/containers/psammead-timestamp-container/src/__snapshots__/index.test.jsx.snap
```diff
@@ -110,7 +110,7 @@ exports[`Timestamp should render without a leading zero on the day 1`] = `
 
 <time
   class="c0"
-  datetime="2018-07-07"
+  datetime="۲۰۱۸-۰۷-۰۷"
```
j-pendlebury: `datetime` is not supposed to be localised https://github.com/bbc/psammead/issues/651

j-pendlebury (changes requested):
left a comment about `datetime`

HarveyPeachey:
I think this might be covered in some of the work in #2303 but I don't think it fixes the issue. 

I think what the problem is, is that locales that we have added or overridden in `psammead-locales` aren't being recognised in TimeStamp container, so when we try and change the locale moment defaults to the last used viable locale.

DenisHdz:
> I think this might be covered in some of the work in #2303 but I don't think it fixes the issue.
> 
> I think what the problem is, is that locales that we have added or overridden in `psammead-locales` aren't being recognised in TimeStamp container, so when we try and change the locale moment defaults to the last used viable locale.

They are recognized, as they are already imported in `psammead-locales`. Tested this locally and it works for me.

packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -26,10 +25,6 @@ const TimestampContainer = ({
     return null;
   }
 
-  if (locale) {
-    moment.locale(locale);
-  }
-
   return (
     <Timestamp
       datetime={formatUnixTimestamp(timestamp, dateTimeFormat, timezone)}
```
dr3: Your PR description says

>Remove moment.locale from the container as it is already in the formatUnixTimestamp function, to avoid conflicts.

But we arnt passing locale to `formatUnixTimestamp` here? Surely it defaults to `en-gb`?

dr3 (dismissed):
Can you please write a test for this container that locales are used correctly for both normal and relative timestamps please. We realllllllly should have already had some, but better late than never

packages/containers/psammead-timestamp-container/src/index.jsx
DenisHdz: I believe so, beacuse we set it to `en-gb` in the [formatUnixTimestamp](https://github.com/bbc/psammead/blob/latest/packages/containers/psammead-timestamp-container/src/timestampUtilities.js#L38) function, if not locale is passed. 

@HarveyPeachey is actually planning to always pass the `en-gb`, instead of seting a default in [this](https://github.com/bbc/psammead/blob/latest/packages/containers/psammead-timestamp-container/src/timestampUtilities.js#L38) PR.

DenisHdz: I believe so, because we set it to `en-gb` in the [formatUnixTimestamp](https://github.com/bbc/psammead/blob/latest/packages/containers/psammead-timestamp-container/src/timestampUtilities.js#L38) function, if there is no local passed. 

@HarveyPeachey is actually planning to always pass the `en-gb`, instead of setting a default in [this](https://github.com/bbc/psammead/pull/2303/files#diff-9167111d8797dcd3c0e7444464a1d1c7R50) PR.

tochwill:
Looks good - so if I understand correctly, the 'global' moment locale was being set in the timestamp container, but the formatting functions were setting the 'local' locale on each moment instance they were creating, so the locale that was being set in the code you've deleted wasn't being used anyway? I had a bunch more questions around how this component works but it looks like they're all being addressed in https://github.com/bbc/psammead/pull/2303

DenisHdz:
When running `npm run ci:packages`, neither the timestamp updates to the right locale when changing the service knob, nor the new tests added pass, whereas running `npm run install:packages` locally does.

So I am not sure anymore if the `moment.locale` was the issue or it is the `ci:packages` script, which is the one we run in our CI and seems not be using de right locale and defaults to english.

DenisHdz:
It turns out that to make the locales work when running `npm run ci:packages`, we should add the `psammead-locales` package as a dependency of the `psammead-timestamp-container`, otherwise the container can not access to them, whereas it does when running `npm run install:packages`.

We also need to import our custom locales in the `TimestampContainer` story to be able to use them, otherwise, it will always use the English moment locale. 

Those imports were removed in [this](https://github.com/bbc/psammead/pull/2342/files) PR because they were giving a warning as trying to define the locales twice because they were already imported in the `psammead-locales` story. But it seems that having the `psammead-locales` as dependency of the `psammead-timestamp-container`, doesn't clash with the other story and imports, therefore we don't get the warning.

packages/containers/psammead-timestamp-container/package.json
```diff
@@ -27,7 +27,9 @@
   "dependencies": {
     "@bbc/gel-foundations": "^3.4.0",
     "@bbc/psammead-timestamp": "^2.2.12",
-    "moment-timezone": "^0.5.26"
+    "@bbc/psammead-locales": "^2.23.0",
```
HarveyPeachey: Does this need to be in `dependencies` if it's only being used in the story? Shouldn't it just be a `devDependency`?

Because when storybook gets pushed to the live site the imports in the `index.stories.jsx` should resolve because `psammead-locales` is in the top level `package-lock.json` so `npm run ci:packages` should install it.

packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import { number, string, bool, shape } from 'prop-types';
 import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+import moment from 'moment';
```
HarveyPeachey: What's the reason for changing this from `moment-timezone` to `moment`?

packages/containers/psammead-timestamp-container/src/index.stories.jsx
```diff
@@ -8,8 +8,45 @@ import {
   text,
   withKnobs,
 } from '@storybook/addon-knobs';
-import moment from 'moment-timezone';
-import Timestamp from '.';
+import moment from 'moment';
```
HarveyPeachey, line 11: Same question as comment above.

HarveyPeachey (commented):
LGTM just a few questions

packages/containers/psammead-timestamp-container/src/index.jsx
DenisHdz: I don't see the point to import the `timezone` here when is imported and used in the formatting functions.

packages/containers/psammead-timestamp-container/package.json
DenisHdz: I installed initially as a dev dependency, but if you would like to use this package in any project/application you would still need it as a dependency of this package to make it work. 

The user is not aware of what dependencies should be installed to make packages work. We add it to our top level `package.json` and `package-lock.json`, because we created it and we are aware of it.

DenisHdz: I installed it initially as a dev dependency, but if you would like to use this package in any project/application you would still need it as a dependency of this package to make it work. 

Users should not be aware of what dependencies should be installed to make packages work.

packages/containers/psammead-timestamp-container/src/index.stories.jsx
DenisHdz: Same as above. I believe we will applying the right timezone in the formatting functions and we don't need in the stories either.

packages/containers/psammead-timestamp-container/package.json
HarveyPeachey: Would you need it as a normal dependency? The package should work fine because storybook is only really meant to be used for development and not a finished project/application. 

The only reason we have storybook available to view publically is so people can see the functionality of our components from my understanding.

DenisHdz: Let's say I want to use the `Timestamp` container in my application and I want to use the `hausa` locale. Just importing the `@bbc/psammead-local/ha` would work, without having to install `psammead-locales` in my project base level `package.json`.

packages/containers/psammead-timestamp-container/package.json
```diff
@@ -27,7 +27,9 @@
...
+    "moment-timezone": "^0.5.26",
+    "moment": "^2.24.0"
```
dr3: This should be a peer dep I think :)

packages/containers/psammead-timestamp-container/src/index.jsx
dr3: Why the change from `moment-timezone`?

packages/containers/psammead-timestamp-container/src/index.test.jsx
```diff
@@ -63,4 +65,40 @@ describe('Timestamp', () => {
...
+          format="D MMMM YYYY"
+          isRelative
+          script={latin}
+          locale="ha"
+          service="hausa"
+        />,
+      );
+
+      const time = container.querySelector('time');
+      expect(time.textContent).toEqual('11 months wuce');
```
dr3, line 144: Is this going to break in a months time?

packages/containers/psammead-timestamp-container/src/index.jsx
DenisHdz: https://github.com/bbc/psammead/pull/2369#discussion_r335349318

packages/containers/psammead-timestamp-container/src/index.test.jsx
DenisHdz: What do you suggest to test a localized relative time then?

packages/containers/psammead-timestamp-container/package.json
DenisHdz: With this, do you mean the `moment` one?

packages/containers/psammead-timestamp-container/src/index.test.jsx
simonsinclair: Would setting `timestamp` to always be 11 months from now work?
```
moment(Date.now())
          .subtract(11, 'months')
          .valueOf();
```

simonsinclair (commented):
Nice work, @DenisHdz. This fixes the service locale issue. Please see https://github.com/bbc/psammead/pull/2369#discussion_r335469268.

packages/containers/psammead-timestamp-container/src/index.test.jsx
DenisHdz: Will try that, thanks!

packages/containers/psammead-timestamp-container/package.json
DenisHdz: Discussed IRL and agreed to add it as a `devDependency`.

packages/containers/psammead-timestamp-container/package.json
DenisHdz: Added just `moment` as a peerDep, `moment-timezone` needs to be a normal dependency.

j-pendlebury (approved):

HarveyPeachey (approved):
LGTM 👍

simonsinclair (approved):

jamesbrumpton:
lgtm

PR #2404, opened 2019-10-15
Remove input provider and with dir

### LINKED ISSUE #2371: Input Provider slots don't do anything other than get a text snippet in a chosen language
**Describe the bug**
There is a psammead helper function called input provider that adds a service dropdown in the knobs panel. This is fine but there is an argument to input provider called `slots` with not very clear documentation. After digging around the code I've discovered that `slots` are supposed to get a certain snippet of text that you can use to place in our storybook examples and then if you change the service dropdown in the knobs panel then the snippet of text will change to that service's language. For example you could ask for a `caption` or a `paragraph` or an `imageAlt` slot and then pass the text snippets into relevant components as props and the snippets will dynamically change when the service changes.

The problem is that this hasn't really taken off because for every service there is only one defined slot and it's called `text` and the logic for getting slots ultimately only ever gets a `text` slot anyway so the slot stuff is fundamentally broken. It also gets the `text` slot even if you don't ask for any slot.

I guess the slot idea was abandoned because it would be too time consuming to define different slots for every single service so only one slot was created.

I think there's 2 approaches we can take:

- Remove the slots argument and make the input provider always return the `text` slot and rename it to something clearer like `textSnippet`. Also delete the documentation around slots.
- Take the time to define different slots for every single service e.g. paragraph slot, button slot, caption slot, image alt text slot.

Whatever approach we take I think we should refactor input provider into a decorator. The reason for this is that the default service arg for input provider doesn't work and I don't think will be possible to work in it's current implementation. Also the current implementation means we have to add it to every single story which requires quite a bit of boilerplate and a non-typical way of writing stories e.g.

```jsx
stories
  .add(
    'containing an inline link',
    inputProvider({
      slots: [{ name: 'Paragraph' }, { name: 'Inline link' }],
      componentFunction: ({
        slotTexts: [paragraph, linkText],
        script,
        service,
      }) => (
        <Paragraph script={script} service={service}>
          {`${paragraph} `}
          <InlineLink href="https://www.bbc.com">{linkText}</InlineLink>
          {` ${paragraph}`}
        </Paragraph>
      ),
    }),
  );
```

with decorators we could just do

```jsx
stories
  .addDecorator(inputProvider)
  .add('containing an inline link', ({ textSnippet, script, service }) => (
    <Paragraph script={script} service={service}>
      {textSnippet}
      <InlineLink href="https://www.bbc.com">{textSnippet}</InlineLink>
      {textSnippet}
    </Paragraph>
  ));
```

or to add to all stories we do

```js
addDecorator(inputProvider)
```

jroebu14:
Resolves https://github.com/bbc/psammead/issues/2371

**Overall change:** Removes deprecated psammead test helpers `dirDecorator` and `inputProvider` in favour of `withServicesKnob` decorator.

**Code changes:**

- Replaces any use of `dirDecorator` and `inputProvider` with  `withServicesKnob`
- Removes mentions of  `dirDecorator` and `inputProvider` from `README.md`
- Fixes tests

---


thekp (commented):
@jroebu14 

looks like you have to update each package.json/package-lock/changelog for the psammead packages you've updated: https://ci.news.tools.bbc.co.uk/blue/organizations/jenkins/psammead/detail/remove-input-provider-and-with-dir/16/pipeline

😿

jroebu14:
@thekp oh yeh. so easy with the package bumping script!

chris-hinds (approved):

thekp (approved):
nice work 💯 

Do we also have issues to use `withServiceKnob` in Simorgh and fully remove instances of input-provider since it is deprecated?

packages/components/psammead-section-label/src/index.stories.jsx
```diff
@@ -1,57 +1,50 @@
...
+        dir={dir}
+        bar={boolean('show bar?', true)}
+        visuallyHidden={boolean(
+          'visually hide component for all breakpoints?',
+          false,
+        )}
+        labelId="example-section-label"
+        service={service}
+      >
+        {service === 'news' ? 'Most read' : text}
```
thekp:
```suggestion
        {service === 'news' ? 'Most Read' : text}
```

this was flagged up as a difference on chromatic: https://www.chromaticqa.com/snapshot?appId=5d28eb5ee163f6002046d6fb&id=5da87bc6d0b479002050de4d

thekp: spoke on slack, Jonathan is looking into the chromatic changes

jroebu14:
@thekp good call. no we don't. I can create these just now.

thekp (approved):

PR #2406, opened 2019-10-15
Remove padding - useful links

OlgaLyubin:
**Overall change:** 
Remove top padding from the first list item for breakpoints under 600px. This is done so that in Simorgh the distance to section label is consistent with other sections.

**Code changes:**
- Remove top padding from the first list item for breakpoints under 600px.
- Update snapshot.

---


packages/components/psammead-useful-links/src/index.jsx
```diff
@@ -34,6 +37,12 @@ export const UsefulLinksUl = styled.ul.attrs({ role: 'list' })`
   margin: 0;
   list-style-type: none;
 
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
```
DenisHdz: I think you don't need this media query. It can default to 0, and it will update once it reaches the `GEL_GROUP_3_SCREEN_WIDTH_MIN`

OlgaLyubin: Yes but in that case I will need to add 
`> li:first-child {
      padding-top: ${GEL_SPACING};
    }`
inside the `@media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN})` media query (padding does not update automatically once it reaches `GEL_GROUP_3_SCREEN_WIDTH_MIN` because `> li:first-child` takes precedence). So it's either a media query or selector+padding repetition.

OlgaLyubin: Kept the `@media query` but placed it directly inside `li` instead of accessing `li` from `ul`.

EinsteinNjoroge (approved):
LGTM 👍

DenisHdz (approved):
Looks good :)

AlistairGempf (approved):
Looks good.

jroebu14 (approved):

jamesbrumpton:
lgtm

PR #2409, opened 2019-10-15
POC - Talos without npm-check updates

pharingee:
**Overall change:** 
Replace npm-check-updates in Talos' run

**Code changes:**

---


jamesbhobbs:
@pharingee what's wrong with continuing to use npm-check-updates ?

pharingee:
@benjaminhobbs  I had a conversation [here](https://github.com/bbc/psammead/pull/2244) that outlines my points, and to summarise:

- `npm-check-updates` by virtue of the fact that it needs to query the npm database with 30+ network calls is prone to hanging.
- This means more resources are used and when it doesn't hang it takes a long time to complete.
- All the information we need to do what `npm-check-updates` does is right in the codebase so I see removing it as the next logical step.


In addition to these:
- We need to move from terminal based executions if we can because they are based on environments which are different from computer to computer. By moving to api based calls, we can effectively make the codebase more testable and reliable.


I feel quite strongly about this, however there's a point I'm willing to push this for. I just need one more opinion and I do not mind going back to using `npm-check-updates`

jamesbhobbs:
@pharingee a middle ground would be to use `npm outdated` then those which are outdated you can `ncu [package-names]` this cuts network requests on ncu and I hope the primary npm command is not making separate requests just to check if outdated.

amywalkerdev:
Out of scope for Talos

Issue #2126, opened 2019-09-13
async/await in components


jroebu14:
**Is your feature request related to a problem? Please describe.**
Async/await does not work in psammead components. This is because the regenerator runtime plugin is not included in the babel config. When it is included it causes problems in IE11.

**Describe the solution you'd like**
To be able to use async/await

**Describe alternatives you've considered**
Use promises instead.

**Testing notes**
[Tester to complete]

Dev insight: Will Cypress tests be required or are unit tests sufficient? Will there be any potential regression? etc

- [ ] This feature is expected to need manual testing.

**Additional context**

PR #2416, opened 2019-10-16
Psammead test helpers match snapshot async

pjlee11:
Arises from https://github.com/bbc/simorgh/pull/4240/

As a POC the logic in this PR was added [here](https://github.com/bbc/simorgh/blob/dd379421fb5b3f12c57d82b0633b2fc50724b77b/src/app/containers/CpsAssetPageMain/testHelpers.js#L1) to prove we could have snapshot tests generated from async it() blocks. 

**Code changes:**

- Creates `createSnapshot` as a refactor to share the snapshot generating code. 
- Adds `matchSnapshotAsync` which is to be used inside of an `async` it() block. This is required because the execution of the async it must be in Simorgh or the describe block is not able to await the test being run. 

---


packages/utilities/psammead-test-helpers/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 3.0.3 | [PR#xxxx](https://github.com/bbc/psammead/pull/xxxx) Adds matchSnapshotAsync export |
```
pjlee11:
```suggestion
| 3.0.3 | [PR#2416](https://github.com/bbc/psammead/pull/2416) Adds matchSnapshotAsync export |
```

packages/utilities/psammead-test-helpers/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-test-helpers",
-  "version": "3.0.2",
+  "version": "3.0.3",
```
dr3:
```suggestion
  "version": "3.1.0",
```

dr3 (changes requested):
LGTM, just version tweak

dr3 (approved):

rhenshaw56 (approved):
Looks good! most of the lines aren't covered though

packages/utilities/psammead-test-helpers/index.test.jsx
```diff
@@ -232,4 +233,14 @@ describe('Psammead test helpers', () => {
     'should match the snapshot for the test component with helmet and other content',
     <HelmetWithContent />,
   );
+
+  it('should create a snapshot from an async it() block with component and done props', async done => {
+    const data = Promise.resolve('Foobar');
```
jroebu14, line 238: I think this should be `const data = await Promise.resolve('Foobar');`

packages/utilities/psammead-test-helpers/index.test.jsx
```diff
@@ -232,4 +233,14 @@ describe('Psammead test helpers', () => {
     'should match the snapshot for the test component with helmet and other content',
     <HelmetWithContent />,
   );
+
+  it('should create a snapshot from an async it() block with component and done props', async done => {
```
jroebu14, line 237: you don't need to call done when using async/await https://jestjs.io/docs/en/tutorial-async#async-await

jroebu14: I didn't think it was possible to use async/await in Psammead https://github.com/bbc/psammead/issues/2126

I haven't tried this release yet but I suspect this package is broken if the above issue hasn't been sorted

pjlee11: As the async/await is only in the tests it shouldn't be affected by #2126

PR #2421, opened 2019-10-17
Remove done from matchSnapshotAsync

pjlee11:
Follows up https://github.com/bbc/psammead/pull/2416 with changes based on the comments in https://github.com/bbc/simorgh/pull/4240#discussion_r335853599

**Overall change:** Removes `done` from `matchSnapshotAsync` and explicitly returns the contents of the function. 

**Code changes:**

- Splits out `createSnapshot` as it isn't reuse-able if `done` removed, also the `matchSnapshotAsnyc` has to be explicitly returned
- Tidies up the `matchSnapshotAsnyc` test to be more readable

---

### DIFF
--- a/packages/utilities/psammead-test-helpers/index.test.jsx
+++ b/packages/utilities/psammead-test-helpers/index.test.jsx
@@ -234,13 +234,12 @@ describe('Psammead test helpers', () => {
     <HelmetWithContent />,
   );
 
-  it('should create a snapshot from an async it() block with component and done props', async done => {
-    const data = Promise.resolve('Foobar');
-    const component = (
+  it('should create a snapshot from an async it() block', async () => {
+    const data = await Promise.resolve('Foobar');
+    await testHelpers.matchSnapshotAsync(
       <div>
-        <p>{await data}</p>
-      </div>
+        <p>{data}</p>
+      </div>,
     );
-    testHelpers.matchSnapshotAsync(component, done);
   });
 });
--- a/packages/utilities/psammead-test-helpers/src/index.js
+++ b/packages/utilities/psammead-test-helpers/src/index.js
@@ -6,18 +6,22 @@ import renderWithHelmet from './renderWithHelmet';
 // select the first child to remove the pointless wrapping div from snapshots
 const removeWrappingDiv = container => container.firstChild;
 
+const createSnapshot = container => {
+  const hasOneChild = container.children.length === 1;
+  /*
+   * if the container has more than one child then it's a component that uses a
+   * fragment at the top level so we should not select the first child because it
+   * wouldn't snapshot the whole component
+   */
+  expect(
+    hasOneChild ? removeWrappingDiv(container) : container,
+  ).toMatchSnapshot();
+};
+
 export const shouldMatchSnapshot = (title, component) => {
   it(title, done => {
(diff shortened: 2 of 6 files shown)


packages/utilities/psammead-test-helpers/src/index.js
```diff
@@ -3,10 +3,28 @@ import 'jest-styled-components';
...
-  // select the first child to remove the pointless wrapping div from snapshots
-  const removeWrappingDiv = container => container.firstChild;
-  renderWithHelmet(component).then(({ container }) => {
+// select the first child to remove the pointless wrapping div from snapshots
+const removeWrappingDiv = container => container.firstChild;
+
+export const shouldMatchSnapshot = (title, component) => {
+  it(title, done => {
+    renderWithHelmet(component).then(({ container }) => {
+      const hasOneChild = container.children.length === 1;
```
pjlee11: Struggling to find a nice way to reuse
```
    const hasOneChild = container.children.length === 1;
    /*
     * if the container has more than one child then it's a component that uses a
     * fragment at the top level so we should not select the first child because it
     * wouldn't snapshot the whole component
     */
    expect(
      hasOneChild ? removeWrappingDiv(container) : container,
    ).toMatchSnapshot();
```

Because `matchSnapshotAsync` needs to return `renderWithHelmet`, whereas `shouldMatchSnapshot` doesn't need to return it and also needs done(); at the end of the `renderWithHelmet` function.

pjlee11: Actually I'm just being slow ... commit incoming

jroebu14 (approved):

ryanmccombe (approved):

packages/utilities/psammead-test-helpers/index.test.jsx
```diff
@@ -234,13 +234,12 @@ describe('Psammead test helpers', () => {
...
-    const data = Promise.resolve('Foobar');
-    const component = (
+  it('should create a snapshot from an async it() block', async () => {
+    const data = await Promise.resolve('Foobar');
+    await testHelpers.matchSnapshotAsync(
       <div>
-        <p>{await data}</p>
-      </div>
+        <p>{data}</p>
+      </div>,
```
chris-hinds, line 242: what is the comma used here for?

ibMadbouly (approved):
👍

packages/utilities/psammead-test-helpers/index.test.jsx
pjlee11: Prettier is taking `<div><p>{data}</p></div>` as the first param in the function call, and because it's multi-line prettier adds a comment for future params to be passed. 

Alternatively it could be:
```
const component = (
  <div>
    <p>{data}</
  </div>
);
await testHelpers.matchSnapshotAsync(component);
```

pjlee11: Also I agree the comma currently looks gross and confusing :trollface:

chris-hinds: happy with how it is, just wanted to check. Thought it might be a linting thing

chris-hinds (approved):

Issue #2384, opened 2019-10-14
Accessibility swarm on the Bulletin component


DenisHdz:
**Is your feature request related to a problem? Please describe.**
Carry out a `two devs` accessibility swarm for the bulletin component once it is completed.

This could be done when integrating the alpha component locally in Simorgh.

**Describe the solution you'd like**
Go through the [accessibility checklist](https://bbc-news.github.io/accessibility-news-and-you/accessibility-news-and-testers) and fix any issue we might face.

**Describe alternatives you've considered**

**Testing notes**
[Tester to complete]

Dev insight: Will there be any potential regression? etc

- [ ] This feature is expected to need manual testing.

**Additional context**

DenisHdz:
Blocked on https://github.com/bbc/psammead/issues/2382

DenisHdz:
Block on https://github.com/bbc/psammead/issues/2622

ghost:
❓ = Not done
✅ = Done, pass
❌ = Done, fail
👀 = Review with UX

## Checklist - Manual ##

  - ### Design  ✅

  - ### Color contrast   ✅

  - ### Color blindness  ✅

  - ### High Contrast mode PC ✅ 

![radio-hc](https://user-images.githubusercontent.com/46446236/69972732-0a0a4680-151a-11ea-8222-a06f6b9ed796.jpg)
![tv-hc](https://user-images.githubusercontent.com/46446236/69972763-1c848000-151a-11ea-9e4c-41d60e7d646e.jpg)



  - ### Color preferences changed in Firefox   ✅

<img width="1009" alt="Screenshot 2019-11-29 at 11 31 12" src="https://user-images.githubusercontent.com/48688870/70021729-c14aa000-15a2-11ea-83ba-ca4cbdb5a48a.png">



  - ### Text resized to 200%  ✅

  - ### Page zoomed to 200%  ✅

  - ### Heading order ✅

  - ### Icons  ✅
   
  - ### Keyboard navigation: tab order, focus css, trace tab path  ✅

  - ### aXe ✅

  - ### No CSS  ✅

  - ### No Javascript  ✅

  - ### bbc-a11y ✅

### Update
After discussion with UX (Nick) we have decided to add a transparent border around the `CTA` for now and fix forward if we come up with a better solution for the `high contrast mode`

ghost:
What is the expected trace tab path for keyboard accessibility?
### 1
[...]

PR #2435, opened 2019-10-18
Create Bulletin component

### LINKED ISSUE #2382: Create Bulletin component
Create Bulletin component.

Build a new Bulletin alpha package that exports two components, Radio bulletin and TV bulletin, matching the UX designs. These will be displayed just under 600px for now.

The Radio and TV can use common `styled-components` and they can accept props in order to hide the image, display the right media indicator and tweak the button colors.

Make sure about not adding paddings or margins around the components. This will be done in Simorgh if necessary.

The link to designs can be found on WSFRONTPGS-86 and WSFRONTPGS-227.

tv-radio

N/A

**Testing notes**
[Tester to complete]

**Additional context**

AlistairGempf:
Resolves #2382

**Overall change:** Creates the new Radio and TV Bulletins to be displayed below 600px.

**Code changes:**

- Creates new alpha components for Radio and TV Bulletins

---


packages/components/psammead-bulletin/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+# ⛔️ This is an alpha component ⛔️
```
j-pendlebury, line 1: should be moved to the top

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,277 @@
+import React from 'react';
+import styled from 'styled-components';
+import { GEL_GROUP_CD_MIN_WIDTH } from '@bbc/gel-foundations/breakpoints';
```
j-pendlebury: I believe that `GEL_GROUP_CD_MIN_WIDTH` should be used for typography settings (based on the comments in `gel-foundations/src/breakpoints.js`, so we should be using `GEL_GROUP_3_SCREEN_WIDTH_MIN` here. This also means that we don't have to add `rem` on line 25

packages/components/psammead-bulletin/src/index.stories.jsx
```diff
@@ -0,0 +1,57 @@
...
+import notes from '../README.md';
+import { RadioBulletin, TVBulletin, BulletinImage } from './index';
+
+storiesOf('Components|Bulletin', module)
+  .addDecorator(withKnobs)
+  .addDecorator(withServicesKnob())
+  .add(
+    'Audio',
+    ({ text, script, service, dir }) => {
+      const liveIndicator = boolean('Live', false) ? 'Live ' : null;
```
j-pendlebury: could this be replaced with `&&`? 
```suggestion
      const liveIndicator = boolean('Live', false) && 'Live ';
```

packages/components/psammead-bulletin/README.md
```diff
@@ -0,0 +1,113 @@
...
+<!-- prettier-ignore -->
+| Argument  | Type | Required | Default | Example |
+| --------- | ---- | -------- | ------- | ------- |
+| `title` | string | yes  | - | `'TV: Example TV show'` |
+| `liveText` | string | no | `null` | `'Live '` |
+| `dir` | `'ltr' || 'rtl'` | yes | - | `'ltr'` |
+| `script` | scriptObject | yes | - | `{ canon: { groupA: { fontSize: '28', lineHeight: '32',}, groupB: { fontSize: '32', lineHeight: '36', }, groupD: { fontSize: '44', lineHeight: '48', }, }, trafalgar: { groupA: { fontSize: '20', lineHeight: '24', }, groupB: { fontSize: '24', lineHeight: '28', }, groupD: { fontSize: '32', lineHeight: '36', }, }, }` |
+| `service` | string | yes | - | `'pidgin'` |
+| `link` | url | yes | - | `'https://www.bbc.com/telugu/bbc_telugu_tv/tv_programmes/w13xttld'` |
+| `CTAText` | string | yes | - | `'Watch'` |
```
j-pendlebury: if we're following camel case for prop names then this should be `ctaText` 🙈

j-pendlebury (changes requested):
Couple of comments from me 🙊

j-pendlebury:
I'm removing the `a11y-swarm` and `ux` labels since those checks can be done once this component is merged into Psammead and can be viewed in Storybook, as per the new alpha component guidelines.

~~Will create an issue to make sure a11y swarm is completed~~
EDIT: issue already exists https://github.com/bbc/psammead/issues/2384

packages/components/psammead-bulletin/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@bbc/psammead-bulletin",
+  "version": "0.1.0-alpha.0",
```
OlgaLyubin: Should be "version": "0.1.0-alpha.1"

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,277 @@
...
+  link,
+  CTAText,
+  hiddenTitleText,
+  dir,
+}) => (
+  <BulletinBodyWrapper dir={dir}>
+    <StyledHeading script={script} service={service}>
+      <span>
+        <VisuallyHiddenText>{hiddenTitleText}</VisuallyHiddenText>
+        {liveText && <LiveIndicator text={liveText} />}
```
tochwill: I’m not sure about tying the state of ‘isLive’ to whether or not the translation copy is being passed in, would it make more sense to always pass in copy that's coming from translation bundles and have a separate prop to toggle the live state?

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,277 @@
...
+  script,
+  service,
+  link,
+  CTAText,
+  hiddenTitleText,
+  dir,
+}) => (
+  <BulletinBodyWrapper dir={dir}>
+    <StyledHeading script={script} service={service}>
+      <span>
```
tochwill: What is this `span` needed for?

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,277 @@
...
+  width: 100%;
+  display: block;
+  text-align: center;
+  padding: ${GEL_SPACING};
+  vertical-align: middle;
+`;
+
+const StyledSummary = styled.p`
+  ${({ script }) => script && getLongPrimer(script)};
+  ${({ service }) => getSansRegular(service)}
```
OlgaLyubin: We should check if `service` exists before calling it, same way we do with `script`. `${({ service }) => service && getSansRegular(service)}`

packages/components/psammead-bulletin/README.md
```diff
@@ -0,0 +1,113 @@
...
+| `hiddenTitleText` | string | yes | - | `'Watch, '` |
+
+### TV Bulletin
+
+<!-- prettier-ignore -->
+| Argument  | Type | Required | Default | Example |
+| --------- | ---- | -------- | ------- | ------- |
+| `title` | string | yes  | - | `'TV: Example TV show'` |
+| `liveText` | string | no | `null` | `'Live ` |
```
OlgaLyubin: Missing a closing '    | `'Live' `|

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,277 @@
...
+        {liveText && <LiveIndicator text={liveText} />}
+        <HeadingLink href={link} dir={dir}>
+          {title}
+        </HeadingLink>
+      </span>
+    </StyledHeading>
+    <StyledSummary script={script} service={service}>
+      {summary}
+    </StyledSummary>
+    <PlayCTA type={type} isLive={liveText && true} dir={dir}>
```
OlgaLyubin: What is the reason for adding `&& true` here?

packages/components/psammead-bulletin/src/index.jsx
AlistairGempf: Ah yes, I wondered about this too. The reason I went for this is that the text in the `liveText` prop gets styled differently to the rest of the title and if the prop is live then the colour of the call to action also changes. If there were a separate prop `isLive` and liveText were always shown we could have a non-live CTA with `liveText` displayed or vice-versa which seems like it shouldn't be possible to me. What are your thoughts about that case?

packages/components/psammead-bulletin/src/index.jsx
AlistairGempf: There isn't one. Not even sure why I put it in in the first place :D

packages/components/psammead-bulletin/README.md
AlistairGempf: Good spot

packages/components/psammead-bulletin/src/index.jsx
tochwill: I agree with your reasoning! I would keep the logic exactly the same, but where you're checking for `liveText`, check instead to see if the `isLive` prop is true. Then you can just pass `liveText` in as a translation string (rather than setting it to `const liveText = isLive ? 'Live ' : null;` outside of the component). So you'd keep the condition `{isLive && <LiveIndicator text={liveText} />}`, and only render the LiveIndicator when `isLive` is true.

packages/components/psammead-bulletin/src/index.jsx
AlistairGempf: This was added while I was trying some things out, and I forgot to remove it. It can be removed now.

packages/components/psammead-bulletin/src/index.stories.jsx
```diff
@@ -0,0 +1,59 @@
...
+          link="foo/bar"
+          ctaText="Listen"
+          hiddenTitleText="Listen, "
+        />
+      );
+    },
+    { notes },
+  )
+  .add(
+    'Video',
```
ghost:
```suggestion
    'TV',
```

packages/components/psammead-bulletin/src/index.stories.jsx
```diff
@@ -0,0 +1,59 @@
...
+import { storiesOf } from '@storybook/react';
+import { withServicesKnob } from '@bbc/psammead-storybook-helpers';
+import notes from '../README.md';
+import { RadioBulletin, TVBulletin, BulletinImage } from './index';
+
+storiesOf('Components|Bulletin', module)
+  .addDecorator(withKnobs)
+  .addDecorator(withServicesKnob())
+  .add(
+    'Audio',
```
ghost:
```suggestion
    'Radio',
```

ghost (dismissed):
Storybook doesn't render anything and has errors.
Screenshot 2019-10-22 at 15 34 21

packages/components/psammead-bulletin/src/index.test.jsx
```diff
@@ -0,0 +1,241 @@
...
+        link="foo"
+        ctaText="Listen"
+        hiddenTitleText="Listen, "
+      />,
+    );
+
+    expect(container.querySelector('h3').textContent).toEqual(
+      'Listen, Hello World',
+    );
+  });
```
ghost: We should wrap this in a `describe('assertions', () => {`

packages/components/psammead-bulletin/src/index.jsx
AlistairGempf: Gotcha! Sounds like a sensible idea to me so have implemented it.

AlistairGempf:
> Storybook doesn't render anything and has errors.

@sadickisaac Interesting, ~I don't get any errors and I don't see anything that would cause that error in the code~ (I've found it now and looking into it). It will only render when the screen-width is below 600px otherwise it shows nothing. Is that why you aren't seeing anything?

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,285 @@
...
+RadioBulletin.defaultProps = {
+  liveText: null,
+  isLive: false,
+};
+
+export const BulletinImage = ({ src, alt }) => (
+  <BulletinImageWrapper>
+    <StyledBulletinImage src={src} alt={alt} />
+  </BulletinImageWrapper>
+);
```
AlistairGempf: Based on looking at the `StoryPromo` component in Simorgh, this may have to change before actually being used to work in a similar way so that amp pages work properly.

OlgaLyubin:
Screenshot 2019-10-23 at 09 24 36

Watch/Listen label on the button is not displayed with the font style like in the designs

OlgaLyubin:
Screenshot 2019-10-23 at 09 33 18

I̶n̶ ̶t̶h̶e̶ ̶d̶e̶s̶i̶g̶n̶s̶ ̶w̶h̶e̶n̶ ̶R̶a̶d̶i̶o̶ ̶i̶s̶ ̶n̶o̶t̶ ̶l̶i̶v̶e̶,̶ ̶t̶h̶e̶ ̶h̶e̶a̶d̶i̶n̶g̶ ̶h̶a̶s̶ ̶a̶ ̶g̶r̶e̶y̶ ̶c̶o̶l̶o̶u̶r̶ ̶i̶n̶s̶t̶e̶a̶d̶ ̶o̶f̶ ̶b̶l̶a̶c̶k̶ My bad, everything is correct, didn't notice the explanation in the markup design :)

AlistairGempf:
> In the designs when Radio is not live, the heading has a grey colour instead of black

@OlgaLyubin as per the markup design 'the Visited state is the same as standard promo’s e.g headline / H3 changes to #6E6E73 (metal)' - so it's black until it's visited then switches to the grey.

OlgaLyubin:
Do you have any specifications about what should be clickable? Right now clicking on anything on the bulletin (pic, heading, button, summary) redirects me to the url. Is this the way it's supposed to be?

AlistairGempf:
> Do you have any specifications about what should be clickable? Right now clicking on anything on the bulletin (pic, heading, button, summary) redirects me to the url. Is this the way it's supposed to be?

Not specifically, however the designs refer to similarities to standard story promos in several places, so I have based it on that. This also makes sense to allow the biggest touch target possible and makes the site consistent.

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,293 @@
...
+
+const PlayCTA = styled.div.attrs({ 'aria-hidden': true })`
+  background-color: ${({ isLive }) => (isLive ? C_POSTBOX : C_EBON)};
+  ${({ service }) => service && getSansRegular(service)};
+  ${({ script }) => script && getPica(script)};
+  color: ${C_WHITE};
+  width: 100%;
+  display: block;
+  text-align: center;
+  padding: ${GEL_SPACING};
```
OlgaLyubin: Padding for the button should be 12px not 8px

packages/components/psammead-bulletin/src/index.stories.jsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { withKnobs, boolean } from '@storybook/addon-knobs';
+import { storiesOf } from '@storybook/react';
+import { withServicesKnob } from '@bbc/psammead-storybook-helpers';
+import notes from '../README.md';
+import { RadioBulletin, TVBulletin, BulletinImage } from './index';
+
```
DenisHdz, line 35: You could add a black container to the stories, so we could see the white bg applied the `BulletinWrapper`.

```
const Container = styled.div`
  background-color: black;
  padding: 1rem;
  height: 100vh;
`;

<Container>
 <Radiobulletin ... />
</Container>

<Container>
 <TVBulletin ... />
</Container>
```

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,295 @@
...
+  liveText,
+  dir,
+  script,
+  service,
+  link,
+  ctaText,
+  hiddenTitleText,
+  isLive,
+}) => (
+  <BulletinWrapper dir={dir}>
```
DenisHdz: Do we need to pass the `dir` prop to the `BulletinWrapper` and `BulletinBodyWrapper`?

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,295 @@
...
+import { mediaIcons } from '@bbc/psammead-assets/svgs';
+import VisuallyHiddenText from '@bbc/psammead-visually-hidden-text';
+
+const GEL_SPACING_THREE_QUARTERS = '0.75rem';
+
+const BulletinWrapper = styled.div`
+  background-color: ${C_WHITE};
+  padding: ${GEL_SPACING};
+  position: relative;
+  text-decoration: none;
```
DenisHdz: Do we need this?

packages/components/psammead-bulletin/src/index.jsx
AlistairGempf: Doesn't look like it. Removed.

packages/components/psammead-bulletin/src/index.jsx
AlistairGempf: Doesn't look like it. Removed.

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,295 @@
...
+`;
+
+const StyledHeading = styled.h3`
+  ${({ script }) => script && getPica(script)}
+  ${({ service }) => getSansBold(service)}
+  color: ${C_EBON};
+  margin: 0; /* Reset */
+`;
+
+const HeadingLink = styled.a`
```
DenisHdz: It looks like this `HeadingLink` has the same styles as the `Story Promo Link`. Do you think it would be better to import the link and use it here instead?

DenisHdz: Also, in the Story Promo, we did it in a different way. We export all the elements and then we put them together in the stories and in the Simorgh container, I wonder what is the best approach.

packages/components/psammead-bulletin/src/index.stories.jsx
AlistairGempf: Good idea, have added this.

packages/components/psammead-bulletin/src/index.jsx
AlistairGempf: Based on discussion in slack, as `RadioBulletin` is mapped over data, the dynamic construction will be done in Simorgh.

AlistairGempf: Ok, I've now created a `psammead-bulletin-container` which extracts the relevant data and composes the `Bulletin` into the form expected in its `TVBulletin` and `RadioBulletin` containers. This should be easy to expand into a `ul` with `li` items in it when we get to that stage, but I don't think that's this task.

packages/components/psammead-bulletin/src/index.stories.jsx
```diff
@@ -0,0 +1,28 @@
...
+import { storiesOf } from '@storybook/react';
+import { withServicesKnob } from '@bbc/psammead-storybook-helpers';
+import notes from '../README.md';
+import { BulletinCTA } from './index';
+
+storiesOf('Components|Bulletin', module)
+  .addDecorator(withKnobs)
+  .addDecorator(withServicesKnob())
+  .add(
+    'Call to Action',
```
AlistairGempf: I wasn't sure whether to leave this story in to help with further development on the CTA or to just use the full container's story. Would be good to hear what others think.

ghost:
@AlistairGempf Great work on the Bulletin Component and Container. One suggestion would be to move `Bulletin` container to Simorgh.

packages/components/psammead-bulletin/src/index.test.jsx
```diff
@@ -0,0 +1,97 @@
...
+const BulletinComponent = ({ script, service, isLive, dir, type, ctaText }) => {
+  const summaryText = 'This is the summary text';
+  const headlineText = 'This is the headline';
+  const imgSrc =
+    'https://ichef.bbci.co.uk/news/660/cpsprodpb/11897/production/_106613817_999_al_.jpg';
+  const bulletinLink = 'https://bbc.co.uk';
+  const imgAltText = 'Iron man';
+
+  return (
+    <Bulletin>
```
DenisHdz: After having a discussion with @sareh and @tochwill, we think that in this particular situation, since the Bulletin is going to have a fixed static layout, it is better we compose it in the `index.jsx` (e.g `BulletinPromo`), so we can easily import it in the stories and later on in Simorgh. Basically what @AlistairGempf initially did.

Also, it would be nice to have a couple of stories showing the `PlayCTA` component for audio and video and when it is live or not.

packages/components/psammead-bulletin/src/index.stories.jsx
```diff
@@ -0,0 +1,84 @@
...
+  const imgAltText = 'Iron man';
+
+  return (
+    <Bulletin>
+      {type === 'video' && <Image src={imgSrc} alt={imgAltText} />}
+      <BulletinHeading script={script} service={service}>
+        <VisuallyHiddenText>{ctaText}, </VisuallyHiddenText>
+        {isLive && (
+          <LiveLabel service={service} dir={dir}>
+            Live
```
tochwill: I'm a little concerned by the hardcoded 'Live' text in english - it's in a couple of places in the codebase and I'd have assumed it should be part of a translation bundle and passed in?

ghost: Yes it should be a part of translation. This would be passed in from the simorgh configs.

packages/components/psammead-bulletin/src/index.test.jsx
ghost: If it will be static it makes sense to compose it in `index.js`.
> Also, it would be nice to have a couple of stories showing the PlayCTA component for audio and video and when it is live or not.

I don't think this is necessary, we are already showing it as part of the `Bulletin`

Screenshot 2019-10-30 at 21 41 33

DenisHdz: > I don't think this is necessary, we are already showing it as part of the Bulletin

I am happy either way 👍

packages/components/psammead-bulletin/src/index.stories.jsx
DenisHdz: If that is the case, should we pass the `live` string also as a prop? At the moment is hardcoded in English, and it should be in uppercase.

DenisHdz (approved):
Apart from [this](https://github.com/bbc/psammead/pull/2435#discussion_r341251680), it looks good to me 👍🏼

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -0,0 +1,110 @@
...
+  image: node,
+  summaryText: string.isRequired,
+  headlineText: string.isRequired,
+  liveText: string,
+};
+
+Bulletin.defaultProps = {
+  isLive: false,
+  image: null,
+  liveText: 'Live',
```
DenisHdz:
```suggestion
  liveText: 'LIVE',
```

OlgaLyubin:
Snapshots should be updated too since `liveText: 'Live',` is now changed to `liveText: 'LIVE',`

OlgaLyubin (approved):
Looks great 👍🏼

j-pendlebury (approved):
LGTM 👍

Bopchy (approved):

PriyaKR:
Testing:

@AlistairGempf few UX issues:

- Font size of the summary text in TV and radio bulletin is 14px/16px according to the designs but actual its same as the headline text.
- Hover over state change on the watch and listen CTA is missing
- Also should the hover over state change text for live Radio bulletin be `Listen live` as in the designs?

AlistairGempf:
> Font size of the summary text in TV and radio bulletin is 14px/16px according to the designs but actual its same as the headline text.

Thanks, I'll change this.

> Hover over state change on the watch and listen CTA is missing

There is a comment on the design to not do this (Note 12 on the marked up version).

> Also should the hover over state change text for live Radio bulletin be `Listen live` as in the designs?

This will be passed in in the `ctaText` prop. I'll change the story to match the design though.

DenisHdz:
With that last change about adding `Live` to the ctaText, we will be repeating the word `Live` twice, as it will be in the `VisuallyHiddenText` and also in the `LiveLabel`.

AlistairGempf:
> With that last change about adding `Live` to the ctaText, we will be repeating the word `Live` twice, as it will be in the `VisuallyHiddenText` and also in the `LiveLabel`.

Good point @DenisHdz. Do you have any recommendations? Maybe add an aria-hidden attribute to the `LiveLabel` in this component?

DenisHdz:
> > With that last change about adding `Live` to the ctaText, we will be repeating the word `Live` twice, as it will be in the `VisuallyHiddenText` and also in the `LiveLabel`.
> 
> Good point @DenisHdz. Do you have any recommendations? Maybe add an aria-hidden attribute to the `LiveLabel` in this component?

Or you could use the `liveText` in the button (assuming we also need it translated):

```
 <PlayCTA isLive={isLive} service={service} script={script}>
     <IconWrapper>{mediaIcons[type.toLowerCase()]}</IconWrapper>
     {isLive && `${ctaText} ${liveText}`}
 </PlayCTA>
```

AlistairGempf:
The `LiveLabel` imported from `StoryPromo` is already Aria hidden, so passing in the `ctaText` as 'Listen Live' will have the expected behaviour on all fronts.

Also note that UX has iterated the design of the Radio Bulletin a little bit, which I have now implemented.

DenisHdz (approved):
Great work @AlistairGempf and @sadickisaac 👍

paruchurisilpa:
Looks good to me...

Issue #2319, opened 2019-10-07
Grid component: gutters and margins fallback


DenisHdz:
**Is your feature request related to a problem? Please describe.**
Currently, the fallback for browsers that don't support CSS Grid is that there are no gutters or margins applied. 

<img width="1046" alt="ie-gutters" src="https://user-images.githubusercontent.com/46446236/66476135-c3095300-ea8c-11e9-96f5-0c09f7b2ad34.png">

**Describe the solution you'd like**
Implement `gutters` and `margins` to work on IE.

**Describe alternatives you've considered**

**Testing notes**
- Ensure it works across supported browsers

Dev insight: Will there be any potential regression? etc

- [X] This feature is expected to need manual testing.

**Additional context**

DenisHdz:
Close as done in [this](https://github.com/bbc/psammead/pull/2439) PR

PR #2439, opened 2019-10-21
Grid compoment - Add start offset, gutters and margins fallbacks

### LINKED ISSUE #2285: Grid component: no start offset fallback for IE
**Describe the bug**

**To Reproduce**

**Expected behaviour**

**Screenshots**
Screenshot 2019-10-01 at 16 00 55

**Testing notes**

**Additional context**

DenisHdz:
Resolves #2285 and #2319

**Overall change:**

Add fallbacks to the Grid component to support start offset, gutters and margins for browsers that don't support CSS Grid.

**Code changes:**
- Create a `Wrapper` styled component to add a wrapper with padding when CSS Grid is not supported and `enableGelMargins` is passed

- Create `gridOffsetFallback` function to calculate the percentage the element should move from the start as the `grid-column-start` would do. This value will be used to apply a `margin-left` to the element and, if the child has the same number of cols than the parent, to also deduct it from its width.

- Update `gridFallbacks` to apply right fallback styles to the first outer Grid element or children.
   - Create `outerGridFallback`:
      - If `enableGutters` prop is passed: apply a **negative** `margin-left and margin-right` that compensate with the `margin-left` and `margin-right` that later the children will have.
       - If `gridStartOfsset` prop is passed: apply a `margin-left` to move the element from the start.
   - Create `childrenFallback`:
     - If Grid item and `enableGutters` prop is passed to the parent:
             - Apply a `margin-left and margin-right` to thee Grid item
             - Calculate fallback width (minus de `gutterSize` minus the `startOffset` if the grid item columns are the same as the parent)
      - If Grid element
             - Calculate fallback width (minus the `startOffset` if the grid element columns are the same as the parent)
          
- Add support for `rtl`
---


EinsteinCarrey:
@DenisHdz there's a small difference between how `Grid Article image example` appears on IE vs chrome.
Screen Shot 2019-10-22 at 12 32 54 PM
Screen Shot 2019-10-22 at 12 32 34 PM

EinsteinNjoroge (commented):
@DenisHdz there's a small difference between how `Grid Article image example` appears on IE vs chrome.
Screen Shot 2019-10-22 at 12 32 54 PM
Screen Shot 2019-10-22 at 12 32 34 PM

packages/components/psammead-grid/src/index.jsx
```diff
@@ -110,31 +112,142 @@ const gridMediaQueries = ({
...
+        `
+      : `
+        width: calc(${(100 * columnsGroup) /
+          parentColumnsGroup}%${negativeOffset(
+          columnsGroup,
+          parentColumnsGroup,
+          gridStartOffset,
+          gridStartOffsetGroup,
+        )});
+        `
```
ghost: Can we refactor this and  have a variable set to the results of 👇 
```js
negativeOffset(
  columnsGroup,
  parentColumnsGroup,
  gridStartOffset,
  gridStartOffsetGroup,
)})
```

ghost (changes requested):

packages/components/psammead-grid/src/index.jsx
DenisHdz: Done!

DenisHdz:
> @DenisHdz there's a small difference between how `Grid Article image example` appears on IE vs chrome.
> Screen Shot 2019-10-22 at 12 32 54 PM
> Screen Shot 2019-10-22 at 12 32 34 PM

In fact, when passing that `startOfsset` object to the outer grid element, we should expect it to be displayed as you see it in IE. 

@sareh Do you remember why that offset is not working properly when grid is supported?

EinsteinNjoroge (approved):
I checked this on browserstack, looks Good 👍

sareh (approved):
👍 Looks great. 

The outstanding example with `startOffset` for Articles now works in the fallback as expected. 🎉 

See this on IE11 on Browserstack: 
Screenshot IE11

And this on the latest Chrome: 
Screenshot Chrome
Screenshot 2019-10-30 at 10 29 17

ghost (approved):

jamesbrumpton:
lgtm

Issue #581, opened 2019-06-05
Create psammead-most-popular list component


andrew-nowak:
**Is your feature request related to a problem? Please describe.**
Create List and ListItem components to position items in Most Read. 

Columns:
<400px: a single column
400px-1007px: two columns
1280px+: 5 columns

Screenshot of view at a specific breakpoint that has two columns:
<img width="1045" alt="Screenshot of the component" src="https://user-images.githubusercontent.com/3028997/63435373-b66a6600-c41e-11e9-9c69-8da06e7035ea.png">

UX A11y criteria are here: https://app.zeplin.io/project/5baa564c05c2eba0d2b89568/screen/5cee7072c6c48d1d928c5335

Acceptance criteria

Most Read - Ordered list

    Given I use a screen reader, including VoiceOver  (Note we are particularly checking VoiceOver as there is a bug with this screen reader)
    And I am viewing Most Read
    When navigating to the ordered list
    Then it should be announced as a list

Most Read - List item

    Given I use a screen reader
    And I am viewing Most Read
    When navigating to a list item in the ordered list
    Then the text should be announced
    And it should be prefixed by the announcement of the list item number

    Given I am viewing Most Read
    When the list items are displayed in 2 columns or more
    Then each column must read from left to right

    Given I am viewing a a Persian or other RTL page
    and I am viewing Most Read
    When the list items are displayed in 2 columns or more
    Then each column must read from right to left

    Given I am viewing Most Read
    When the list items are in columns
    Then they must be within one ordered list
    The list item number must meet the colour contrast ratio of 4.1:5

**Describe the solution you'd like**
Export styled `<ol>` and `<li>` components, which position items into the most read component. Use of css grid requires a fallback. We should use the `@bbc/psammead-grid` for the grid layout.

<400px: a single column
400px-1007px: two columns
1280px+: 5 columns

**Describe alternatives you've considered**
N/A

**Testing notes**
[Tester to complete]

Dev insight: Snapshot tests, plus storybook stories, plus a dev a11y swarm (2 devs) should be sufficient. Full a11y test will be carried out at the end. 

**Additional context**
N/A

sareh:
~Blocked on Item being created #579~

Issue #1429, opened 2019-07-23
Integrate links, list & title within Most Read component


sareh:
**Is your feature request related to a problem? Please describe.**
Within the Psammead component, ensure the `@bbc/psammead-most-read` main index file exports the whole Most Read component. 

As a result, you can use the Most Read component in any page:

```
import MostRead from '@bbc/psammead-most-read';

... return (<MostRead dir={dir} header={header} items={items} lastUpdated={lastUpdated} script={script} servie={service} />)
```
Noting that `header` & `lastUpdated` are strings that are the translations of 'Most read' and 'Last updated'. 

**Acceptance criteria**
Check it meets all previously defined acceptance criteria - specifically when tabbing through the component. A11y notes: https://app.zeplin.io/project/5baa564c05c2eba0d2b89568/screen/5cee7072c6c48d1d928c5335 & RTL examples at different breakpoints: https://app.zeplin.io/project/5baa564c05c2eba0d2b89568/dashboard?tag=Most%20read

Most Read - Ordered list

    Given I use a screen reader, including VoiceOver  (Note we are particularly checking VoiceOver as there is a bug with this screen reader)
    And I am viewing Most Read
    When navigating to the ordered list
    Then it should be announced as a list

Most Read - List item

    Given I use a screen reader
    And I am viewing Most Read
    When navigating to a list item in the ordered list
    Then the text should be announced
    And it should be prefixed by the announcement of the list item number

    Given I am viewing Most Read
    When the list items are displayed in 2 columns or more
    Then each column must read from left to right

    Given I am viewing a a Persian or other RTL page
    and I am viewing Most Read
    When the list items are displayed in 2 columns or more
    Then each column must read from right to left

    Given I am viewing Most Read
    When the list items are in columns
    Then they must be within one ordered list
    The list item number must meet the colour contrast ratio of 4.1:5

Most Read - List item links

    Given I use a keyboard only
    When I navigate to Most Read by pressing the TAB key
    Then the link for each list item must be focusable
    And these must have visible state change on focus
    And this must not be identified by colour alone
    And be in addition to the browser default focus outline styling

    Given I use a mouse
    When I hover over the list item number
    Then the mouse cursor will change from a pointer to a hand
    And the link must have a visible state change
    And this must not be identified by colour alone

Designs at different breakpoints: https://app.zeplin.io/project/5baa564c05c2eba0d2b89568/screen/5cee708039f2e51e018597fa

**Describe the solution you'd like**
[...]

PR #2491, opened 2019-10-24
Export MostReadItem from psammead-most-read 

### LINKED ISSUE #579: Create psammead-most-popular Item
Create a package `@bbc/psammead-most-read` as a Psammead component. Ensure this is an alpha version, since it will not yet be accessibility tested fully. 

Create the Item in the psammead-most-read package.

Most Read - List item links

    Given I use a keyboard only
    When I navigate to Most Read by pressing the TAB key
    Then the link for each list item must be focusable
    And these must have visible state change on focus
    And this must not be identified by colour alone
    And be in addition to the browser default focus outline styling

    Given I use a mouse
    When I hover over the list item number
    Then the mouse cursor will change from a pointer to a hand
    And the link must have a visible state change
    And this must not be identified by colour alone

image

Should match the UX designs, and work correctly rtl.  https://app.zeplin.io/project/5baa564c05c2eba0d2b89568/dashboard?tag=Most%20read

Links: GreatPrimer (typography), SerifMedium weight for Reith (Note it falls back to Bold for other services that do not have Reith) Colour C_EBON - hover/focus/visted colours should be idential to those used for headlines.
Numbers: Foolscap (typography), SerifLight weight for Reith (Falls back to Regular for other services that do not have Reith) Colour C_POSTBOX

Service | Numerals
-- | --
Bengali | Bengali
Burmese | Burmese
Pashto | Eastern Arabic
Persian | Eastern Arabic
All others | Western Arabic

Numerals can be found here: https://bbc.github.io/psammead/?path=/story/utilities-psammead-locales--numerals

N/A

**Testing notes**
[Tester to complete]

Dev insight: Snapshot tests should be added. Storybook should show examples - including RTL and LTR - using example numerals. 

**Additional context**
N/A

EinsteinNjoroge:
Resolves #579 

**Overall change:** _Create `psammead-most-read` package and most-read items component._

**Code changes:**

- _Create the `psammead-most-read` package._
- _Create the most-read items alpha component, with snapshot tests, storybook story and documentation._

**Note**
Currently, this package exports both the `MostReadItem` and `StyledCountSpan` as separate components. We will add a wrapper that will export a list of most read components using `@bbc/psammead-grid` to display the most read items on a grid in  https://github.com/bbc/psammead/issues/581

---


packages/components/psammead-most-read/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Description
+
+The `MostRead` component is designed to be used on the `article` pages, which are pages that display articles/stories. It's designed to have a numerical count and a link to another article.
```
hotinglok: Added a little more detail to the description. Might be a little verbose but we can cut the last bit detailing how the component is made based on individual parts.

```suggestion
The `MostRead` component is designed to display the most read articles given a designated period of time which is dependent on service. The component comprises of a `MostReadTitle`, a `MostReadList` which is a grid containing `MostReadItems`. A `MostReadItem` comprises of a numerical counter representing it's ranking and a link to the article.
```

packages/components/psammead-most-read/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+<MostRead item={item} count="1" service="news" dir="ltr" />;
+
+```
+
+![most-read-single-item](documentation/most-read-single-item.png)
+
+### When to use this component
+
+<!-- ### When not to use this component -->
```
hotinglok:
```suggestion
This component is to be used on `article` pages.
```

hotinglok (commented):
Will recheck tomorrow but I think some of the padding/margins are a bit buggy when playing around on storybook.
image

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
rhenshaw56: @EinsteinNjoroge, question, why is this file name `item.jsx` ? and the `main` parameter in `package.json` still `index.js`

EinsteinNjoroge: The idea is to have an `index.js` file which will export a list of most read items. That will be the next step in this component's roadmap.

rhenshaw56: thanks, that makes sense

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,79 @@
...
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    margin-top: 6px;
+  }
+`;
+
+export const CountWrapper = styled.p`
+  ${({ script }) => script && getCanon(script)};
+  ${({ service }) => getSerifLight(service)}
+  color: ${C_POSTBOX};
+  margin: 0; /* Reset */
```
rhenshaw56, line 17: could we get rid of the commented bit

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,79 @@
...
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    width: 4rem;
+  }
+`;
+
+export const ItemWrapper = styled.a`
+  ${({ script }) => script && getPica(script)};
+  ${({ service }) => getSerifMedium(service)}
+  color: ${C_EBON};
+  margin: 0; /* Reset */
```
rhenshaw56: here as well

rhenshaw56 (approved):

packages/components/psammead-most-read/src/item.jsx
EinsteinNjoroge: Done

packages/components/psammead-most-read/src/item.jsx
EinsteinNjoroge: This comment explains that we're resetting any margins that are pre-defined by `GEL` and/or `styled component` similar to what we have on other components, see https://github.com/bbc/psammead/blob/2932750562232090da693abd707a5b39d6ac6da7/packages/components/psammead-headings/src/index.jsx#L17

packages/components/psammead-most-read/src/item.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+const ltrItem = {
+  header: 'John Lewis staff bonus cut again as profits fall',
+  href: 'https://www.bbc.com/vietnamese/institutional-49283563',
+};
+const rtlItem = {
+  header: "ایران از لغو 'رزمایش قطع اینترنت' خبر داد",
+  href: 'https://www.bbc.com/vietnamese/institutional-49283563',
+};
+
+describe('Most read', () => {
```
sareh: Good to test both the RTL and LTR variants. 👍

packages/components/psammead-most-read/README.md
```diff
@@ -0,0 +1,108 @@
...
+## Description
+
+The `MostReadItem` component is designed to display the most read articles given a designated period of time which is dependent on service. The component comprises of a `MostReadTitle`, a `MostReadList` which is a grid containing `MostReadItems`. A `MostReadItem` comprises of a numerical counter representing it's ranking and a link to the article.
+
+## Installation
+
+`npm install @bbc/psammead-most-read/esm/item`
+
+## Props
+
```
sareh, line 13:
```suggestion
### MostReadItem props
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+import {
+  getSerifLight,
+  getSerifMedium,
+} from '@bbc/psammead-styles/font-styles';
+
+const MostReadWrapper = styled.div`
+  display: flex;
+
+  @media (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    margin-top: 4px;
```
sareh: We should be using rems instead of pixels, so that when someone zooms in on a page, everything scales accordingly, including the spacing. 
```suggestion
    margin-top: 0.25rem;
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    margin-top: 6px;
```
sareh:
```suggestion
    margin-top: 0.0625rem;
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+`;
+
+const ItemWrapper = styled.div``;
+
+export const CountWrapper = styled.p`
+  ${({ script }) => script && getCanon(script)};
```
sareh: The number should use the `Foolscap` type:
```suggestion
  ${({ script }) => script && getFoolscap(script)};
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+  padding-bottom: ${GEL_SPACING};
+  display: inline-block;
+  min-width: 3rem;
+  
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    width: 4rem;
+  }
+`;
+
+export const TitleWrapper = styled.a`
```
sareh: Calling this StyledLink would be in keeping with the naming convention of other styled components in Psammead, since it is reflective of the type of underlying element, in this case a link.
```suggestion
export const StyledLink = styled.a`
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    margin-top: 6px;
+  }
+`;
+
+const ItemWrapper = styled.div``;
+
+export const CountWrapper = styled.p`
```
sareh: It'd be better to use a span here surrounding the number. The naming change is to follow the convention of highlighting the underlying HTML element of the styled component. 
```suggestion
export const StyledCountSpan = styled.span`
```

packages/components/psammead-most-read/README.md
```diff
@@ -0,0 +1,108 @@
...
+<!-- prettier-ignore -->
+| Argument | Type | Required | Default | Example |
+| -------- | ---- | -------- | ------- | ------- |
+| item | object | yes | N/A | `{ title: 'This is a item', href: 'https://www.bbc.com' }` |
+| count | string | yes | N/A | `'rtl'`  |
```
12: Is this prop name correct? I don't think the naming of this prop matches the values I'd expect it to take?

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+  display: inline-block;
+  min-width: 3rem;
+  
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    width: 4rem;
+  }
+`;
+
+export const TitleWrapper = styled.a`
+  ${({ script }) => script && getPica(script)};
```
sareh: The type for the link item here should be Double Pica
```suggestion
  ${({ script }) => script && getDoublePica(script)};
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+  ${({ service }) => getSerifMedium(service)}
+  color: ${C_EBON};
+  text-decoration: none;
+  padding-bottom: ${GEL_SPACING};
+  
+  &:hover,
+  &:focus {
+    text-decoration: underline;
+    cursor: pointer;
```
sareh: We shouldn't need to define the cursor style on hover and focus - that should be handled using browser defaults.
```suggestion
```

sareh (changes requested):
@EinsteinNjoroge Great start to the most read component! I've suggested some changes, as comparing the component with the Zeplin designs.

packages/components/psammead-most-read/README.md
sareh:
```suggestion
| count | string | yes | N/A | `'2'`  |
```

hotinglok (changes requested):
May need to add padding to `ItemWrapper`
Edit: Github killed all my suggestion comments it seems... _sigh_
image

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import { node, oneOf, shape, string } from 'prop-types';
+import styled from 'styled-components';
+import { getCanon, getPica } from '@bbc/gel-foundations/typography';
```
hotinglok:
```suggestion
import { getFoolscap, getDoublePica } from '@bbc/gel-foundations/typography';
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,89 @@
...
+`;
+
+const MostReadItem = ({
+  count,
+  lastUpdated,
+  item: { title, href },
+  ...props
+}) => (
+  <MostReadWrapper {...props}>
+    <CountWrapper {...props}>{count}</CountWrapper>
```
hotinglok:
```suggestion
    <StyledCountSpan {...props}>{count}</StyledCountSpan>
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,89 @@
...
+    <ItemWrapper>
+      <TitleWrapper href={href} {...props}>
```
hotinglok:
```suggestion
      <StyledLink href={href} {...props}>
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,89 @@
...
+        {title}
+      </TitleWrapper>
```
hotinglok:
```suggestion
      </StyledLink>
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+`;
+
+const MostReadItem = ({
+  count,
+  lastUpdated,
+  item: { title, href },
+  ...props
+}) => (
+  <MostReadWrapper {...props}>
```
sareh: Since we add `{...props}` at three levels in this component, we duplicate certain props such as `dir`. It'd be better to add the `dir` prop explicitly on the outer-most element.

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+const MostReadItem = ({
+  count,
+  dir,
+  lastUpdated,
+  item: { title, href },
+  ...props
+}) => (
+  <MostReadWrapper dir={dir}>
+    <StyledCountSpan {...props}>{count}</StyledCountSpan>
```
sareh: You no longer need to spread props onto these elements
```suggestion
    <StyledCountSpan>{count}</StyledCountSpan>
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+    <ItemWrapper>
+      <StyledLink href={href} {...props}>
```
sareh:
```suggestion
      <StyledLink href={href}>
```

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,90 @@
...
+  ${({ script }) => script && getFoolscap(script)};
+  ${({ service }) => getSerifLight(service)}
+  color: ${C_POSTBOX};
+  margin: 0; /* Reset */
+  padding-bottom: ${GEL_SPACING};
+  display: inline-block;
+  min-width: 3rem;
+  
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    width: 4rem;
```
sareh: We can have the width set at 3rem for all breakpoints. Then we can remove this media query and we shouldn't need the `min-width` value above either. This can simplify the CSS we have for this component.

sareh (changes requested):

packages/components/psammead-most-read/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Description
+
+The `MostReadItem` component is designed to display the most read articles given a designated period of time which is dependent on service. The component comprises of a `MostReadTitle`, a `MostReadList` which is a grid containing `MostReadItems`. A `MostReadItem` comprises of a numerical counter representing it's ranking and a link to the article.
```
HarveyPeachey:
```suggestion
The `MostReadItem` component is designed to display the most read articles given a designated period of time which is dependent on service. The component comprises of a `MostReadTitle`, a `MostReadList` which is a grid containing `MostReadItems`. A `MostReadItem` comprises of a numerical counter representing its ranking and a link to the article.
```

packages/components/psammead-most-read/README.md
```diff
@@ -0,0 +1,108 @@
...
+### MostReadItem props
+<!-- prettier-ignore -->
+| Argument | Type | Required | Default | Example |
+| -------- | ---- | -------- | ------- | ------- |
+| item | object | yes | N/A | `{ title: 'This is a item', href: 'https://www.bbc.com' }` |
+| count | string | yes | N/A | `'2'`  |
+| script | object | yes | N/A | `{ canon: { groupA: { fontSize: '28', lineHeight: '32',}, groupB: { fontSize: '32', lineHeight: '36', }, groupD: { fontSize: '44', lineHeight: '48', }, }, trafalgar: { groupA: { fontSize: '20', lineHeight: '24', }, groupB: { fontSize: '24', lineHeight: '28', }, groupD: { fontSize: '32', lineHeight: '36', }, }, }` |
+| service | string | yes | N/A | `'news'` |
+| dir | string | no | `'ltr'` | `'rtl'`  |
+| lastUpdated | node | no | N/A | `<time>12 March 2019</time>` |
```
HarveyPeachey:
```suggestion
| lastUpdated | node | no | null | `<time>12 March 2019</time>` |
```

packages/components/psammead-most-read/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Usage
+
+A typical use-case of this component is as displayed below. It contains a count and an info elements. The info element is a link which point to the corresponding article while the count is a number that represent the position of the info element on the list of info elements.
```
HarveyPeachey:
```suggestion
A typical use-case of this component is as displayed below. It contains a count and an info element. The info element is a link which points to the corresponding article while the count is a number that represents the position of the info element on the list of info elements.
```

HarveyPeachey (changes requested):
Looks good so far just a few things.

hotinglok:
image

hotinglok:
Lagged and closed by accident... my b, please ignore.

RayNjeri (approved):
Good job on the documentation.
Its on point 🙌

packages/components/psammead-most-read/src/item.stories.jsx
```diff
@@ -0,0 +1,76 @@
...
+    datetime="2019-03-01T14:00+00:00"
+    script={script}
+    padding={false}
+    service={service}
+  >
+    Last updated: 5th November 2016
+  </Timestamp>
+);
+
+const stories = storiesOf('Components|MostRead/MostReadItem', module)
```
hotinglok:
```suggestion
const stories = storiesOf('Components|MostRead/Item', module)
```

packages/components/psammead-most-read/src/item.stories.jsx
```diff
@@ -0,0 +1,76 @@
...
+  .addDecorator(withKnobs)
+  .addDecorator(withServicesKnob());
+
+items.forEach(({ item, dir }) => {
+  stories.add(`MostReadItem ${dir}`, ({ script, service }) => (
```
lokh01:
```suggestion
  stories.add(`Item ${dir}`, ({ script, service }) => (
```

hotinglok (commented):
Looks good, just a few minor things like naming etc. I'm also wondering if there's a way to set the default service as persian instead of news for the rtl stories. Will look into the `buildRTLSubstory` function a bit more to see if we can make it a bit more consistent with other stories.

HarveyPeachey (commented):
Everything looks good but the `psammead-most-read` component needs to be added to the  packages [README](https://github.com/bbc/psammead/blob/latest/packages/README.md#components) file.

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,77 @@
...
+  </ItemWrapper>
+);
+
+StyledCountSpan.propTypes = {
+  service: string.isRequired,
+  script: shape(scriptPropType).isRequired,
+};
+
+MostReadItem.propTypes = {
+  dir: string,
```
ghost, line 57: Do we really need `dir` prop? I can see it is being used on the stories, but removing it and changing the service from the service knob to a `rlt` service would achieve the same result.

EinsteinNjoroge:
> Looks good, just a few minor things like naming etc. I'm also wondering if there's a way to set the default service as persian instead of news for the rtl stories. Will look into the `buildRTLSubstory` function a bit more to see if we can make it a bit more consistent with other stories.

@lokh01 Since this component is still in `alpha`, I'm thinking we can refine it and its stories in future PR. Currently, I'm focusing on getting this PR merged so that we can unblock https://github.com/bbc/psammead/issues/581 and https://github.com/bbc/psammead/issues/1429

packages/components/psammead-most-read/src/item.jsx
```diff
@@ -0,0 +1,77 @@
...
+import { C_EBON, C_POSTBOX } from '@bbc/psammead-styles/colours';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+import { GEL_SPACING } from '@bbc/gel-foundations/spacings';
+import { GEL_GROUP_3_SCREEN_WIDTH_MIN } from '@bbc/gel-foundations/breakpoints';
+import {
+  getSerifLight,
+  getSerifMedium,
+} from '@bbc/psammead-styles/font-styles';
+
+const ItemWrapper = styled.div``;
```
ghost: `ItemWrapper` is an empty div with no styles in it. I think this should be removed.

ghost (changes requested):
Add a story with both the `count` and the `title`

hotinglok:
@sadickisaac The way we've chosen to go about doing this now is by adding the `count` and `item` into two separate columns of a grid, so we can't make a 'complete' item story with both a `count` and an `item` at the moment.

ghost:
@lokh01
 > The way we've chosen to go about doing this now is by adding the count and item into two separate columns of a grid

But in storybook `count` and `item` could be composed to form a complete picture of how the `most-read` item is supposed to look like. We could use `psammead-grid` in storybook to compose `count` and `item`.

Is there something am missing? @EinsteinNjoroge @lokh01

hotinglok:
@sadickisaac 
My b, I thought for some reason grid wasn't usable yet because I mixed it up with another ongoing grid pr we have atm. I blame the fact that it's Monday morning 😅

packages/components/psammead-most-read/src/item.jsx
EinsteinNjoroge: I think it would be useful to have this prop being explicitly declared even if it's being passed automatically in storybook.

EinsteinNjoroge:
> @lokh01
> 
> > The way we've chosen to go about doing this now is by adding the count and item into two separate columns of a grid
> 
> But in storybook `count` and `item` could be composed to form a complete picture of how the `most-read` item is supposed to look like. We could use `psammead-grid` in storybook to compose `count` and `item`.
> 
> Is there something I am missing? @EinsteinNjoroge @lokh01

@sadickisaac based on the designs, adding a story for these two components at this level will give an erroneous representation of how this component will look like on the list.

packages/components/psammead-most-read/src/item.jsx
ghost: In most of our components in `psammead` we pass in `dir` either for correct padding or floats. `dir` in `MostReadItem` doesn't require any of those.

packages/README.md
```diff
@@ -92,6 +92,7 @@ Or if you're using [styled-components](https://styled-components.com), you can u
...
+| [`@bbc/psammead-most-read`](./components/psammead-most-read) |[![npm version](https://img.shields.io/npm/v/@bbc/psammead-most-read.svg)](https://www.npmjs.com/package/@bbc/psammead-most-read) |[![Dependency Status](https://david-dm.org/bbc/psammead.svg?path=packages/components/psammead-most-read)](https://david-dm.org/bbc/psammead?path=packages/components/psammead-most-read) | [![peerDependencies Status](https://david-dm.org/bbc/psammead/peer-status.svg?path=packages/components/psammead-most-read)](https://david-dm.org/bbc/psammead?path=packages/components/psammead-most-read&type=peer) |
```
sareh: Can we put this after `@bbc/psammead-media-player` so it's alphabetical?

packages/components/psammead-most-read/README.md
```diff
@@ -3,6 +3,109 @@
...
+
+Currently this component is in alpha. This is because it has not yet been tested with various assistive technologies. After it has had an accessibility swarm, this will be published under a standard version.
+
+## Roadmap
+
+Currently, this package only exports the `MostReadItem` component. We will add a wrapper that will export a list of most read components using `@bbc/psammead-grid` to display the most read items on a grid.
+=======
 The `MostRead` component is designed to display the most read articles given a designated period of time which is dependent on service. The component comprises of a `MostReadTitle`, a `MostReadList` which is a grid containing a `MostReadCount/Rank` (name subject to change) and a `MostReadItem`.
 
 ## Installation
```
sareh: You can delete this section (Installation / MostReadTitle Props table), since it is duplicated with the installation steps above.

packages/components/psammead-most-read/src/item.jsx
sareh: I don't believe this comment was addressed yet?

packages/components/psammead-most-read/src/item.jsx
sareh: @sadickisaac In this case, `dir` isn't for layout - it is for correctly stating the direction of the text contained within the MostReadItem. It is an element attribute needed for the HTML output, not for the CSS.

sareh (approved):
Thanks for the updates! This looks good. 👍

ghost (approved):

HarveyPeachey (approved):
LGTM

Bopchy (approved):

hotinglok (approved):

PR #2493, opened 2019-10-25
Show canonical placeholder while media is loading.

### LINKED ISSUE #2381: Canonical media player placeholder while load scenario
Currently the `psammead-media-player` canonical component placeholder logic is as follows:
1) Placeholder is set to false, no placeholder is used and the iframe'd in SMP shows once loaded.
2) Placeholder is set to true, a placeholder is rendered and the iframe is not downloaded until the user clicks on the placeholder. 

Therefore the following scenario does not exist:
Placeholder is set to true, a placeholder is rendered and the iframe'd in SMP shows once loaded. 

I believe this is due to the placeholder logic being defined by Articles and their use case is to delay the loaded of the SMP instance until the user have interacted with the media content. For Live radio and MAPs we require a placeholder to be visible while the iframe source is downloading as part of our UX (mainly for user in slow connections to not scroll past the content). 

Change the articles logic for the prop `placeholder` to be a prop such as `loadPlayerBeforeInteraction` that defaults to `true` and in articles is set to `false`. Then make a placeholder image or component (articles currently build a component as the placeholder) a new required prop in the Media Player. 

We could maintain the 2 current scenarios and add a new 3rd scenario but this seems like an anti pattern to me. Also as we are AMP-first and the AMP version of the Media Player requires a placeholder we should likely mimic this logic on Canonical for UX best practices. 

**Testing notes**
[Tester to complete]

Dev insight: All scenarios should be regression checked to ensure the behaviour is correct and both media and articles requirements are met. This may require confirming with each team what those requirements are. 

**Additional context**
I've labelled this with Articles for exposure and potential input.

pharingee:
Resolves #2381

**Overall change:** 
Show a canonical placeholder while media is loading

**Code changes:**
- Add a `placeholderSrc` prop to Canonical Media player
- Set `placeholderSrc` to the placeholder passed to the container component

---


packages/components/psammead-media-player/src/Canonical/index.jsx
```diff
@@ -12,19 +12,27 @@ const StyledIframe = styled.iframe`
...
-const Canonical = ({ src, title }) => (
+const Canonical = ({ src, title, placeholderSrc }) => (
   <StyledIframe
     src={src}
     title={title}
     allow="autoplay; fullscreen"
     gesture="media"
     allowFullScreen
+    style={
+      placeholderSrc ? { backgroundImage: `url(${placeholderSrc})` } : null
```
dr3: We should use styled components :)

pharingee: Done

jroebu14 (approved):
looks good

pjlee11:
Checking this locally on storybook I can't see any placeholder being shown. Are you sure that an `background-color` styles show on an `<iframe>`?

no placeholder

dr3 (dismissed):
image
Code looks nice, and the placeholder shows for me, but then it never goes away, and doesnt scale :(

ryanmccombe (approved):

PR #2476, opened 2019-10-23
AV: Add empty img alt tag to placeholder

HarveyPeachey:
Resolves #1840

**Overall change:** Updated image alt to be empty to follow desired behaviour for screen reader

**Code changes:**

- Replaced `alt="Image Alt"` with `alt=""` on video image placeholder
- Updated snapshots

---

- [x] I have assigned myself to this PR and the corresponding issues
- [x] Automated jest tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

PR #2516, opened 2019-10-28
Fix prop type warnings

### LINKED ISSUE #2497: Fix psammead-media-player prop types warnings
The `psammead-media-player` is emitting prop-type warnings both on Storybook and the console.

Fix these warnings.

N/A.

**Testing notes**
[Tester to complete]

**Additional context**

Bopchy:
Resolves #2497 

**Overall change:** _Add notes to `psammead-media-player` stories. Fix prop type warnings._

**Code changes:**

- _Add notes to psammead-media-player stories._
- _Fix prop type warnings._

---


Bopchy:
This PR hasn't fixed - `Warning: Failed prop type: The prop `width` is marked as required in `AmpImg`, but its value is `undefined`.` and ` Warning: Failed prop type: The prop `height` is marked as required in `AmpImg`, but its value is `undefined`.`

Considered making <AmpImg /> height & width props optional in `psammead-image` (see #2523), however this will require `AmpImg` in `psammead-image/src/index.amp.jsx` to always be used within a container with specified height or width so that AMP can use the correct ratio for scaling the images

**UPDATE** - Closed #2523 and instead provided a default value for the height and width

ghost (dismissed):
Screenshot 2019-10-29 at 19 08 45
Still seeing warnings while running `test:unit`

ghost (approved):
Looks great.

packages/components/psammead-media-player/src/Amp/index.jsx
```diff
@@ -41,9 +52,15 @@ AmpMediaPlayer.propTypes = {
   placeholderSrc: string.isRequired,
   placeholderSrcset: string,
   title: string.isRequired,
+  alt: string,
```
ghost: Should this be a required prop?

packages/components/psammead-media-player/src/Amp/index.jsx
```diff
@@ -41,9 +52,15 @@ AmpMediaPlayer.propTypes = {
...
+  height: number,
+  width: number,
 };
 AmpMediaPlayer.defaultProps = {
   placeholderSrcset: null,
+  alt: 'Image alt',
+  height: 640,
+  width: 450,
```
ghost: Where are we getting this values from?

packages/components/psammead-media-player/src/Amp/index.jsx
DenisHdz: Also, the `width` and `height` are required in the `AmpImg` component.

rhenshaw56 (approved):

packages/components/psammead-media-player/src/Amp/index.jsx
Bopchy: Provided random default values as I was not sure where to get the specific height and width

packages/components/psammead-media-player/src/Amp/index.jsx
Bopchy: Added the alt, width and height as optional props because the `<AmpMediaPlayer />` seems to work without them - as it's getting the height and width from the `StyledVideoContainer` and `StyledAudioContainer`

packages/components/psammead-media-player/src/Amp/index.jsx
ghost: Let's have `null` as the default values. The same thing we have on `psammead-image` component

Bopchy: When height and width are null, the error changes to `Warning: Failed prop type: The prop height is marked as required in AmpImg, but its value is null.`.  React doesn't allow required props to be null. Found this issue on that - https://github.com/facebook/react/issues/3163

packages/components/psammead-media-player/src/index.stories.jsx
```diff
@@ -2,61 +2,90 @@ import React from 'react';
...
+      service="news"
+      mediaInfo={{ type: 'audio', title: 'Live show intro.' }}
       title="Audio player"
     />
-  ));
+  ),
+  { notes, knobs: { escapeHTML: false } },
+);
+
+storiesOf('Components|Media Player', module)
```
EinsteinNjoroge: I think a simple `.add` to line `75` will be better than redefining this story.

packages/components/psammead-media-player/src/index.stories.jsx
```diff
@@ -2,61 +2,90 @@ import React from 'react';
...
-    skin="audio"
-    service="news"
-    title="Audio player"
-  />
-));
+  ),
+  { notes, knobs: { escapeHTML: false } },
+);
 
 storiesOf('Components|MediaPlayer', module)
```
EinsteinNjoroge: Think we should update this as well to `Media Player` for consistency.

packages/components/psammead-media-player/src/Amp/index.jsx
DenisHdz: I think the width and height should be required since they are in the `AmpImg` component. Can you double-check this with @RayNjeri, please?

packages/components/psammead-media-player/src/Amp/index.jsx
```diff
@@ -13,7 +13,15 @@ const AmpHead = () => (
   </Helmet>
 );
 
-const AmpMediaPlayer = ({ src, placeholderSrc, placeholderSrcset, title }) => {
+const AmpMediaPlayer = ({
+  src,
+  placeholderSrc,
+  placeholderSrcset,
+  title,
+  alt,
```
jamesdonoh: I don't think we need a new `alt` prop, it can be hardcoded to `""` on L41 for consistency with https://github.com/bbc/psammead/pull/2476

packages/components/psammead-media-player/src/Amp/index.jsx
jamesdonoh: Thanks for catching this. I think `height` and `width` should be required props here and not given arbitrary default values.

Given the width and height are only used before the AMP runtime fetches the image, can we compute [values matching the aspect ratio](https://amp.dev/documentation/components/amp-img/#setting-the-aspect-ratio) based on `portrait`, see how we do https://github.com/bbc/psammead/blob/7b7d4eb86022e7edbfa37bd1a43c4d1633f6f967/packages/components/psammead-media-player/src/index.jsx#L11 and pass these in?

If it doesn't look straightforward I'm happy if you want to create a separate issue for the articles pod to implement that and we live with the warnings for now.

jamesdonoh (changes requested):

Bopchy:
Added a height and width of 16 or 9 depending on if `portrait == true` or not

RayNjeri (approved):

EinsteinNjoroge (approved):

jamesdonoh (approved):
Thanks for the changes.

sareh (approved):

PR #2565, opened 2019-11-05
Psammead Grid - handling case where child is undefined

sareh:
Part of  https://github.com/bbc/simorgh/issues/3686

**Overall change:** Currently, if you use `<Grid>` without a child element, it errors with a `child is undefined` error. This fixes it so that it can gracefully degrade.

**Code changes:**

- Check that child is defined (not null) prior to checking `child.type`. 

I have manually checked this fixes the issue by pulling in the change in this Simorgh PR: https://github.com/bbc/simorgh/pull/4493

---


packages/components/psammead-grid/src/index.jsx
```diff
@@ -287,13 +287,15 @@ const Grid = ({
...
-          parentEnableGelGutters: otherProps.enableGelGutters,
-        });
+        if (isNestedGridComponent) {
+          return React.cloneElement(child, {
+            parentColumns: otherProps.columns,
+            parentEnableGelGutters: otherProps.enableGelGutters,
+          });
+        }
       }
       return child;
```
jroebu14, line 300: I think it's possible to remove this line because `undefined` will be returned by default which also shouldn't render anything.

jroebu14 (approved):

packages/components/psammead-grid/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 1.0.0-alpha.12 | [PR#x](https://github.com/bbc/psammead/pull/x) Fix case where child of Grid component is not defined |
```
DenisHdz:
```suggestion
| 1.0.0-alpha.12 | [PR#2565](https://github.com/bbc/psammead/pull/2565) Fix case where child of Grid component is not defined |
```

packages/components/psammead-grid/src/index.jsx
sareh: I believe we need this so that the `renderChildren` function returns a child when child is defined.

jroebu14: Yeh you're right ignore me!

EinsteinNjoroge (approved):
👍

DenisHdz (approved):

PR #2575, opened 2019-11-06
Explicitly set h1 styles display and font weight

pjlee11:
Related to xxx

**Overall change:** Explicitly adds `font-weight: 500` and `display: block` which are defaults on `<h1>`, so that we can use this component `as="strong"` for the `OnScreenHeadline`

---

### DIFF
--- a/packages/components/psammead-headings/src/__snapshots__/index.test.jsx.snap
+++ b/packages/components/psammead-headings/src/__snapshots__/index.test.jsx.snap
@@ -9,7 +9,6 @@ exports[`Headline component should render correctly 1`] = `
   font-style: normal;
   color: #3F3F42;
   display: block;
-  font-weight: 500;
   margin: 0;
   padding: 2rem 0;
 }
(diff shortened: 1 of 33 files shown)


packages/components/psammead-headings/src/__snapshots__/index.test.jsx.snap
```diff
@@ -47,6 +49,8 @@ exports[`Headline component should render correctly with arabic script typograph
   font-weight: 700;
```
dr3, line 48: :O

dr3 (changes requested):

packages/components/psammead-headings/src/__snapshots__/index.test.jsx.snap
pjlee11: Nothing to see here .... pushed changes

dr3 (approved):

ryanmccombe (approved):

12 (approved):

PR #2577, opened 2019-11-07
Scrollable Navigation

### LINKED ISSUE #2553: Mobile styling for swipeable navigation
The new navigation designs require the list to remain on one line < 600px, rather than wrapping to a second line. The navigation should then be horizontally scrollable to allow access to further links. Designs can be seen on zeplin:

Screenshot 2019-11-04 at 17 17 09

This story does not include adding the burger menu on the left of the navigation - but this component should allow for the burger menu to placed alongside it, outside of the scrollable area (i.e. don't make the whole psammead navigation horizontally scrollable).

\> 600px, the navigation should wrap as currently, according to these designs:

Screenshot 2019-11-04 at 17 17 52

Considerations:
- Scrollbars should not be visible 
- Consider momentum -  use the overflow scrolling css property to provide natural scrolling to mobile devices
- Gradient overlay - check if we need any IE fallback, and also check with high contrast mode.
- How does horizontal scrolling interaction work with voiceover enabled?
- Test against https://bbc.github.io/accessibility-news-and-you/accessibility-news-and-developers
- The navigation should also be scrollable for rtl services, as can be seen on the following designs:

< 600px:
Screenshot 2019-11-04 at 17 18 21

\> 600px:
Screenshot 2019-11-04 at 17 18 50

N/A

**Testing notes**
- Regression test in psammead storybook
- Gradient overlay - check if we need any IE fallback, and also check with high contrast mode.
- ~How does horizontal scrolling interaction work with voiceover enabled?~ The swipeable nav should not be exporsed to screen readers and the links should be removed from the tab order.

**Additional context**

DenisHdz:
Resolves #2553

**Overall change:**
Add Scrollable Navigation.

**Code changes:**
- Add `CanonicalSwipeableNavigation` and `AmpSwipeableNavigation` components which allow the list to remain on one line under 600px and to be horizontally scrollable.
- Add a gradient overlay as a pseudo-element of the `SwipeableNav`.
- Add `C_POSTBOX_TRANSPARENT` and `C_POSTBOX_OPAQUE` constants to be used in the linear gradient as IE doesn't support 8 digit hex like `{C_POSTBOX}00` and `{C_POSTBOX}FF`.
- Add margin-right/left to the last list item with the same width as the gradient.
- Add custom `useMediaQuery` hook to determine if the document matches the media query `(max-width: 600px)`.
- ~~Add custom `useOverflowed` hook to apply a gradient to the swipeable navigation if the scroll width is bigger than the offset width.~~
- ~~Add custom `useWindowWidth` hook to listen for the window resize and set the width of the window so we can pass it to the `useOverflowed` hook and apply the gradient also on resize.~~
- Add `aria-hidden='true'` to the `CanonicalSwipeableNavigation` under 600px to not expose it to screen readers, as we want that just the links in the Menu dropdown are announced.
- Remove the links from the tab order with `tabindex=-1` under 600px.
- Clone children to pass the `isSwipeable` prop so we can add the `tabindex=-1` under 600px.

---


packages/components/psammead-navigation/src/index.jsx
```diff
@@ -157,11 +179,68 @@ const StyledNav = styled.nav`
...
-    <NavWrapper>{children}</NavWrapper>
-  </StyledNav>
-);
+const SwipeableNav = styled.div`
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
+    position: relative;
+    overflow-x: scroll;
+    white-space: nowrap;
+    scroll-behavior: smooth;
+    -webkit-overflow-scrolling: touch;
```
DenisHdz: Tested this on iOS 13 and the navigation doesn't scroll.. If anyone has an idea why, please shout it out.

thekp:
Probably worth adding the new the new prop ('isSwipeable') to the NavigationUl readme section.

Also, more of a preference but could the prop type and prop default declarations go at the bottom of the file after the exported components ? then the file structure would be imports>consts>exports>props.

DenisHdz:
> Probably worth adding the new the new prop ('isSwipeable') to the NavigationUl readme section.
> 
> Also, more of a preference but could the prop type and prop default declarations go at the bottom of the file after the exported components ? then the file structure would be imports>consts>exports>props.

I agree, but I have moved them after the component declaration as we are trying to do that from now on. Take a look at other components.. happy to discuss it further with all the team.

thekp (approved):
> I agree, but I have moved them after the component declaration as we are trying to do that from now on. Take a look at other components.. happy to discuss it further with all the team.

if this the case then, LGTM.

HarveyPeachey:
Just tried on my iPhone 7 and swiping isn't working on the `navigation` story on storybook

HarveyPeachey:
> Just tried on my iPhone 7 and swiping isn't working on the `navigation` story on storybook

Actually it does work nvm. It's just the Igbo example has the opaque shadow which makes it look like it's scrollable even though it's at the end of the navigation container:


ezgif-3-6a4c30d3aadb

ghost (commented):
Well executed. The only thing missing is  a comment why we would want the tabIndex to be `-1` when the navigation is swappable .

DenisHdz:
> > Just tried on my iPhone 7 and swiping isn't working on the `navigation` story on storybook
> 
> Actually it does work nvm. It's just the Igbo example has the opaque shadow which makes it look like it's scrollable even though it's at the end of the navigation container:


Thanks @HarveyPeachey. I am working at the moment on applying the gradient just when the navigation is swipeable 👍

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -109,43 +118,91 @@ const CurrentLink = ({ children: link, script, currentPageText }) => (
...
+}) => {
+  const tabIndex = isSwipeable && { tabIndex: -1 };
+
+  return (
+    <StyledListItem role="listitem">
+      {active && currentPageText ? (
+        <StyledLink
+          href={url}
+          script={script}
+          currentLink="true"
```
tochwill: Looking good! I realise this hasn't been changed in this PR, but why is `currentLink` passed as a string instead of a boolean?

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -157,45 +214,73 @@ const StyledNav = styled.nav`
...
+    &:before {
+      content: ' ';
+      height: 100%;
+      width: 3rem;
+      position: absolute;
+      ${({ dir }) => css`
+        ${dir === 'ltr' ? 'right' : 'left'}: 0;
+      `}
+      bottom: 0;
+      z-index: 3;
```
tochwill: It might be an idea to have z-index constants at some point (something like https://css-tricks.com/handling-z-index/), not for this PR though

tochwill (commented):
Looks good! If you're working on only adding the gradient when the nav is swipeable I'll take a look again when that's done :)

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -157,45 +214,73 @@ const StyledNav = styled.nav`
...
+      );
+    }
+  }
+`;
 
-NavigationUl.propTypes = {
-  children: node.isRequired,
-};
+const Navigation = ({ children, dir }) => {
+  const [width, setWidth] = useState(window.innerWidth);
```
tochwill: Hmm, just thinking, how does this work on the server? Is innerWidth undefined at this point?

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -33,22 +39,25 @@ const StyledUnorderedList = styled.ul`
   padding: 0;
   margin: 0;
   position: relative;
-  overflow: hidden;
+
+  ${({ isSwipeable }) => !isSwipeable && `overflow: hidden;`}
```
tochwill: As `isSwipeable` is set currently set by javascript checking the window's width, could we replace this with a media query here? Then we will get the correct navigation behaviour on the server render as well?

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -157,45 +214,103 @@ const StyledNav = styled.nav`
...
-  script: shape(scriptPropType).isRequired,
-  currentPageText: string,
+function useWindowWidth() {
+  const [width, setWidth] = useState(window.innerWidth);
+
+  useEffect(() => {
+    const handleResize = () => {
+      setWidth(window.innerWidth);
+    };
+    window.addEventListener('resize', handleResize);
```
tochwill: Could we replace this with a media query listener? https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList - Then we can update the value of `isSwipeable` when that event fires, rather than handling every resize event? (Would need to check browser support, I think it's fine!)

packages/components/psammead-navigation/src/index.jsx
DenisHdz: Good spot 👍

packages/components/psammead-navigation/src/index.jsx
DenisHdz: Done

Bopchy (commented):
The `### When to use this component` section on the README needs updating as the items no longer wrap to the next line

packages/components/psammead-navigation/src/index.jsx
DenisHdz: Looking at this now, but I am not sure if we can get rid of the resize event as we need that the `useOverflowed` effect is run every time we resize the window, passing the `width`.

DenisHdz: I have created another custom hook `useMediaQuery` which uses the `matchMedia()` method to determine if the document matches the media query (`max-width: 600px`) https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia, so we can then pass the boolean value to `isSwipeable`. 

On the other hand, I am not sure if we can get rid of the resize event as we need that the `useOverflowed` effect is run every time we resize the window, passing the width.

packages/components/psammead-navigation/src/index.jsx
DenisHdz: I have added a condition so if it is executed on the server-side, the value of the width will be undefined.

AlistairGempf:
image

The gradient will cover the final item of the navigation item, which makes it look like there is more to swipe to, despite being the end of the list. In the prototype I got around this, I added an extra bit of space at the end of the list, another option is to remove the gradient using javascript when at or near the end of the swipable section.

packages/components/psammead-navigation/hooks/useWindowWidth.js
```diff
@@ -0,0 +1,28 @@
...
+
+  useEffect(() => {
+    if (!isClient) {
+      return null;
+    }
+
+    const handleResize = () => {
+      setWidth(getSize());
+    };
+    window.addEventListener('resize', handleResize);
```
tochwill: should we debounce this event handling?

DenisHdz:
> image
> 
> The gradient will cover the final item of the navigation item, which makes it look like there is more to swipe to, despite being the end of the list. In the prototype I got around this, I added an extra bit of space at the end of the list, another option is to remove the gradient using javascript when at or near the end of the swipable section.

@AlistairGempf I have added a function to handle the scroll and remove the gradient when scrolling to the end of the navigation.

packages/components/psammead-navigation/hooks/useWindowWidth.js
DenisHdz: Done. I've added a delay of 100ms.

packages/components/psammead-navigation/hooks/useWindowWidth.js
```diff
@@ -0,0 +1,30 @@
...
+  const [width, setWidth] = useState(getSize);
+
+  useEffect(() => {
+    if (!isClient) {
+      return null;
+    }
+
+    const handleResize = () => {
+      setWidth(getSize());
+      console.log('entro');
```
tochwill: Just remove this :)

DenisHdz: Ups 🙈

AlistairGempf:
Overall looks really good!

Is there any unit testing we can do of the hooks? I don't know how much is possible and how much benefit it would add, but thought it'd be good to bring it up as a question.

https://stackoverflow.com/questions/45868042/figuring-out-how-to-mock-the-window-size-changing-for-a-react-component-test suggests we can mock changing the window size (although I don't know if it would trigger the 'resize' event.

tochwill:
Just had another look! It looks really good, one thing to consider is how is this going to behave on AMP? The majority of functionality is CSS which will work, but the client side JS won't run to check scroll position and screen width

OlgaLyubin:
Since you added `react-test-renderer` and `@testing-library/react-hooks` shouldn't the main version of Psammead be bumped up? + CHANGELOG updated

DenisHdz:
> Since you added `react-test-renderer` and `@testing-library/react-hooks` shouldn't the main version of Psammead be bumped up? + CHANGELOG updated

Yeah, working on it 👍

packages/components/psammead-navigation/src/SwipeableNavigation/index.jsx
```diff
@@ -0,0 +1,99 @@
...
+            ${C_POSTBOX_OPAQUE} 100%
+          );
+        }
+      `}
+  }
+`;
+
+const SwipeableNavigation = ({ children, dir }) => {
+  const isSwipeable = useMediaQuery('(max-width: 600px)');
+  const ariaHidden = isSwipeable && { 'aria-hidden': true };
```
OlgaLyubin: As far as I know, for now a non-js navigation is just a swipeable one with dropdown disabled. In that case should we have an option to set `aria-hidden` to false so that tabbing is not disabled? Or is this too much of an exception case that we will not consider?

tochwill: That's something to consider - on the same note, what happens at the moment during SSR? The useMediaQuery hook uses window.matchMedia - does this fall over during server rendering? Storybook just renders on the client doesn't it?

tochwill: We can't know what the width of the page is on server render, so `isSwipeable` will be undefined on the server and then set on the client - so as is in the example @OlgaLyubin mentioned, aria-hidden will not be set. This will also be the case for AMP - so it's something to think about for that too

AlistairGempf:
The tests look good to me. So other than the questions around amp and disabled js, it all looks good! 👍

packages/components/psammead-navigation/src/SwipeableNavigation/index.jsx
DenisHdz: As discussed on Slack, if the `window` is not defined the `useMediaQuery` will return null so `isSwipeable` will be null too, and we won't be applying `aria-hidden: true` and `tabIndex: -1`, therefor it will remain accessible when non-js.

Looking now at AMP.

AlistairGempf: Should this comment be put in the accessibility section of the readme or a comment in the code? It's not an obvious result of the code so might be worth capturing the result.

packages/components/psammead-navigation/hooks/useMediaQuery.js
```diff
@@ -0,0 +1,25 @@
...
+const useMediaQuery = query => {
+  const isClient = typeof window !== 'undefined';
+
+  const mediaMatch = isClient ? window.matchMedia(query) : undefined;
+  const mediaMatches = isClient ? mediaMatch.matches : undefined;
+
+  const [matches, setMatches] = useState(mediaMatches);
+
+  useEffect(() => {
+    if (!isClient) {
```
tochwill: Does useEffect ever run on the server? I believe it's a no-op

DenisHdz: https://github.com/gatsbyjs/gatsby/issues/13947#issuecomment-490992453
https://github.com/gatsbyjs/gatsby/issues/13947#issuecomment-491214724

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -90,6 +80,44 @@ const StyledLink = styled.a`
   }
 `;
 
+const StyledListItem = styled.li`
+  display: inline-block;
+  position: relative;
+  z-index: 2;
+
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
```
tochwill, line 88: Screenshot 2019-11-21 at 14 11 09

This doesn't quite look right - would it be easier to keep the hover/focus states as they are, and just add the extra padding to the the list element instead of the link? Or the whole `<ul>` if that works?

packages/components/psammead-navigation/hooks/useMediaQuery.js
tochwill: That second comment is what I mean, `useEffect` doesn't run on the server so we don't need to check if we're on the client?

packages/components/psammead-navigation/src/index.jsx
DenisHdz: Using `margin` instead of `padding` solves the issue :)

packages/components/psammead-navigation/hooks/useMediaQuery.js
DenisHdz: Yep, but I have to check it at the top level to set `mediaMatch` as undefined when not on the client? Unless I set it as undefined by default.

tochwill: Yeah sorry - so you need that check, it's just this one inside the useEffect hook that I don't think we need :)

packages/components/psammead-navigation/src/index.jsx
tochwill: Ok cool! :) Could you just do
```
&:last-child {
      ${({ dir }) => css`
        margin-${dir === 'ltr' ? 'right' : 'left'}: ${GRADIENT_WIDTH}; 
  `}
}
```
and not have to override StyledLink styles at all?

CHANGELOG.md
```diff
@@ -3,9 +3,10 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 2.0.11 | [PR#2577](https://github.com/bbc/psammead/pull/2577) Add `react-test-renderer` and `@testing-library/react-hooks` |
 | 2.0.10 | [PR#2666](https://github.com/bbc/psammead/pull/2666) Talos - Bump Dependencies - @bbc/psammead-image-placeholder, @bbc/psammead-media-indicator |
 | 2.0.9 | [PR#2657](https://github.com/bbc/psammead/pull/2657) Update codeowners |
-| 2.0.8 | [PR#2663](https://github.com/bbc/psammead/pull/2663) Talos - Bump Dependencies - @bbc/psammead-assets |
+| 2.0.8 | [PR#2663(https://github.com/bbc/psammead/pull/2663) Talos - Bump Dependencies - @bbc/psammead-assets |
```
AlistairGempf:
```suggestion
| 2.0.8 | [PR#2663](https://github.com/bbc/psammead/pull/2663) Talos - Bump Dependencies - @bbc/psammead-assets |
```

AlistairGempf (approved):
I think this all looks good. Good job with this!

packages/components/psammead-navigation/README.md
```diff
@@ -8,6 +8,10 @@ This component is currently tagged as alpha and is not suitable for production u
 
 The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl` and `NavigationLi`. They use `ul` and `li` HTML elements respectively.
 
+## Exports
+
+`/swipeable-navigation` - Makes the navigation to be swipeable under 600px.
```
ghost: Since this is already under the `psammead-navigation` namespace, adding `navigation` to `swipeable` is redundant IMO.
Instead of `psammead-navigation/swipeable-navigation` we would have `psammead-navigation/swipeable`. Any thoughts on this?

packages/components/psammead-navigation/README.md
```diff
@@ -39,52 +44,78 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
 | active   | boolean | No       | `false` | `true`   |
 | currentPageText | string | No | `null`  | `Current page` |
 | service | string | Yes | N/A | `'news'` |
+| dir      | string  | No       | `ltr`   | `rtl` |
```
ghost:
```suggestion
| dir      | string  | No       | `'ltr'`   | `'rtl'` |
```

packages/components/psammead-navigation/README.md
```diff
@@ -39,52 +44,78 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
...
+| isSwipeable | boolean | No | `false` | `true` |
+
+### CanonicalSwipeableNavigation
+
+<!-- prettier-ignore -->
+| Argument | Type | Required | Default | Example |
+| -------- | ---- | -------- | ------- | ------- |
+| children | node | Yes      | N/A     | `<NavigationUl><NavigationLi url="/" script={latin} active="true">Home</NavigationLi><NavigationLi url="/sport" script={latin}>{Sport}</NavigationLi></NavigationUl>` |
+| dir      | string  | No       | `ltr`   | `rtl` |
```
ghost:
```suggestion
| dir      | string  | No       | `'ltr'`   | `'rtl'` |
```

packages/components/psammead-navigation/README.md
```diff
@@ -39,52 +44,78 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
...
+| isSwipeable | boolean | No | `false` | `true` |
+
+### AmpSwipeableNavigation
+
+<!-- prettier-ignore -->
+| Argument | Type | Required | Default | Example |
+| -------- | ---- | -------- | ------- | ------- |
+| children | node | Yes      | N/A     | `<NavigationUl><NavigationLi url="/" script={latin} active="true">Home</NavigationLi><NavigationLi url="/sport" script={latin}>{Sport}</NavigationLi></NavigationUl>` |
+| dir      | string  | No       | `ltr`   | `rtl` |
```
ghost:
```suggestion
| dir      | string  | No       | `'ltr'`   | `'rtl'` |
```

packages/components/psammead-navigation/README.md
```diff
@@ -39,52 +44,78 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
...
-    <NavigationLi url="/weather" script={latin} service="news">
-      {Weather}
-    </NavigationLi>
-  </NavigationUl>
+  <CanonicalSwipeableNavigation>
+    <NavigationUl>
+      <NavigationLi
+        url="/"
+        script={latin}
+        active="true"
```
ghost:
```suggestion
        active
```

packages/components/psammead-navigation/README.md
```diff
@@ -39,52 +44,78 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
...
 
-It includes a ["skip link"](https://www.w3.org/TR/WCAG20-TECHS/G1.html) giving users the option to skip to the main content before the assistive technology reads the full content of the interjection.
-
 We have added the role `list` and `listitem` to the `NavigationUl` and `NavigationList` respectively, due to a VoiceOver bug to reinstate the list semantics.
 
 We have also added visually hidden text to let the user know which item in the Navigation is the current page. Note the use of visually hidden text here is due to lack of support at this time for the aria-current page attribute. Also note the use of `role="text"` to stop splitting in VoiceOver.
 
+In the screen reader UX only the menu button and its content should be available to assistive technology. The exposed swipeable navigation should be hidden. To achieve this we add `aria-hidden:true` to the exposed swipeable navigation so that this is not expose to these users and also add `tabindex=-1` to the links contained within this to remove them from the tab order.
```
ghost: Can we replace the word `exposed` with `visible`?

packages/components/psammead-navigation/hooks/useMediaQuery.js
```diff
@@ -0,0 +1,21 @@
...
+const useMediaQuery = query => {
+  const isClient = typeof window !== 'undefined';
+
+  const mediaMatch = isClient ? window.matchMedia(query) : undefined;
+  const mediaMatches = isClient ? mediaMatch.matches : undefined;
+
+  const [matches, setMatches] = useState(mediaMatches);
+
+  useEffect(() => {
+    const handler = e => setMatches(e.matches);
```
ghost: `e` is not very informative. Is it `event` or `mediaMatch`.

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -109,43 +125,94 @@ const CurrentLink = ({ children: link, script, currentPageText }) => (
...
-      <StyledLink
-        href={url}
-        script={script}
-        currentLink="true"
-        service={service}
-        {...props}
-      >
-        <CurrentLink script={script} currentPageText={currentPageText}>
+}) => {
+  const tabIndex = isSwipeable && { tabIndex: -1 };
```
ghost:
```suggestion
  const tabIndex = isSwipeable &&  -1;
```
`tabIndex` is already defined as a variable. We don't need to create an object with `tabIndex` as a key again IMO.

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -109,43 +125,94 @@ const CurrentLink = ({ children: link, script, currentPageText }) => (
...
+  return (
+    <StyledListItem dir={dir} role="listitem">
+      {active && currentPageText ? (
+        <StyledLink
+          href={url}
+          script={script}
+          service={service}
+          currentLink
+          {...props}
+          {...tabIndex}
```
ghost, line 174:
```suggestion
          tabIndex={tabIndex}
```

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -109,43 +125,94 @@ const CurrentLink = ({ children: link, script, currentPageText }) => (
...
+          <CurrentLink script={script} currentPageText={currentPageText}>
+            {link}
+          </CurrentLink>
+        </StyledLink>
+      ) : (
+        <StyledLink
+          href={url}
+          script={script}
+          service={service}
+          {...tabIndex}
```
ghost, line 186:
```suggestion
          tabIndex={tabIndex}
```

ghost (commented):
This looks really nice.

packages/components/psammead-navigation/README.md
DenisHdz: I think `exposed` is more appropriate in this context because the navigation will be still visible, but not exposed to screen readers.

ghost: Aaah ok. I was misinterpreting what was meant by the statement because of the use of `exposed` and `hidden` in the same sentence. Would it be better to restructure the sentence a little.

```suggestion
In the screen reader UX only the menu button and its content should be available to assistive technology. The swipeable navigation is exposed to assistive technologies and  should be hidden. To achieve this we add `aria-hidden=true` to the swipeable navigation. We also add `tabindex=-1` to the links contained within this to remove them from the tab order.
```

DenisHdz: I renamed it in a different way, hope this is more clear.

Bopchy (approved):
Looks really good

tochwill (approved):

packages/components/psammead-navigation/src/index.test.jsx
```diff
@@ -2,34 +2,62 @@ import React from 'react';
...
+);
+
 describe('Navigation', () => {
+  shouldMatchSnapshot('should render correctly', NavigationExample);
+});
+
+describe('Swipeable Navigation', () => {
+  window.matchMedia = jest.fn().mockImplementation(query => {
+    return {
+      matches: false,
```
ghost:
```suggestion
      matches: true,
```
For the swipeable  navigation should the `matches` evaluate to true?

packages/components/psammead-navigation/hooks/useMediaQuery.test.js
```diff
@@ -0,0 +1,17 @@
...
+        matches: query === '(max-width: 600px)',
+        addListener: jest.fn(),
+        removeListener: jest.fn(),
+      };
+    });
+
+    const { result } = renderHook(() => useMediaQuery('(max-width: 600px)'));
+    expect(result.current).toBe(true);
+  });
+});
```
ghost: This seems to not test the `useMediaQuery` hook. If I delete the `useEffect` block in `useMediaQuery` hook the test still pass.  This test seem to test `useState given an initial state it will return the given value`. For this test to be complete we would need to find a way to trigger a window resize to make use of the `useEffect` block.

packages/components/psammead-navigation/src/index.jsx
ghost: @DenisHdz can we revert this change.  Applying my suggestion leads to console warnings which we have been trying to get rid off in both `psammead` and `simorgh`

Screenshot 2019-11-24 at 16 52 27

DenisHdz: Sure. I initially did it in that way to just add `tabIndex` to the DOM when `isSwipeable` is true.

packages/components/psammead-navigation/hooks/useMediaQuery.test.js
DenisHdz: You are right. Since I didn't find a way to trigger the window resize from the test, what I have done is to put 600px in the mock, so when I call the `useMediaQuery` hook with `(max-width: 600px)` I expect it to return true.

AlistairGempf: We can check that the `addListener` function has been called. We may also be able to check the effects of the function are what we expect. Looking into that as well.

```jsx
    const mockAddListener = jest.fn();
    const mockRemoveListener = jest.fn();

    window.matchMedia = jest.fn().mockImplementation(query => {
      return {
        matches: query === '(max-width: 600px)',
        addListener: mockAddListener,
        removeListener: mockRemoveListener,
      };
    });

    const { result } = renderHook(() => useMediaQuery('(max-width: 600px)'));
    expect(result.current).toBe(true);
    expect(window.matchMedia.addListener).toBeCalled();
```

AlistairGempf: Here's a version of test that also checks the `addListener` function updates the `result.current`. Note that this should probably be at least two separate tests.

```jsx
  it('it should test that the media query matches', async () => {
    const mockAddListener = jest.fn();
    const mockRemoveListener = jest.fn();

    window.matchMedia = jest.fn().mockImplementation(query => {
      return {
        matches: query === '(max-width: 600px)',
        addListener: mockAddListener,
        removeListener: mockRemoveListener,
      };
    });

    const { result } = renderHook(() => useMediaQuery('(max-width: 600px)'));
    expect(result.current).toBe(true);
    expect(mockAddListener).toBeCalled();

    act(() => mockAddListener.mock.calls[0][0]({ matches: false }));
    expect(result.current).toBe(false);
    act(() => mockAddListener.mock.calls[0][0]({ matches: true }));
    expect(result.current).toBe(true);
  }
```

packages/components/psammead-navigation/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 5.0.0-alpha.1 | [PR#2577](https://github.com/bbc/psammead/pull/2577) Add swipeable navigation |
```
greenc05: Can we call this scrollable rather than swipeable please?

packages/components/psammead-navigation/README.md
```diff
@@ -8,6 +8,10 @@ This component is currently tagged as alpha and is not suitable for production u
 
 The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl` and `NavigationLi`. They use `ul` and `li` HTML elements respectively.
 
+## Exports
+
+`/swipeable` - Makes the navigation to be swipeable under 600px.
```
greenc05: Re name to 'scrollable' please - we aren't adding any swiping gestures?

greenc05: Same goes for anywhere else we've used the word swipeable...

packages/components/psammead-navigation/README.md
```diff
@@ -39,52 +44,78 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
...
+    </NavigationUl>
+  </CanonicalSwipeableNavigation>
 </Navigation>;
 ```
 
 The `Navigation` is designed to show a navigation bar on `index` pages, which will show all sections on a site. If there are too many items to fit on one line, the items will wrap to the next lines.
 
+On the other hand, with `CanonicalSwipeableNavigation` or `AmpSwipeableNavigation` we can make the list to remain on one line and to be horizontally swipeable to allow access to further links, under 600px.
```
greenc05: 'horizontally swipeable to' should be 'horizontally scrollable to'

packages/components/psammead-navigation/hooks/useMediaQuery.js
```diff
@@ -0,0 +1,21 @@
...
+  const mediaMatches = isClient ? mediaMatch.matches : undefined;
+
+  const [matches, setMatches] = useState(mediaMatches);
+
+  useEffect(() => {
+    const handler = event => setMatches(event.matches);
+
+    mediaMatch.addListener(handler);
+    return () => mediaMatch.removeListener(handler);
+  });
```
tochwill: Is this going to re-run useEffect after every render? As you're just adding a listener when the component mounts, could you add `[]` as the second argument to useEffect?

tochwill (changes requested):

packages/components/psammead-navigation/README.md
```diff
@@ -39,52 +44,78 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
...
 
-It includes a ["skip link"](https://www.w3.org/TR/WCAG20-TECHS/G1.html) giving users the option to skip to the main content before the assistive technology reads the full content of the interjection.
-
 We have added the role `list` and `listitem` to the `NavigationUl` and `NavigationList` respectively, due to a VoiceOver bug to reinstate the list semantics.
 
 We have also added visually hidden text to let the user know which item in the Navigation is the current page. Note the use of visually hidden text here is due to lack of support at this time for the aria-current page attribute. Also note the use of `role="text"` to stop splitting in VoiceOver.
 
+In the screen reader UX only the menu button and its content should be available to assistive technology, for which reason the swipeable navigation should be hidden. To achieve this we add `aria-hidden:true` to the exposed swipeable navigation so that this is not visible to these users and also add `tabindex=-1` to the links contained within this to remove them from the tab order.
```
greenc05: In the screen reader UX only the menu button and its content should be available to assistive technology, **meaning the scrollable navigation will be hidden**. To achieve this we add `aria-hidden:true` to the exposed **scrollable** navigation so that this is not visible to these users and also add `tabindex=-1` to the links contained within this to remove them from the tab order.

packages/components/psammead-navigation/hooks/useMediaQuery.test.js
DenisHdz: Done. Thanks for the suggestions :)

OlgaLyubin (commented):
There's still a folder named swipeable
Screenshot 2019-11-25 at 15 13 11

packages/components/psammead-navigation/hooks/useMediaQuery.test.js
```diff
@@ -0,0 +1,35 @@
...
+    mockRemoveListener.mockReset();
+  });
+
+  it('it should test that the media query matches', async () => {
+    const { result } = renderHook(() => useMediaQuery('(max-width: 600px)'));
+    expect(result.current).toBe(true);
+    expect(mockAddListener).toBeCalled();
+  });
+
+  it('should add a listener that updates the state', async () => {
```
tochwill: Just wondering why these functions are async?

tochwill (approved):

packages/components/psammead-navigation/src/ScrollableNavigation/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+CanonicalScrollableNavigation.propTypes = {
+  children: node.isRequired,
+  dir: oneOf(['ltr', 'rtl']),
+};
+
+CanonicalScrollableNavigation.defaultProps = { dir: 'ltr' };
+
+export const AmpScrollableNavigation = ({ children, dir }) => (
+  <StyledScrollableNav dir={dir}>{children}</StyledScrollableNav>
```
ghost, line 67: Should we have a `aria-hidden` here?

DenisHdz: No, since we can not use JS in AMP to determine if it should use `aria-hidden` or not, the scrollable nav will be fully available to keyboard users via the tab key and to screen reader users.

ghost (approved):
approved

j-pendlebury (approved):

jamesbrumpton:
lgtm

PR #2582, opened 2019-11-07
add ContentAnchor component

### LINKED ISSUE #2279: Generic component wrapper to avoid content reflow in viewport
We need a component that can load 3rd party content without breaking our coding conventions: https://github.com/bbc/simorgh/blob/latest/docs/Code-Standards.md the particularly troublesome one being content reflow avoidance.

I suggest a generic highly reusable wrapper component which (in a way that meets as best as possible our browser support needs) does not every reflow content that is in a users viewport. To do this the logic is: 
if thing in viewport do not resize
if thing outside of viewport allow resize
    if below viewport do nothing else
    if above viewport adjust user scroll position so the content in the viewport is unchanged.

Broadly speaking this is not too tricky, presuming the css changes to the height of wrapper component and the scroll position are done at the same time. But on Opera mini in particular this is much more challenging.

NB we probably want to add to this component timeouts on the content loading - this does not need to be done for the first iteration but should probably be planned.

See suggestions of how in additional context below.

This would be used for Visual Journalism content, IDT (data) components and adverts. Given that there will be a very large number doing our best to avoid this problem is highly important and urgent as adverts will be blocking releases quite soon (technically they already are).

**Testing notes**
Third party components include ads,visual journalism or IDT content which are wrapped inside an iframe.This issue requires testing the components only on storybook for now.

Dev insight: So much x-browser and x-device testing. I suggest we ensure that content is accessible as a fallback to rendering correctly, i.e. it is better for a thing in the wrapper component to render in a permanently ugly way that to disrupt the reading experience by vertically reflowing.

**Additional context**
NB although this refers to vertical reflowing it should also be enabled for horizontal reflowing.

**_In general I suggest the following logic:_**
Display overflow by default on the wrapper with a scrollbar i.e. it's box size is not define by its children.

Check if thing is in viewport and if it's not whether it's above or below the viewport.

If not in viewport and below it, let it resize and remove the scroll bar.
If not in viewport and above it, resize it while at the same time adjusting the scroll position by the amount changed so there is no perciptible difference.
If it is in the viewport do nothing.

NB you need to accommodate scrolling up and down and entry point on the page that do not presume the user loads the page at the top of the document, e.g. they are linked to an anchor/id lower down the page.

More detail on implementation: the wrapper component should have an optional prop, it default to 0 and may be given any value in px. This value is the min-height of the component. This will be useful when we know for a given use of the wrapper that it will almost always be "N"px high - therefore reserving the space we want it to have - we should have same logic for width.

In case the maths on the change in size of the element is wrong it might be an idea to add a css transition. But if it's correct I'm not sure if this will slowly transition creating a perceived reflow.

**_Resources for how to do this:_**
https://stackoverflow.com/questions/2481350/how-to-get-scrollbar-position-with-javascript

Set min-height and add transition if height changes: https://css-tricks.com/content-jumping-avoid/

Show ad but hide its overflow: https://css-tricks.com/findingfixing-unintended-body-overflow/

Get current things in viewport: https://stackoverflow.com/questions/2481350/how-to-get-scrollbar-position-with-javascript

Scroll to position top https://caniuse.com/#feat=mdn-api_scrolltooptions_top

NB this example which implements a similar but much more simplistic solution to a subset of the problem described https://itnext.io/how-to-stop-content-jumping-when-images-load-7c915e47f576#913e It includes important details such as ensuring it's executed only once.

Check if element is in the viewport: https://gomakethings.com/how-to-test-if-an-element-is-in-the-viewport-with-vanilla-javascript/

Opera mini needs a bookmark to scroll to, so add one to top of element at top of viewport in order to then link, see second answer on https://stackoverflow.com/questions/21174555/trigger-scroll-in-opera-mini oh and you might need this to check what's in the viewport to opera mini https://gomakethings.com/how-to-test-if-an-element-is-in-the-viewport-with-vanilla-javascript/

UA sniffing for edge cases: https://dev.opera.com/articles/fixing-the-scrolltop-bug/

Nice sort of related doc: https://javascript.info/size-and-scroll
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight

jroebu14:
Resolves https://github.com/bbc/psammead/issues/2279

**Overall change:** Adds a component that can wrap content such as ads to detect and mitigate content shifting. For more info refer to the [README.md](https://github.com/bbc/psammead/blob/eb794cb4b18eb3c0f5857bc5e5476f92fbc7c1cc/packages/components/psammead-script-link/README.md)

---

ezgif com-video-to-gif

Above is the Storybook story. To test and showcase the component we have here content that loads a new, different sized image every few seconds. The green shows where the browser repainted sections of the screen as a result of the image changing and causing content reflows. You can see from this example that the content within the wrapper will never resize when in view and hide any overflow of the content within. When it is not in view the wrapper and it's content is allowed to resize and the scroll position is adjusted so the user will not experience any unexpected jumps.

---

**Code changes:**

- Makes use [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) and [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) and dynamically loads polyfills for each if the browser has no support for these APIs
- Polyfills used are:
    - w3c/IntersectionObserver - https://github.com/w3c/IntersectionObserver. 
    - ResizeObserver - https://github.com/juggle/resize-observer
- If [scroll anchoring](https://drafts.csswg.org/css-scroll-anchoring/) isn't supported (IE, Safari) then javascript is used to adjust the scroll position when the content resizes.

**NB**
- **This component needs a better name! Suggestion welcome.**
- Ideas for more test cases inside of Storybook are welcome.
- This is the first time I've used dynamic imports and I haven't seen it used in Simorgh or Psammead so maybe we need to have a discussion about it's use.
- Needs an a11y swarm
- IntersectionObserver is polyfilled whereas ResizeObserver is ponyfilled. There's nothing strategic about this, it's just the way these modules were designed. One view is that [ponyfills should be preferred over polyfills](https://github.com/sindresorhus/ponyfill). I'm not totally convinced but if anyone feels strongly about it I could search for an IntersectionObserver ponyfill
- Writing good unit tests for this component was a challenge. Open to ideas to make them better. I don't think this can be e2e tested until it is implemented in Simorgh(?).
- Unit tests are minimised in the file changed tab. Make sure to expand and have a look.
- Happy to set up a swarm review for this if people feel we need it.

---

**Testing notes**
Needs tested in all browsers. Can be tested in Storybook.


packages/components/psammead-content-shift-blocker/src/index.test.jsx
```diff
@@ -0,0 +1,423 @@
...
+  await render(
+    <ContentShiftBlocker>
+      <Content />
+    </ContentShiftBlocker>,
+  );
+
+  expect(true).toEqual(true); // TODO
+});
+
+xit('should load IntersectionObserver polyfill if not natively supported', async () => {
```
jroebu14: Any ideas on how to test this. Mocking and checking `observe` is called does not work, I think because this module is dynamically loaded.

packages/components/psammead-content-shift-blocker/src/index.test.jsx
```diff
@@ -0,0 +1,423 @@
...
+        },
+      },
+    ]);
+  });
+
+  expect(global.scrollTo).toHaveBeenCalledWith(expectedXPos, expectedYPos);
+  expect(global.pageYOffset).toEqual(1100);
+});
+
+xit('should load ResizeObserver polyfill if not natively suported', async () => {
```
jroebu14: Any ideas on how to test this. Mocking and checking `observe` is called does not work, I think because this module is dynamically loaded.

ryanmccombe:
This is great - I thought the performance of this would be terrible on a lot of the devices our audience uses, but the polyfill idea is great.

Looks like most people will have native support for this, so no scroll listeners at all

packages/components/psammead-content-shift-blocker/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "homepage": "https://github.com/BBC-News/psammead/blob/latest/packages/components/psammead-content-shift-blocker/README.md",
+  "dependencies": {
+    "@juggle/resize-observer": "^2.4.0",
+    "intersection-observer": "^0.7.0"
+  },
+  "peerDependencies": {
+    "react": "^16.11.0"
+  },
+  "devDependencies": {
+    "@bbc/psammead-headings": "^3.1.18"
```
sareh: You shouldn't need to have this as a devDependency, since Psammead's tests & Storybook have all the psammead packages available already.

packages/components/psammead-content-shift-blocker/src/index.jsx
```diff
@@ -0,0 +1,159 @@
...
+const initResizeObserver = ({ wrapperEl, setContentElRect }) => {
+  const init = ResizeObserver => {
+    const callback = ([contentEntry]) => {
+      setContentElRect(contentEntry.contentRect);
+    };
+    const RO = new ResizeObserver(callback);
+
+    RO.observe(
+      wrapperEl.current
+        .firstChild /* will break if using a fragment :thinking: */,
```
sareh: Would be good to explicitly mention in the Readme that this component does not support passing in a Fragment wrapping several children, but that it should be e.g. a `div`.

packages/components/psammead-content-shift-blocker/src/index.jsx
```diff
@@ -0,0 +1,159 @@
...
+    // wrapper did resize
+    if (scrollAnchoringIsSupported.current || scrollHeight.current === null) {
+      return;
+    }
+
+    const { scrollHeight: newScrollHeight } = document.body;
+    const { current: prevScrollHeight } = scrollHeight;
+    const scrollHeightDiff = prevScrollHeight - newScrollHeight;
+
+    if (scrollHeightDiff) {
```
sareh:
```suggestion
    if (scrollHeightDiff !== 0) {
```

packages/components/psammead-content-shift-blocker/package.json
jroebu14: @sareh storybook build fails without this as a dev dep 😕

packages/components/psammead-content-shift-blocker/src/index.stories.jsx
```diff
@@ -0,0 +1,173 @@
...
+          <Gallery delay={8000} script={script} service={service} />
+          <Gallery delay={2000} script={script} service={service} />
+          <Gallery delay={6000} script={script} service={service} />
+          <Gallery delay={8000} script={script} service={service} />
+          <Gallery delay={3000} script={script} service={service} />
+          <Gallery delay={2000} script={script} service={service} />
+        </Wrapper>
+      );
+    },
+    { notes },
```
sareh: These Storybook stories will be run with Chromatic QA our visual regression checker. Whenever the content changes it will show a false negative, stating that the content has changed. 
Can you please add this to the config (next to notes) to disable chromatic on these stories `chromatic: { disable: true },`.

packages/components/psammead-content-shift-blocker/README.md
```diff
@@ -0,0 +1,70 @@
...
+```jsx
+import ContentShiftBlocker from  '@bbc/psammead-content-shift-blocker';
+
+<ContentShiftBlocker
+  initialHeight={400}
+  initialWidth={400}
+>
+  <Advertisement>
+</ContentShiftBlocker>
+```
```
sareh: In a note above, you've recommended this:
>I would even recommend you explicitly set `overflow-anchor: auto` on the body element in case browser vendors decide to make this feature opt-in rather than opt-out


It'd be good to have a usage example here in this package. Perhaps even a styled-component-based export, so that this component can be used immediately in a page. 

I'd like to note that when carrying out the accessibility review of this component, we should also look at the impact that `overflow-anchor: auto` has on the page.

packages/components/psammead-content-shift-blocker/src/index.test.jsx
```diff
@@ -0,0 +1,398 @@
+/* eslint-disable no-underscore-dangle */
+
```
sareh: Brilliant work with writing tests for all the different cases. 💯

packages/components/psammead-content-shift-blocker/src/index.stories.jsx
```diff
@@ -0,0 +1,173 @@
...
+          {renderParagraphs({ script, service })}
+          {renderParagraphs({ script, service })}
+          {renderParagraphs({ script, service })}
+        </Wrapper>
+      );
+    },
+    { notes },
+  )
+  .add(
+    'nothing but cats',
```
sareh: I've only been able to replicate this bug in Firefox at the moment - when viewing this story, scrolling down half-way and waiting for about 30 seconds, something seems to be going out of sync, because the boxes jump around the page.

Gif of 'nothing but cats' story with large jumps in the page

sareh (commented):
Brilliant work on this component! I've just added a few comments below.

packages/components/psammead-content-shift-blocker/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Description
+
+`ContentShiftBlocker` is a component to mitigate the jarring user experience of the page moving while the user is in the middle of consuming the content.
+
+> Today, users of the web are often distracted by content moving around due to changes that occur outside the viewport. Examples include script inserting an iframe containing an ad, or non-sized images loading on a slow network. - [CSS Working Group](https://drafts.csswg.org/css-scroll-anchoring/#intro*)
+
+Such content can be wrapped with `ContentShiftBlocker` to prevent the page from jumping around.
+
+⚠️ For browsers that support [scroll-anchoring](https://drafts.csswg.org/css-scroll-anchoring/) with `overflow-anchor`, `ContentShiftBlocker` assumes you have not disabled it by explicitly setting `overflow-anchor: none` on any parent elements. If you have then the user may experience content shifting. I would even recommend you explicitly set `overflow-anchor: auto` on the body element in case browser vendors decide to make this feature opt-in rather than opt-out.
```
andrewscfc: I have no idea how to do this, but could we add a lint rule that makes that mandatory if we agree it is the right thing to do?

packages/components/psammead-content-shift-blocker/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Props
+
+| Argument      | Type          | Required | Default | Example         |
+| ------------- | ------------- | -------- | ------- | --------------- |
+| children      | node          | true     |         | <Advertisement> |
+| initialHeight | String/Number | false    | 'auto'  | 400             |
+| initialWidth  | String/Number | false    | 'auto'  | 400             |
+
+**Note** - `children` does not support passing in a React `Fragment` wrapping several children. Should be e.g. a div.
```
andrewscfc: Another area I _think_ you could control with a lint rule, maybe overkill though, a bit more subtle than the `overflow-anchor` example above.

packages/components/psammead-content-shift-blocker/src/index.jsx
```diff
@@ -0,0 +1,160 @@
...
+  const scrollAnchoringIsSupported = useRef(null);
+  const [wrapperIO, setWrapperIO] = useState({});
+  const [contentElRect, setContentElRect] = useState({});
+  const [wrapperDimensions, setWrapperDimensions] = useState({
+    height: initialHeight,
+    width: initialWidth,
+  });
+
+  useEffect(() => {
+    // component did mount
```
andrewscfc: to be accurate I think it would be 
```suggestion
    // component did mount/update
```
Unsure if the comments are needed?

packages/components/psammead-content-shift-blocker/src/index.jsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    const { scrollHeight: newScrollHeight } = document.body;
+    const { current: prevScrollHeight } = scrollHeight;
+    const scrollHeightDiff = prevScrollHeight - newScrollHeight;
+
+    if (scrollHeightDiff !== 0) {
+      // adjust scrollY position to prevent visible jump
+      window.scrollTo(0, window.pageYOffset - scrollHeightDiff);
+    }
```
andrewscfc: My understanding of this section is that it is an alternative where browsers do not support scroll anchoring, modifying the scroll position appropriately.

IMO, this code could be a seperate module and unit tested in isolation. It could be a pure function taking the relevant values and computing the expected scroll position, decoupling from the browser APIs. I think this would make it easier to read and maybe decrease the amount of tests you have to write at a higher level? I think I'm arguing for the traditional 'testing pyramid' while approach taken here maybe a bit more [testing trophy](https://kentcdodds.com/blog/write-tests) .

packages/components/psammead-content-shift-blocker/src/index.stories.jsx
```diff
@@ -0,0 +1,173 @@
...
+      setSizes(`${getRandomInt()}/${getRandomInt()}`);
+    }, delay);
+
+    return () => clearInterval(interval);
+  }, []);
+
+  return (
+    <StyledImage
+      src={`https://placekitten.com/${sizes}`}
+      alt="Naughty cats >:("
```
andrewscfc: As I mentioned in the swarm review, I think we'll need a stable source of test data rather than random cats.

That makes me sad :(:(:(:(:(:(:(:

packages/components/psammead-content-shift-blocker/src/index.stories.jsx
andrewscfc: Be nice if we could include at least one test in chromatic? Could we render and then pretty much instantly change the content size, basically snapshotting that nothing has moved about. I think I'm missing the need for the delays personally?

packages/components/psammead-content-shift-blocker/src/index.test.jsx
```diff
@@ -0,0 +1,398 @@
...
+  );
+
+  // check initial size
+  expect(
+    getComputedStyle(container.firstChild).getPropertyValue('width'),
+  ).toEqual(`${initialSize}px`);
+
+  expect(
+    getComputedStyle(container.firstChild).getPropertyValue('height'),
+  ).toEqual(`${initialSize}px`);
```
andrewscfc: I am a big advocate of AAA (Arrange Act Assert) in tests (I think I mentioned it to you (@jroebu14) in a call at one point?). In this test we:
- Setup the component (Arrange)
- Render it (Act)
- Check its initial size (Assert)
- Simulate an intersection (Act)
- Simulate a resize (Act)
- Check the resize did not occur on the wrapper (Assert)

I'd consider seperating this section into a seperate test for the initial size:
- Setup the component (Arrange)
- Render it (Act)
- Check its initial size (Assert)

then another test for suppressing the resize:
- Setup the component (Arrange)
- Render it (Act)
- Simulate an intersection (Act)
- Simulate a resize (Act)
- Check the resize did not occur on the wrapper (Assert)

The idea of it is to make sure tests only fail because one aspect of a component has broken and not others. Also assertions only check the 'end result' not that the setup (Arrange) has been successful. I hope that make sense, please feel free to disagree but fancy putting this out there being new and being a fan of this practice!

andrewscfc (commented):
Sorry if I'm sounding a bit preachy on this review, just dropping a few approaches to consider for discussion. Awesome PR, very complicated and very compact considering!

packages/components/psammead-content-shift-blocker/src/index.test.jsx
andrewscfc: Now this set of tests is a good advocate for the testing trophy approach re my previous comment: https://github.com/bbc/psammead/pull/2582#discussion_r350858961 

Reading them carefully, I can see clearly what they are simulating in the browser, I'm disagreeing with myself :anguished: Either way, these tests are cool :sunglasses:

packages/components/psammead-content-shift-blocker/src/index.jsx
jroebu14: @andrewscfc because the dependencies are specified and they are empty (empty array as 2nd argument) then this useEffect will only ever be called after the first render. If the empty array was not passed in then it would be component did mount/update event. See https://dev.to/trentyang/replace-lifecycle-with-hooks-in-react-3d4n for a better explanation

packages/components/psammead-content-shift-blocker/src/index.jsx
jroebu14: @andrewscfc I'm not sure how to make this a pure function when it relies heavily on state, refs and the window object outside of it's own scope. Maybe we need to have a chat about what you're thinking.

packages/components/psammead-content-shift-blocker/src/index.stories.jsx
jroebu14: I agree with Sareh on this one. I don't see the value in a ChromaticQA test especially because there is no visible UI and we have unit tests for the functionality. ChromaticQA has in the past flagged up IE javascript errors so if we think there's value in that I could render a div with a background colour that this component wraps as a ChromaticQA test

packages/components/psammead-content-shift-blocker/src/index.jsx
andrewscfc: @jroebu14, here's what I was thinking:

```javascript
    const { scrollHeight: newScrollHeight } = document.body;
    const { current: prevScrollHeight } = scrollHeight;
    const { pageYOffset : scrollY } = window;

    const scrollHeight = calculateScrollY({ newScrollHeight, prevScrollHeight, scrollY })
    // adjust scrollY position to prevent visible jump
    if(scrollHeight) {
      window.scrollTo(0, scrollHeight);
    }

    const calculateScrollY = ({ newScrollHeight, prevScrollHeight, scrollY }) => {
      scrollHeightDiff = prevScrollHeight - newScrollHeight;
      
      if(scrollHeightDiff !== 0) {
        return scrollY - scrollHeightDiff;
      }
      return null;
    }
```
The pure function is `calculateScrollY()` but maybe it's a bit to simple to merit its own module, when I looked over your tests carefully I was more comfortable: https://github.com/bbc/psammead/pull/2582/files#r350880186

packages/components/psammead-content-shift-blocker/README.md
jroebu14: @sareh I've added `overflow-anchor: auto` as  global style and [updated the readme ](https://github.com/bbc/psammead/pull/2582/files#diff-9a017ca6f78e0d80149523c2520305d8R11)

packages/components/psammead-content-shift-blocker/README.md
jroebu14: I've also no idea how to do this. It's a good idea but may take . me some time to figure out and seems low priority. I can create an issue that we could pick up as improvement when we have the time.

packages/components/psammead-content-shift-blocker/src/index.jsx
jroebu14: Nice idea. I've updated the code. Thanks!

packages/components/psammead-content-shift-blocker/src/index.stories.jsx
jroebu14: It was painful but I got rid of the dogs and cats and swapped out doggo ipsum for our example text provided by the services knob. For the images I went with my hero Eliud Kipchoge 🏃🏿🇰🇪❤️

packages/components/psammead-content-shift-blocker/src/index.test.jsx
jroebu14: Great suggestion and thanks for educating me on this best practice 👍

packages/components/psammead-content-shift-blocker/src/index.stories.jsx
andrewscfc: I'm happy with this as you say @jroebu14 given there are unit tests

andrewscfc (commented):
I'd be happy to approve this once this bug is resolved: https://github.com/bbc/psammead/pull/2582#discussion_r350148391 

Great work 👍

packages/components/psammead-content-shift-blocker/src/index.stories.jsx
jroebu14: @sareh I changed the scroll position adjustment to happen in `useLayoutEffect` instead of `useEffect` and that has fixed this issue.

jroebu14:
thanks @andrewscfc see https://github.com/bbc/psammead/pull/2582#discussion_r351887327 and feel free to check out the branch and verify!

andrewscfc (approved):
Approved following: https://github.com/bbc/psammead/pull/2582#issuecomment-559567459

jroebu14:
@sareh I think all I have to do now is think of a better name for this component!

sareh (approved):
🎉 Great work on this PR, @jroebu14! Thanks for making the updates & addressing the bug. The changes look great.

sareh:
Just one point that's erroring when running on CI -
```
packages/components/psammead-content-anchor/src/index.jsx

 56:9  ✖  This experimental syntax requires enabling the parser plugin: 'dynamicImport' (56:9)
```

PriyaKR:
@jroebu14 can we have some examples in psammead storybook for a content which is inside iframe and uses the content wrapper when displayed?

For eg on this news page there is a quiz inside an iframe https://www.bbc.co.uk/news/world-50377297

PriyaKR:
@jroebu14 Looks good to me chrome,FF, safari. Would be able to cover devices when this is implemented in simorgh. Also on IE browsers IE edge,11,10 and 9 the scroll appears different compared to other browsers.

Also can we have some examples in psammead storybook for a content which is inside iframe and uses the content wrapper when displayed?

For eg on this news page there is a quiz inside an iframe https://www.bbc.co.uk/news/world-50377297

PriyaKR:
Also one nice to have ie)at smaller viewports the horizontal scroll is not visible until the user scrolls right or left which makes the reading experience feel as if the content is truncated.

HarveyPeachey:
I'm testing this component out with `psammead-media-player` to see the potential of solving reflow issues https://github.com/bbc/simorgh/issues/2454 and I've pulled it into Simorgh to test. The component is working fine however I'm getting a warning about `useLayoutEffect`: 
```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's ou
tput format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useL
ayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayoute
ffect-ssr for common fixes.
```

PR #2611, opened 2019-11-12
Change yeoman generator

### LINKED ISSUE #2188: Give ability to choose what package is being created with `npm run create:package`
At the moment when you create a brand new package for psammead using the command `npm run create:package` which uses a generator called [yeoman](https://yeoman.io/), it generates all the necessary files to start work on a new package. However, it automatically assumes you're creating a `component` as it places it in the `packages/components` directory.

We need to add something to our yeoman configuration to ask what package we're creating between `components`, `containers` and `utilities` which places the generated files into their corresponding directories. 

Depending on which package is being created the default `index.stories.jsx` should reflect this as well e.g the [`Psammead Locales`](https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-locales/numerals/index.stories.jsx) is added to the utilities subsection in storybook whereas by default it's adding it to the `other` section by default. The URL in the default `CHANGELOG.MD` is incorrect it should be `https://github.com/bbc/psammead/pull/XX`

We could just let the user type the directory they wish the package to go to, after generation.

**Testing notes**
Doesn't require any testing from a tester.

**Additional context**
The generated `index.stories.jsx` has linting issues which makes sense to be addressed as part of this PR as well:
Screenshot 2019-09-19 at 15 19 58

oluoluoxenfree:
Resolves #2188 

**Overall change:** _Used yeoman's list functionality to let the user choose a template kind/type._

**Code changes:**

- _Specified the destination folder, based on the selected template kind/type._

---


mykeels:
Went with yeoman's list functionality.

image

chris-hinds (commented):
should this be including new templates for the new package types?

mykeels:
@hindsc52 I've created [this issue](https://github.com/bbc/psammead/issues/2642), and its linked PR to track adding unique templates for package types.

hotinglok (changes requested):
It seems like the condition to stop people from creating packages of the same name doesn't work the same way for `containers` and `utilities` as it does for `components`. In components, a prompt stops you immediately if you enter a package with the same name as an existing one whereas for the new templates, it lets you carry on with the whole process only asking at the end if you would like to overwrite the existing package. I think it should be updated for consistency.

mykeels:
@lokh01 this should work now.

.yeoman/src/get-existing/index.js
```diff
@@ -1,10 +1,17 @@
...
+}
+
+const getExistingPackages = () => {
+  return getPackageTypes()
+          .map(getPackagesFromFolder)
+          .reduce((arr, item) => arr.concat(item))
 };
 
-module.exports = getExistingPackages;
+module.exports  = getExistingPackages;
```
hotinglok:
```suggestion
module.exports = getExistingPackages;
```

.yeoman/src/get-existing/index.js
```diff
@@ -1,10 +1,17 @@
...
+  const path = `./packages/${folder}`;
   return fs
     .readdirSync(path)
     .filter(file => fs.statSync(`${path}/${file}`).isDirectory());
+}
+
+const getExistingPackages = () => {
+  return getPackageTypes()
+          .map(getPackagesFromFolder)
+          .reduce((arr, item) => arr.concat(item))
```
hotinglok: Remove extra tab space

hotinglok (approved):
Besides linting issues, the functionality is fine now so it looks good to me on that front. Just need to clean up some of the files.

.yeoman/src/get-existing/index.js
mykeels: Fixed!

mykeels:
Fixed linting issues.

j-pendlebury (approved):
LGTM 👍

PR #2302, opened 2019-10-03
DO NOT MERGE - Scrolling navigation with menu prototype

AlistairGempf:
Resolves #2291

**Overall change:** Prevents wrapping of the navigation bar, instead having it horizontally scroll, with a menu dropdown.

**Code changes:**

- Set navigation bar to scroll horizontally on overflow-x
- Add a Menu button styled link
- Display a menu with the same links as in the navigation bar
- Update stories to have knobs for different types of prototype
  - Has a prototype for a grid menu
  - A single column menu
  - And a menu that takes the full length of the screen
    - This is broken with the Brand, but is useful to show that it doesn't work
- Adds an amp parameter to the `Navigation` component for the showing and hiding of the menu in amp

---

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Automated jest tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

PR #2613, opened 2019-11-12
Create dropdown navigation

### LINKED ISSUE #2557: Navigation: Create dropdown and toggle components
The new navigation designs contain a burger menu icon, which opens a dropdown of navigation items when selected.

See burger menu in navigation in the designs in zeplin:
Screenshot 2019-11-04 at 17 19 57

Once the burger menu has been interacted with, the dropdown menu is presented:
Screenshot 2019-11-04 at 17 20 51

(See zeplin for margin + padding values)

Right to left designs:
Screenshot 2019-11-04 at 17 21 47

Screenshot 2019-11-04 at 17 22 15

- Add burger menu and cross icons to psammead assets.

- Currently this contains the same list of nav items as the scrollable navigation, but this will not always be the case, so the navigation dropdown needs to accept a separate list than the scrollable nav.

- This dropdown menu pushes the rest of the page content down, it does not overlay on top of it (just need to build the dropdown in psammead).

When the user interacts with the cross icon, the dropdown menu is closed. The interaction will need to be handled differently between Canonical and AMP pages - the psammead package should export presentational components that can be composed into the navigation, and controlled when integrated.

We can reuse the presentational components between AMP and Canonical exports, which accept event handlers in a format such as this (this is not meant to represent the final component design, just to show how to accept AMP and Canonical actions):

```
export AmpDropdown = ({ closeAction, children, etc... }) => (
    <DropdownWrapper>
        <button on={closeAction}><CloseIcon /></button>
        <DropdownList>
            {children}
        </DropdownList>
    </DropdownWrapper>
); 
```

```
export CanonicalDropdown = ({ onClose, children, etc... }) => (
    <DropdownWrapper>
        <button onClick={onClose}><CloseIcon /></button>
        <DropdownList>
            {children}
        </DropdownList>
    </DropdownWrapper>
); 
```

Similary for the burger menu:

```
export AmpBurger = ({ action, etc... }) => (
    <button on={action}><BurgerIcon /></button>
); 
```

```
export CanonicalBurger = ({ onClick, etc... }) => (
    <button onClick={onClick}><BurgerIcon /></button>
); 
```

(There may be a preferable name for this component than 'Burger' 🍔 )
- The burger menu and dropdown are only rendered < 600px.

- Screenreader UX can be found in zeplin: https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/5db9725f3cab221121ebab72

**Testing notes**
Storybook testing in psammead

**Additional context**

OlgaLyubin:
Resolves #2557 

**Overall change:** 
Implemented new navigation designs that contain a burger menu button which opens a dropdown of navigation items when selected. For now these are just presentational components that will be exported to Simorgh (where the logic for interaction will be implemented).

Toggling for `AmpMenuButton` now works in storybook (keep in mind that you might need to refresh the page because of the AMP related issue that exists in storybook for now). `AmpMenuButton` uses `expandedHandler` to update the `expanded` value of a state and depending on the value changes the icon to be rendered (hamburger or cross).

**Code changes:**
- Imported Navigation icons from `psammead-assets`
- Made hamburger and cross buttons
- Added a dropdown navigation for amp and canonical versions
- Added snapshot testing for dropdown navigation
---


packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,226 @@
...
+
+const HamburgerButton = styled.button`
+  width: 2.75rem;
+  height: 2.75rem;
+  background-color: ${C_POSTBOX};
+  padding: 0;
+  margin: 0;
+  border: 0;
+
+  &:hover,
```
ghost: Do we need to have a `hover` styling  and `cursor` considering this will be only visible on mobile devices?

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,226 @@
...
+  background-color: ${C_EBON};
+  width: 2.75rem;
+  height: 2.75rem;
+  padding: 0;
+  margin: 0;
+  border: 0;
+
+  &:hover,
+  &:focus {
+    box-shadow: inset 0 0 0 0.25rem ${C_WHITE};
```
ghost: Why are you not using `border: 0.25rem solid ${C_WHITE}`.

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,226 @@
...
+  GEL_GROUP_2_SCREEN_WIDTH_MAX,
+  GEL_GROUP_3_SCREEN_WIDTH_MIN,
+} from '@bbc/gel-foundations/breakpoints';
+import { getPica } from '@bbc/gel-foundations/typography';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+
+const DropdownWrapper = styled.div`
+  background-color: ${C_EBON};
+  width: 100vw;
```
ghost: Do we need to specify the width?

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,226 @@
...
+DropdownNavigationLi.defaultProps = {
+  active: false,
+  currentPageText: null,
+  dir: 'ltr',
+};
+
+AmpDropdown.propTypes = {
+  announcedText: string.isRequired,
+  children: node.isRequired,
+  closeAction: func.isRequired,
```
ghost: Should we rename this to `onClose`

packages/components/psammead-navigation/src/Dropdown/index.jsx
ghost: I have seen why.

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,226 @@
...
+`;
+
+const StyledCurrentLink = styled.span`
+  ${({ dir }) =>
+    dir === 'ltr'
+      ? `border-left: ${GEL_SPACING_HLF} solid ${C_WHITE};
+        padding-left: ${GEL_SPACING};`
+      : `border-right: ${GEL_SPACING_HLF} solid ${C_WHITE};
+        padding-right: ${GEL_SPACING};`}
+`;
```
ghost: What do you think about making this to.
```js
const getStyles = dir => {
  const direction = dir === 'ltr' ? 'left' : 'right';
  return `border-${direction}: /*style*/ ;padding-${direction}: /*style*/`;
}

// and then use it as
getStyle(dir)
```

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,226 @@
...
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
+    display: inline-block;
+  }
+`;
+
+export const AmpHamburgerMenu = ({ announcedText, openAction }) => (
+  <HamburgerButton
+    aria-label={announcedText}
+    on={openAction}
+    aria-expanded={false}
```
ghost: This means that even if the `HambergerButton` is expanded on `openAction` `aria-expanded` would remain `false`. This should be a prop to allow control outside the component.

ghost (changes requested):
We should be testing if the `openAction` and `closeAction` are called and also check the state of `aria-expanded`.

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: Zeplin specified styling for `hover`&`focus` so I went along with it, but I guess since it's for mobile devices we don't need it. Will delete those now.

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: Works without it, will delete this too

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: My understanding of this was that if the `HamburgerButton` is expanded, then it is no longer visible on the page and instead of it the `CrossButton` is rendered with the dropdown menu. And a `CrossButton` has `aria-expanded=true`.

ghost: That makes sense, in that case instead of passing the `prop` to `aria-expanded` you would pass it to `HamburgerButton` or the `Wrapper` to hide it, if that state is true. I guess that will be handled in https://github.com/bbc/simorgh/issues/4571

OlgaLyubin:
> We should be testing if the `openAction` and `closeAction` are called and also check the state of `aria-expanded`.

Can we test those in Psammead though? Since all the logic for `openAction` and `closeAction` is going to be written later in Simorgh

ghost:
> Can we test those in Psammead though?

Yes you can. You can create a mock function and assert that it was called. That is likely outside the scope of this PR though and should be done while resolving bbc/simorgh#4571

packages/components/psammead-navigation/README.md
```diff
@@ -40,7 +40,37 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
...
+| currentPageText | string | No | `null`  | `Current page` |
+| dir      | string  | No       | `ltr`   | `rtl` |
+
+### CanonicalHamburgerMenu
+
+<!-- prettier-ignore -->
+| Argument | Type | Required | Default | Example |
+| -------- | ---- | -------- | ------- | ------- |
+| announcedText      | string  | Yes       | N/A   | `'Menu'` |
+| onOpen      | function  | Yes       | N/A   |  |
```
ghost: Can we have consistent use of space within the tables
i.e `one space between text and |`

HarveyPeachey (commented):
LGTM, this will need a dev a11y swarm

packages/components/psammead-navigation/README.md
```diff
@@ -40,7 +40,37 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
...
+| currentPageText | string | No | `null` | `Current page` |
+| dir | string | No | `ltr` | `rtl` |
+
+### CanonicalHamburgerMenu
+
+<!-- prettier-ignore -->
+| Argument | Type | Required | Default | Example |
+| -------- | ---- | -------- | ------- | ------- |
+| announcedText | string | Yes | N/A | `'Menu'` |
+| onOpen | function | Yes | N/A |  |
```
OlgaLyubin: Not really sure about what kind of function to put as an example of an `onOpen` argument here. Any suggestions?

ghost: I would put something like `() => console.log('handle open action)`

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,216 @@
...
+  &:focus {
+    box-shadow: inset 0 0 0 0.25rem ${C_WHITE};
+  }
+
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    display: none;
+  }
+
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
+    display: inline-block;
```
AlistairGempf: I think this also needs a `float: left;` to prevent this happening on Safari.

image

OlgaLyubin: Thank you, fixed!

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -159,6 +170,7 @@ const StyledNav = styled.nav`
 
 const Navigation = ({ children, dir }) => (
   <StyledNav role="navigation" dir={dir}>
+    <CanonicalHamburgerMenu announcedText="Menu" onOpen={() => {}} />
```
tochwill: This PR looks really good! Only thing is I don't think we want to compose the menu into this export (or have one 'Navigation' export anymore in fact) - I think we'll want to compose the navigation together in simorgh? We can then create the `onOpen` function in Simorgh and pass it into the hamburger menu, and use the AMP or Canonical specific components as required.

packages/components/psammead-navigation/src/Dropdown/index.jsx
AlistairGempf: Just looked at the rtl design and looks like it needs to be based on `dir`, sorry!

```jsx
float: ${({ dir }) => dir === 'ltr' ? 'left' : 'right'};
```

packages/components/psammead-navigation/src/index.jsx
OlgaLyubin: That's a good point to consider. Though if we decide to compose navigation in Simorgh, then we would need to export a lot of components here and it might get a bit messy?

With regards to Canonical/AMP, if we decided to leave the Navigation as is, we could pass a prop and depending on it render canonical or amp burger menu.

Also, if we compose the nav in Simorgh, this is something that will affect @DenisHdz too for the swipeable navigation.

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: :see_no_evil: Always forgetting the rtl

packages/components/psammead-navigation/src/index.jsx
tochwill: Let's chat today after standup with @DenisHdz and make sure we're all on the same page :)

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,217 @@
...
+  </DropdownWrapper>
+);
+
+export const CanonicalDropdown = ({ announcedText, onClose, children }) => (
+  <DropdownWrapper>
+    <CrossButton
+      aria-label={announcedText}
+      onClick={onClose}
+      // eslint-disable-next-line react/jsx-boolean-value
+      aria-expanded={true}
```
tochwill: Quick question about using 'true' and 'false' with aria tags - should it be `aria-expanded` (a boolean attribute as it is here) or `aria-expanded="true"` (an attribute with 'true' as a string)?

tochwill: I'm just wondering as some properties appear to be able to be true/false/undefined

OlgaLyubin: Good point. On [React Accessibility](https://reactjs.org/docs/accessibility.html) page they use strings with any `aria` attributes, so I guess that is the correct option. Will change it now

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,217 @@
...
+
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    display: none;
+  }
+`;
+
+const CrossButton = styled.button`
+  background-color: ${C_EBON};
+  width: 2.75rem;
+  height: 2.75rem;
```
AlistairGempf: Might be good to put this into a constant at the top of the file, since it's also used in the hamburger button.

packages/components/psammead-navigation/src/Dropdown/index.test.jsx
```diff
@@ -0,0 +1,41 @@
...
+  CanonicalDropdown,
+  DropdownNavigationLi,
+  CanonicalHamburgerMenu,
+} from './index';
+import pidginNavData from '../../testHelpers/pidgin';
+
+describe('Navigation', () => {
+  shouldMatchSnapshot(
+    'dropdown should render correctly',
+    <CanonicalDropdown announcedText="Menu" onClose={() => {}}>
```
AlistairGempf: Should probably have a similar test for `AmpDropdown`.

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: Done :)

packages/components/psammead-navigation/src/Dropdown/index.test.jsx
OlgaLyubin: The thing is that for now the only difference between `CanonicalDropdown` and an `AmpDropdown` is `on={onClose}` vs `onClick={onClose}` attribute. And this difference is not registered in a snapshot since it does not impact the appearance of the component. If I add a test for `AmpDropdown`, the snapshot will be exactly the same as the one I currently have for `CanonicalDropdown`. So it would not bring much value.

AlistairGempf: Yes, good point!

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+const CrossButton = styled.button`
+  background-color: ${C_EBON};
+  width: ${MENU_ICON_SIDE_LENGTH};
+  height: ${MENU_ICON_SIDE_LENGTH};
+  padding: 0;
+  margin: 0;
+  border: 0;
+
+  &:focus {
+    box-shadow: inset 0 0 0 0.25rem ${C_WHITE};
```
Bopchy:
```suggestion
    box-shadow: inset 0 0 0 ${GEL_SPACING_HLF} ${C_WHITE};
```

Bopchy: Looks really good 👍. Consider using `GEL_SPACING_HLF` for consistency

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+const HamburgerButton = styled.button`
+  width: ${MENU_ICON_SIDE_LENGTH};
+  height: ${MENU_ICON_SIDE_LENGTH};
+  background-color: ${C_POSTBOX};
+  padding: 0;
+  margin: 0;
+  border: 0;
+
+  &:focus {
+    box-shadow: inset 0 0 0 0.25rem ${C_WHITE};
```
Bopchy:
```suggestion
    box-shadow: inset 0 0 0 ${GEL_SPACING_HLF} ${C_WHITE};
```

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+DropdownNavigationLi.defaultProps = {
+  active: false,
+  currentPageText: null,
+  dir: 'ltr',
+};
+
+AmpDropdown.propTypes = {
+  announcedText: string.isRequired,
+  children: node.isRequired,
+  onClose: func.isRequired,
```
AlistairGempf: The amp version will be a string rather than a function (see https://github.com/bbc/psammead/pull/2302/files#diff-649fabcf7d0f0c3dc0bfa5b342f106b2R431)
```suggestion
  onClose: string.isRequired,
```

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+
+CanonicalDropdown.propTypes = {
+  announcedText: string.isRequired,
+  children: node.isRequired,
+  onClose: func.isRequired,
+};
+
+AmpHamburgerMenu.propTypes = {
+  announcedText: string.isRequired,
+  onOpen: func.isRequired,
```
AlistairGempf: The amp version will be a string rather than a function (see https://github.com/bbc/psammead/pull/2302/files#diff-649fabcf7d0f0c3dc0bfa5b342f106b2R431)
```suggestion
  onClose: string.isRequired,
```

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: Good point, thank you!

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+  <HamburgerButton
+    aria-label={announcedText}
+    onClick={onOpen}
+    aria-expanded="false"
+  >
+    {navigationIcons.hamburger}
+  </HamburgerButton>
+);
+
+DropdownNavigationLi.propTypes = {
```
DenisHdz: Lately, we have been putting the `propTypes` after each component declaration. I am happy either way but it would be nice to keep it consistent.

packages/components/psammead-navigation/README.md
```diff
@@ -71,6 +101,30 @@ import { latin } from '@bbc/gel-foundations/scripts';
 </Navigation>;
 ```
 
+## Dropdown Navigation Usage
+
+```jsx
+import React from 'react';
+import { CanonicalDropdown, DropdownNavigationLi } from '@bbc/psammead-navigation';
```
AlistairGempf: Is this now in the Dropdown folder? I wonder if it's worth importign and re-exporting from the top level index.js anyway.
```suggestion
import { CanonicalDropdown, DropdownNavigationLi } from '@bbc/psammead-navigation/Dropdown';
```

packages/components/psammead-navigation/README.md
```diff
@@ -40,7 +40,37 @@ The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl
 | currentPageText | string | No | `null`  | `Current page` |
 | service | string | Yes | N/A | `'news'` |
 
-## Usage
+### CanonicalDropdown
```
AlistairGempf: There should be an amp example of this and the hamburger since the `onClose` for amp is a string so slightly different.

packages/components/psammead-navigation/README.md
tochwill: Do we need to export these from the index file to be able to import them like this?

tochwill: Ah Alistair already mentioned this, sorry!

OlgaLyubin: It is now in the Dropdown folder. Not sure whether it is a common practice to import and re-export components, have you seen this done before? 
Because for example in `StoryPromo`, `IndexAlsos` are in a separate subfolder but in the "Usage" section of README they are imported from the main index.

tochwill: It looks like there's a seperate package.json file for `IndexAlsos`, that means `@bbc/psammead-story-promo/index-alsos` resolves, which is where `IndexAlsos` are coming from - so that readme isn't quite right. I think we should export everything from the index personally?

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+  <DropdownWrapper>
+    <CrossButton
+      aria-label={announcedText}
+      onClick={onClose}
+      // eslint-disable-next-line react/jsx-boolean-value
+      aria-expanded="true"
+    >
+      {navigationIcons.cross}
+    </CrossButton>
+    <DropdownUl role="list">{children}</DropdownUl>
```
DenisHdz: Should this `DropdownUl` also be exported?

packages/components/psammead-navigation/README.md
```diff
@@ -6,7 +6,7 @@ This component is currently tagged as alpha and is not suitable for production u
 
 ## Description
 
-The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl` and `NavigationLi`. They use `ul` and `li` HTML elements respectively.
+The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl` and `NavigationLi`. They use `ul` and `li` HTML elements respectively. The package also includes a Hamburger Menu and a Dropdown Navigation which are visible for breakpoints under 600px.
```
ghost, line 9: You can also add a `Exports` section so that it becomes clear on how `@bbc/psammead-navigation` includes the `Hamburger Menu` and `Dropdown Navigation`
### Exports
- `/Dropdown` - Which is visible for breakpoints under 600px

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+const CrossButton = styled.button`
+  background-color: ${C_EBON};
+  width: ${MENU_ICON_SIDE_LENGTH};
+  height: ${MENU_ICON_SIDE_LENGTH};
+  padding: 0;
+  margin: 0;
+  border: 0;
+
+  &:focus {
+    box-shadow: inset 0 0 0 ${GEL_SPACING_HLF} ${C_WHITE};
```
DenisHdz: This doesn't look to work fine with high contrast in FF when hover and focus.

focus-border-hamburger
focus-border-cross

You could use a pseudo-element as we do in the Navigation.

packages/components/psammead-navigation/README.md
OlgaLyubin: Got it, thanks! Added the imports/exports

packages/components/psammead-navigation/src/Dropdown/index.jsx
ghost: I think it should. Having 
```jsx
<CanonicalDropdown>
    <DropdownNavigationLi>
```
makes you think that `CannonicalDropdown` is the wrapping `ul`. I would prefer having it exported so that it can be composed on the storybook.
```jsx
<CanonicalDropdown>
  <DropdownNavigationUl>
    <DropdownNavigationLi>
```
which improves clarity.

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+    </CrossButton>
+    <DropdownUl role="list">{children}</DropdownUl>
+  </DropdownWrapper>
+);
+
+AmpDropdown.propTypes = {
+  announcedText: string.isRequired,
+  children: node.isRequired,
+  onClose: string.isRequired,
+};
```
ghost: Since `CannonicalDropdown` and `AmpDropdown` have the same props, we can define a variable to store the propTypes. This will ensure that when a prop is added to `CannonicalDropdown` it also gets added to `AmpDropdown`.

packages/components/psammead-navigation/README.md
```diff
@@ -83,7 +196,9 @@ It includes a ["skip link"](https://www.w3.org/TR/WCAG20-TECHS/G1.html) giving u
 
 We have added the role `list` and `listitem` to the `NavigationUl` and `NavigationList` respectively, due to a VoiceOver bug to reinstate the list semantics.
 
-We have also added visually hidden text to let the user know which item in the Navigation is the current page. Note the use of visually hidden text here is due to lack of support at this time for the aria-current page attribute. Also note the use of `role="text"` to stop splitting in VoiceOver.
+We have also added visually hidden text to let the user know which item in both regular and dropdown Navigation is the current page. Note the use of visually hidden text here is due to lack of support at this time for the aria-current page attribute. Also note the use of `role="text"` to stop splitting in VoiceOver.
```
ghost:
```suggestion
We have also added visually hidden text to let the user know which item in both regular and dropdown Navigation is the current page. Note the use of visually hidden text here is due to lack of support at this time for the aria-current page attribute. Also note the use of `role="text"` to stop text splitting in VoiceOver.
```

packages/components/psammead-navigation/README.md
```diff
@@ -71,6 +118,72 @@ import { latin } from '@bbc/gel-foundations/scripts';
...
+import { AmpDropdown, DropdownNavigationLi } from '@bbc/psammead-navigation';
+import { latin } from '@bbc/gel-foundations/scripts';
+
+<AmpDropdown announcedText="Menu" onClose="tap:menu.toggleVisibility">
+    <DropdownNavigationLi
+        script={latin}
+        service="news"
+        key="Home"
+        url="/"
+        active={true}
```
ghost:
```suggestion
        active
```

packages/components/psammead-navigation/README.md
```diff
@@ -71,6 +118,72 @@ import { latin } from '@bbc/gel-foundations/scripts';
...
+import { CanonicalDropdown, DropdownNavigationLi } from '@bbc/psammead-navigation';
+import { latin } from '@bbc/gel-foundations/scripts';
+
+<CanonicalDropdown announcedText="Menu" onClose={() => {}}>
+    <DropdownNavigationLi
+        script={latin}
+        service="news"
+        key="Home"
+        url="/"
+        active={true}
```
ghost:
```suggestion
        active
```

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+  &:focus {
+    box-shadow: inset 0 0 0 ${GEL_SPACING_HLF} ${C_WHITE};
+  }
+`;
+
+const DropdownUl = styled.ul`
+  list-style-type: none;
+  margin: 0;
+  padding-right: ${GEL_SPACING};
+  padding-left: ${GEL_SPACING};
```
ghost:
```suggestion
padding: 0 ${GEL_SPACING};
```

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+  border-bottom: 0.125rem solid ${C_SHADOW};
+`;
+
+const DropdownLi = styled.li`
+  padding-top: 0.75rem;
+  padding-bottom: 0.75rem;
```
ghost:
```suggestions
padding: .75rem 0;
```

packages/components/psammead-navigation/src/Dropdown/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+          <VisuallyHiddenText>{currentPageText}</VisuallyHiddenText>
+          {children}
+        </StyledCurrentLink>
+      </StyledDropdownLink>
+    ) : (
+      <StyledDropdownLink script={script} service={service} href={url}>
+        {children}
+      </StyledDropdownLink>
+    )}
+  </DropdownLi>
```
ghost: You can place the checks inside ` <StyledDropdownLink>` since that is a common wrapper and have.
```jsx
 <DropdownLi role="listitem">
    <StyledDropdownLink script={script} service={service} href={url}>
      {active && currentPageText ? (
        <StyledCurrentLink
          dir={dir}
          role="text"
        >
          <VisuallyHiddenText>{currentPageText}</VisuallyHiddenText>
          {children}
        </StyledCurrentLink>
      ) : (
        children
      )}
    </StyledDropdownLink>
  </DropdownLi>
```

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: Props names are the same, however in `CanonicalDropdown` `onClose` is a function, but in an `AmpDropdown` `onClose` is a string

ghost: Aaah ok. What do you think about 
```js
const dropdownProps = /* the props */

AmpDropdown.propTypes = {
...dropdownProps
onClose: string
}
```
Am concerned that the `props` on `CannonicalDropdown` and `AmpDropdown` might become out of sync when adding `props` that are common to both.

packages/components/psammead-navigation/README.md
tochwill: Sorry about the back and forth on this - a lot of other packages in psammead export in the pattern suggested by @sadickisaac and @AlistairGempf above rather than exporting everything from the index, it might be better to follow this pattern and we can update the `StoryPromo` README to avoid confusion going forward.

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: Looks good, will add that now

packages/components/psammead-navigation/src/index.stories.jsx
```diff
@@ -148,6 +159,37 @@ const navigationStory = (
   </>
 );
 
+const hamburgerStory = () => {
+  return (
+    <BackgroundContainer>
+      <CanonicalHamburgerMenu announcedText="Menu" onOpen={() => {}} />
+    </BackgroundContainer>
+  );
+};
```
ghost: Are we planning to reuse and include it as part of another story?

packages/components/psammead-navigation/README.md
OlgaLyubin: @tochwill  Just to double check, all of the exports should still be in `./Dropdown` but in the README we should add an `Exports` section and also update imports in the `Usage` section?

packages/components/psammead-navigation/src/index.stories.jsx
OlgaLyubin: Do you mean including it as a part of the Navigation stories with brand?

ghost: Yes.

OlgaLyubin: IMO it does not bring too much value to have a `Brand`+`HamburgerMenu` story for now. After this PR or `SwipeableNav` PR is merged we could add stories for `Brand`+`HamburgerMenu`+`SwipeableNav`

ghost: I agree, but i was thinking more in terms of. This cannot be reused since it wraps `CannonicalHamburgerMenu` in a `BackgroundContainer`, therefore it would make more sense to put in inside `.add('Hamburger', () => {here})` instead of creating a function and then passing it in.

OlgaLyubin: Oh okay, understood. That is a good idea, will fix it

packages/components/psammead-navigation/README.md
tochwill: Yeah, that's correct. You can do the same as the index-alsos folder and put in a package.json file so it resolves as `navigation/dropdown`.

AlistairGempf: There's a similar thing in @DenisHdz's PR: https://github.com/bbc/psammead/pull/2577/files#diff-a29c4745f8cdf126f109082527219790R1

OlgaLyubin: Thanks all, this is now updated :)

packages/components/psammead-navigation/src/Dropdown/index.jsx
OlgaLyubin: Done now:)

packages/components/psammead-navigation/README.md
```diff
@@ -6,7 +6,11 @@ This component is currently tagged as alpha and is not suitable for production u
 
 ## Description
 
-The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl` and `NavigationLi`. They use `ul` and `li` HTML elements respectively.
+The `@bbc/psammead-navigation` package is a set of two components, `NavigationUl` and `NavigationLi`. They use `ul` and `li` HTML elements respectively. The package also includes a Hamburger Menu and a Dropdown Navigation which are visible for breakpoints under 600px.
+
+## Exports
+
+`/dropdown-navigation` - adds a dropdown navigation with hamburger menu which is visible for breakpoints under 600px
```
tochwill: Sorry, just a tiny change - could this just be `/dropdown`? Then this would match `/swipeable`.

packages/components/psammead-navigation/README.md
```diff
@@ -71,6 +122,76 @@ import { latin } from '@bbc/gel-foundations/scripts';
...
+    </DropdownNavigationLi>
+  </DropdownUl>
+</CanonicalDropdown>
+```
+
+## Amp Dropdown Navigation Usage
+
+```jsx
+import React from 'react';
+import { AmpDropdown, DropdownUl, DropdownNavigationLi } from '@bbc/psammead-navigation/dropdown-navigation';
```
tochwill: Could this just be `DropdownLi` so the naming convention is the same for both of these components?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+  <DropdownWrapper>
+    <CrossButton
+      aria-label={announcedText}
+      on={onClose}
+      // eslint-disable-next-line react/jsx-boolean-value
+      aria-expanded="true"
+    >
+      {navigationIcons.cross}
+    </CrossButton>
+    <DropdownUl role="list">{children}</DropdownUl>
```
tochwill: The canonical dropdown just renders children, amp is putting the children within the Ul component - should we keep this consistent?

packages/components/psammead-navigation/src/index.stories.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
```
tochwill, line 1: Could we add tests and stories for the AMP components?

packages/components/psammead-navigation/src/DropdownNavigation/index.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+  DropdownUl,
+  DropdownNavigationLi,
+  CanonicalHamburgerMenu,
+} from './index';
+import pidginNavData from '../../testHelpers/pidgin';
+
+describe('Canonical dropdown navigation', () => {
+  shouldMatchSnapshot(
+    'should render correctly',
+    <CanonicalDropdown announcedText="Menu" onClose={() => {}}>
```
tochwill: Should there also be a test to make sure that the onClose handler runs when the cross icon is clicked?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
OlgaLyubin: Sorry, forgot to update the AMP version too

packages/components/psammead-navigation/src/DropdownNavigation/index.test.jsx
AlistairGempf: I thought the same thing and looked into this. We can do something similar to https://github.com/bbc/psammead/blob/latest/packages/components/psammead-media-player/src/Placeholder/index.test.jsx#L58.

```jsx
  it('should click click', () => {
    const mockOnClick = jest.fn();
    const { container } = render(
      <CanonicalHamburgerMenu announcedText="Menu" onOpen={mockOnClick} />,
    );
    const button = getByRole(container, 'button');
    fireEvent.click(button);
    expect(mockOnClick).toBeCalledTimes(1);
  });
```

packages/components/psammead-navigation/src/index.stories.jsx
OlgaLyubin: I was not sure about adding AMP stories since they will look exactly the same as Canonical ones. Will this be useful?

With regards to tests, if I add snapshots for AMP, they will be identical to Canonical navigation. Since the only difference of `on` vs `onChange` is not affecting appearance and thus is not registered in snapshots. So again, was not sure if it's worth adding an AMP version?

packages/components/psammead-navigation/src/DropdownNavigation/index.test.jsx
OlgaLyubin: Good point, will add that

packages/components/psammead-navigation/src/index.stories.jsx
tochwill: I think we should definitely have a test for the AMP version - even if it looks the same, they're supposed to look the same, and this will guard against accidentally breaking something if changes are made to one but not to the other.

OlgaLyubin: Makes sense, I will add a test then. What do you think about AMP stories though?

tochwill: I also think an AMP story would be useful - you can create everything that needs to be passed into the component once and then pass it into the canonical wrapper and then into the AMP wrapper?

tochwill (approved):

ghost (approved):

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+`;
+
+const iconBorder = css`
+  content: '';
+  position: absolute;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  top: 0;
+  height: ${MENU_ICON_SIDE_LENGTH};
```
DenisHdz: You don't need this `height` and `width` if you add `position: relative` to the `HamburgerButton` and `CrossButton`

DenisHdz (approved):

packages/components/psammead-navigation/README.md
```diff
@@ -71,6 +122,76 @@ import { latin } from '@bbc/gel-foundations/scripts';
...
+```jsx
+import React from 'react';
+import { AmpHamburgerMenu } from '@bbc/psammead-navigation/dropdown';
+
+<AmpHamburgerMenu announcedText="Menu" onOpen="tap:menu.toggleVisibility" />
+```
+
 ### When to use this component
 
 The `Navigation` is designed to show a navigation bar on `index` pages, which will show all sections on a site. If there are too many items to fit on one line, the items will wrap to the next lines.
```
greenc05: Just index pages?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+  const direction = dir === 'ltr' ? 'left' : 'right';
+  return `border-${direction}: ${GEL_SPACING_HLF} solid ${C_WHITE};
+          padding-${direction}: ${GEL_SPACING};`;
+};
+
+const DropdownWrapper = styled.div`
+  background-color: ${C_EBON};
+
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    display: none;
```
greenc05, line 30: and visibility hidden please?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+  background-color: ${C_EBON};
+  width: ${MENU_ICON_SIDE_LENGTH};
+  height: ${MENU_ICON_SIDE_LENGTH};
+  position: relative;
+  padding: 0;
+  margin: 0;
+  border: 0;
+
+  &:focus {
+    box-shadow: inset 0 0 0 ${GEL_SPACING_HLF} ${C_WHITE};
```
greenc05: have we checked support of this in high contrast mode on PC and when the user changes colour preferences in FF?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+const CrossButton = styled.button`
+  background-color: ${C_EBON};
+  width: ${MENU_ICON_SIDE_LENGTH};
+  height: ${MENU_ICON_SIDE_LENGTH};
+  position: relative;
+  padding: 0;
+  margin: 0;
+  border: 0;
+
+  &:focus {
```
greenc05: are we missing hover styling?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+  }
+`;
+
+const StyledDropdownLink = styled.a`
+  ${({ script }) => script && getPica(script)};
+  ${({ service }) => service && getSansRegular(service)}
+  color: ${C_WHITE};
+  text-decoration: none;
+
+  &:focus {
```
greenc05, line 59: are we missing hover?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+const HamburgerButton = styled.button`
+  width: ${MENU_ICON_SIDE_LENGTH};
+  height: ${MENU_ICON_SIDE_LENGTH};
+  background-color: ${C_POSTBOX};
+  position: relative;
+  padding: 0;
+  margin: 0;
+  border: 0;
+
+  &:focus {
```
greenc05, line 127: hover?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+    box-shadow: inset 0 0 0 ${GEL_SPACING_HLF} ${C_WHITE};
+    ::after {
+      ${iconBorder};
+    }
+  }
+
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    display: none;
```
greenc05, line 135: visibility hidden

greenc05:
Just spoken to @OlgaLyubin about fixing up the buttons so it's just one therefore matches the screen reader ux, and also needs focusable false and aria hidden to the svg

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
tochwill: I'm not 100% sure about this convention - do we know of browsers or situations that don't support `display: none`? And if they don't would we not need to deal with the fact the element still affects the layout?

PriyaKR:
Looks good to me.Since its an alpha component accessibility testing not required at this stage.
The hamburger menu icon and the menu dropdown component looks good(the toggle logic not included in this issue will be handled by https://github.com/bbc/simorgh/issues/4571 as mentioned in the simorgh issue)

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
greenc05: I can't actually find any reference to this on the net, this is something we have used for years, compatibility may of improved since we started using it (sorry haven't had time to research), considering the number of AT combo's and versions that could potentially be used across the world I certainly don't think it harms though

tochwill:
Just noticed a couple of things going through the review, can tidy up and move back in :)

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+  url,
+  dir,
+}) => (
+  <StyledDropdownLi role="listitem">
+    <StyledDropdownLink script={script} service={service} href={url}>
+      {active && currentPageText ? (
+        <StyledCurrentLink
+          dir={dir}
+          // eslint-disable-next-line jsx-a11y/aria-role
+          role="text"
```
j-pendlebury: we could use styled component's `attrs` function to add the aria-role, like [here](https://github.com/bbc/psammead/blob/b33d9e8f3b56a7c0ae5b3eb5e13052bbdb5180cd/packages/components/psammead-copyright/src/index.jsx#L7-L9). This then means we won't have to have that ugly comment here. Just a suggestion though

j-pendlebury (approved):
just one suggestion from me, but otherwise I'm happy with this 👍

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
OlgaLyubin: Good point, will add that 👍

packages/components/psammead-navigation/src/index.stories.jsx
```diff
@@ -163,11 +176,36 @@ const navigationStory = (
...
+    </DropdownUl>
+  </Dropdown>
+);
+
 const storiesWithoutBrand = storiesOf(
   'Components|Navigation/without brand',
   module,
 )
   .addDecorator(withKnobs)
+  .addDecorator(ampDecorator)
```
Bopchy: Everything looks really good to me 👍 . Just wondering if adding an ampDecorator to all the storiesWithoutBrand was on purpose?

Bopchy (approved):
Other than the comment on the ampDecorator, everything looks good 👍

packages/components/psammead-navigation/src/index.stories.jsx
OlgaLyubin: Yes, this was on purpose in order to make the AMP menu button work. If you click on the AMP button now, the menu icon will change

DenisHdz (approved):
Good job on the AMP side 👍

PriyaKR:
@OlgaLyubin is the hamburger menu svg supposed to be in red colour?I can see that its in black colour.

tochwill:
> @OlgaLyubin is the hamburger menu svg supposed to be in red colour?I can see that its in black colour.

Hey! It's transparent so it's the same colour as the menu behind it

PriyaKR:
Looks good to me except one small font size with the dropdown menu ie) at 360-599px expected font size is 15px actual its 16px.If UX is fine with it then this PR can be merged.

PriyaKR:
UX is fine with the above small issue.

PR #2625, opened 2019-11-13
Create Embed Error component

### LINKED ISSUE #2624: Create a generic AV embed error component
As part of https://github.com/bbc/simorgh-infrastructure/issues/748 we need to create a component that replaces the full error page with one designed for AV embeds.

Create a component:
- That displays a generic AV embed error.
- That is designed to be displayed in an iframe.

n/a

**Testing notes**
[Tester to complete]

**Additional context**
An example of what this _might_ look like:
image

simonsinclair:
Resolves https://github.com/bbc/psammead/issues/2624.

**Overall change:** Adds `EmbedError` – a lean error message component, designed for embedded content, such as that in an iframe. It will fill its parent container and can optionally fill the viewport, while positioning its content accordingly. Checkout the [readme](https://github.com/bbc/psammead/blob/a7aeb7a32990f2ac4c765b9cc79f46b75bb1e7cb/packages/components/psammead-embed-error/README.md).

**Code changes:**

- Initialise a new Psammead component.

---


packages/components/psammead-embed-error/src/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+const StyledEmbedError = styled.div`
+  ${({ service }) => getSansRegular(service)}
+  ${GEL_BODY_COPY};
+  background-color: ${C_CHALK};
+  background-image: url(data:image/svg+xml;base64,${BBC_BLOCKS});
+  background-position: center center;
+  background-repeat: no-repeat;
+  background-size: 38.2%;
+  color: ${C_EBON};
+  height: 100vh;
```
ryanmccombe: Looks good 👍 

Would be nice if we could get this to not always assume it is going to be in an iframe though - the 100vh will look weird in any other scenario.  If we know the media will fail, we'd like to be able to circumvent the embed endpoint entirely.

This is probably more relevant to media pod pages than articles though, so happy to handle this later

ryanmccombe (approved):

packages/components/psammead-embed-error/src/index.jsx
simonsinclair: Thanks, Ryan, this is a good suggestion.

I've implemented a `fillViewport` prop, which, when active, will cause this component to fill the viewport. By default it will now fill its parent container.

HarveyPeachey (approved):
LGTM, potentially needs a dev a11y swarm

packages/components/psammead-embed-error/src/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+  background-repeat: no-repeat;
+  background-size: 38.2%;
+  color: ${C_EBON};
+  display: flex;
+  flex-direction: column;
+  height: ${({ fillViewport }) => (fillViewport ? '100vh' : '100%')};
+  justify-content: flex-end;
+`;
+
+const StyledErrorMessage = styled.p`
```
rhenshaw56: I think we could just use `<strong>` directly instead of wrapping it in a `<p>`, @simonsinclair what do you think ?

RayNjeri (approved):
Nice work 👍

packages/components/psammead-embed-error/src/index.jsx
simonsinclair: Technically we could, yes. Normally I'd mark up a paragraph of text in a `<p>`, but I guess we can't always be sure we're going to receive a paragraph here. Good spot.

packages/components/psammead-embed-error/src/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+const GEL_SPACING_QRT = '0.125rem';
+
+const StyledEmbedError = styled.div`
+  ${({ service }) => getSansRegular(service)}
+  ${GEL_BODY_COPY};
+  background-color: ${C_CHALK};
+  background-image: url(data:image/svg+xml;base64,${BBC_BLOCKS});
+  background-position: center center;
+  background-repeat: no-repeat;
+  background-size: 38.2%;
```
jamesdonoh: Might be worth commenting the magic number to avoid head-scratching.

packages/components/psammead-embed-error/src/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+  display: flex;
+  margin: ${GEL_SPACING_TRPL};
+
+  strong {
+    font-weight: normal;
+  }
+`;
+
+const IconWrapper = styled.span`
+  margin-right: ${GEL_SPACING};
```
jamesdonoh: This reminds me, how will this work for RTL sites?

jamesdonoh (commented):
Nice job, just wondering have you also tested if this will work on AMP pages?

packages/components/psammead-embed-error/src/index.jsx
simonsinclair: I've moved the `38.2%` to a named constant.

packages/components/psammead-embed-error/src/index.jsx
simonsinclair: Good spot. I've added support using symmetrical margins:

Screen Shot 2019-11-18 at 14 03 54
Screen Shot 2019-11-18 at 14 04 12

pharingee (approved):
LGTM  👍🏾

jamesdonoh (approved):

j-pendlebury (approved):

PR #2610, opened 2019-11-12
Update screen reader UX for guidance message

rhenshaw56:
Resolves #2580 and probably #2573 

**Overall change:** Passed `guidanceMessage` prop to play button 

**Code changes:**

- Passed `guidanceMessage` prop to play button 
- Adds `aria-hidden=true` to media player placeholder guidance

---

- [X] I have assigned myself to this PR and the corresponding issues
- [X] Automated jest tests added (for new features) or updated (for existing features)
- [X] This PR requires manual testing

Issue #2573, opened 2019-11-06
AV: A11y screen reader issues with Guidance in psammead-media-player


HarveyPeachey:
**Describe the bug**
In the dev a11y swarm done for #2517, a few issues were found as documented in this [comment](https://github.com/bbc/psammead/pull/2517#issuecomment-549826195):

When looking at screen readers with IE there were two bugs which were found:
- **JAWS** - Reads out guidance then states it's clickable.

**To Reproduce**
Steps to reproduce the behaviour:

**JAWS**
1. Open internet explorer on windows.
2. Go to [storybook](https://bbc.github.io/psammead/?path=/story/components-media-player--guidance) (When https://github.com/bbc/psammead/pull/2517 has been merged).
3. Open Jaws.
4. Follow step 4 https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-jaws.
5. See error stated.

**Expected behaviour**
**JAWS**
Shouldn't state guidance to be clickable.

**Testing notes**

- [x] This bug fix is expected to need manual testing.

**Additional context**

HarveyPeachey:
This bug is now non-existant, as the new implementation of the guidance message has fixed this.

PR #2627, opened 2019-11-14
Adds guidanceMessage prop to Play button

rhenshaw56:
Resolves #2580 and probably #2573 
part of https://github.com/bbc/psammead/pull/2610

Should go out before https://github.com/bbc/psammead/pull/2628

**Overall change:** Passed `guidanceMessage` prop to play button 

**Code changes:**

- Passed `guidanceMessage` prop to play button 

---

- [X] I have assigned myself to this PR and the corresponding issues
- [X] Automated jest tests added (for new features) or updated (for existing features)
- [X] This PR requires manual testing

PR #2628, opened 2019-11-14
Adds aria-hidden to media player placeholder guidance

### LINKED ISSUE #2580: Update screen reader UX for guidance message in media placeholder
We need to update the screen reader UX so that the guidance message is announced with the play button of the media player, as screen reader users can navigate via buttons and may miss this important message otherwise.

According to the screen reader UX, the guidance message needs to be announced with the play button and before the 'Play video, ...' text, and should follow this format:
```
Guidance, contains strong language and adult humour. Play video, “Video Title”, 1 minute 40 seconds
```
It maybe possible to link up the guidance text that is visually displayed with aria to the play button (instead of passing the guidance text as visually hidden text to the button), though that is only a viable solution if the aria attribute is supported by all supported AT **and** the guidance message is read out first before 'Play video, ...'.

**Testing notes**
[Tester to complete]

**Additional context**

rhenshaw56:
Resolves #2580 and probably #2573 
part of https://github.com/bbc/psammead/pull/2610

**Overall change:** Adds `aria-hidden=true` to media player placeholder guidance

**Code changes:**

- Adds `aria-hidden=true` to media player placeholder guidance

---


rhenshaw56:
This can  be reviewed as it is but should be merged after https://github.com/bbc/psammead/pull/2627

RayNjeri (approved):

packages/components/psammead-media-player/src/Guidance/index.jsx
```diff
@@ -39,7 +39,7 @@ const GuidanceMessage = styled.strong`
 
 const Guidance = ({ guidanceMessage, service }) => (
   <GuidanceWrapper service={service}>
-    <GuidanceMessage>{guidanceMessage}</GuidanceMessage>
+    <GuidanceMessage aria-hidden="true">{guidanceMessage}</GuidanceMessage>
```
ghost, line 42: This is a boolean prop but we are passing a `string`. It should either be `<GuidanceMessage aria-hidden>` or `<GuidanceMessage aria-hidden={true}>`
```suggestion
    <GuidanceMessage aria-hidden>{guidanceMessage}</GuidanceMessage>
```

simonsinclair (approved):

packages/components/psammead-media-player/src/Guidance/index.jsx
simonsinclair: Hi @sadickisaac, aria-hidden can be `true`, `false` or `undefined`.
- `<GuidanceMessage aria-hidden>` is the same as `<GuidanceMessage aria-hidden="true">`
- `<GuidanceMessage aria-hidden="false">` is not the same as `<GuidanceMessage>`

In this case, your suggestion would save us a few bytes, but considering `aria-hidden` can also `="false"` I like the extra verbosity. 😄

j-pendlebury (approved):
LGTM

jamesbrumpton:
lgtm

rhenshaw56:
will be merged after https://github.com/bbc/psammead/pull/2627

PR #2634, opened 2019-11-15
Bulletin component - tablet and desktop layout

### LINKED ISSUE #2622: Bulletin component - tablet and desktop layout
We already have designs for tablet and desktop layout for the Radio and TV bulletin.
At the moment we are hiding the Bulletin component above 600px, but this needs to be visible across all the breakpoints.

The designs can be found [here](https://app.zeplin.io/project/5c8a5e021d454124ef90669f/dashboard?tag=TV%20%2F%20Radio%20Bulletin).

Remove `600px` condition and create wrappers for the image and the text as we do in the `StoryPromo`, passing a prop with the type to determine what number of columns to use.

### 600px-1007px

**With image:**

Note that the image will span 2 columns and the text 4 columns, as the `StoryPromo`.
```
const ImageGridColumnsRadioBuletin = css`
  grid-column: 1 / span 2;
`;

const TextGridColumnsRadioBuletin = css`
  grid-column: 3 / span 4;
`;
```

radio-with-image

**Without image:**

It will look like the mobile version.
radio-without-image

### 1008+px
For now, we don't have to do anything because it will be displayed as a tablet.

But later on, once the Grid is integrated into Simorgh, it will have to use the same design as mobile:

radio-1008px

## TV bulletin

Note that the image and text will take the half width, as the `TopStoryPromo`.

```
const ImageGridColumnsTvBulletin = css`
    grid-column: 1 / span 3;
`;

const TextGridColumnsTvBulletin = css`
    grid-column: 4 / span 3;
`;
```
**600px-1007px:**

tv-600-1007px

**1008+px:**

tv-1008px

N/A

**Testing notes**
Regression test in psammead storybook

**Additional context**

ghost:
Resolves #2622

**Overall change:** _Style `Bulletin` component on desktop and tablet._

**Code changes:**

- _Separate `TV` and `Radio` style to make it more readable._
- _Update snapshots_

---


packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -34,38 +131,74 @@ const IconWrapper = styled.span`
...
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) and (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
+    display: inline-block;
+    padding: ${GEL_SPACING} ${GEL_SPACING_DBL} ${GEL_SPACING} 0;
+  }
+  @media (min-width: ${GEL_GROUP_4_SCREEN_WIDTH_MIN}) {
+    width: 100%;
+  }
 `;
 
 const BulletinSummary = styled.p`
```
AlistairGempf, line 152: Is there a reason this isn't separated into a `RadioSummary` and a `TVSummary` in the same way as other components are separated?

ghost: The difference between the `RadioSummary` and `TvSummary` was quite small. I didn't see it necessary to create separate styled tags.

OlgaLyubin:
Is there a reason why `RadioBulletin` story has an image for 600px to 1008px breakpoints and does not have it for others? There are designs on Zeplin for both Radio Bulletin with an image and without one. So it seemed like there should be separate stories for both options

OlgaLyubin:
TV Bulletin does not seem to match the design for 1008px+ : 

Screenshot 2019-11-19 at 11 23 15

Screenshot 2019-11-19 at 11 23 37

hotinglok:
> Is there a reason why `RadioBulletin` story has an image for 600px to 1008px breakpoints and does not have it for others? There are designs on Zeplin for both Radio Bulletin with an image and without one. So it seemed like there should be separate stories for both options

Adding to this, based on the 'Radio Bulletin Link promo v2 - no image 600px-1007px design, it seems like the image is optional. I think it would be worth separating the stories into folders and having separate stories for that perhaps.  
~I also think the 1008px design doesn't match the design on the zeplin link, it seems as if it should look a lot more like the 400px-599px but in a grid with 'LIVE' and 'Not Live' variants perhaps.~
Just realised that it just looks off because it isn't in a grid. Would like confirmation that this is meant to fill a whole grid cell though and if possible represent it in the story a little better. I still think it would be nice just to show 'LIVE' and 'Not Live' variants like in the design.

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -79,25 +191,42 @@ const Bulletin = ({
...
+  const isAudio = type === 'audio';
+  const BulletinWrapper = isAudio ? RadioBulletinWrapper : TVBulletinWrapper;
+  const ImageWrapper = isAudio ? RadioImageWrapper : TVImageWrapper;
+  const TextWrapper = isAudio ? RadioTextWrapper : TVTextWrapper;
+  const BulletinHeading = isAudio ? RadioHeading : TVHeading;
+
+  return (
+    <BulletinWrapper>
+      <ImageWrapper>{image}</ImageWrapper>
+      <TextWrapper direction={dir}>
```
sadickisaac: Using `dir` as a prop in `TextWrapper` causes the `Bulletin` to shift to the right in `rtl`

Screenshot 2019-11-21 at 10 00 36

ghost:
After discussions with UX. The following were agreed on.
- Lunar Background to the TV Bulletin - to match the Radio bulletin
- We will only have two designs for all breakpoints -  600px+ and Below 600px (Refer to the attachments)

Also, we have agreed that we will work with Editorial to ensure they always add an image to the Radio bulletin and if they don't, we accept the risk the current radio bulletin build will be shown (So no fall back implementation required)

ghost:
@lokh01 
> I still think it would be nice just to show 'LIVE' and 'Not Live' variants like in the design.

There is a `Live` knob on the story, you can use that to toggle

Bopchy:
For RTL margin-left needs to be added between the headline and `LIVE` for both radio and TV 
Screenshot 2019-11-21 at 12 55 20

ghost:
@Bopchy fixed the `LiveLabel` margins.

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -16,14 +20,99 @@ import { getPica, getLongPrimer } from '@bbc/gel-foundations/typography';
...
+  width: ${fullWidthColumnsMaxScaleable};
+  @supports (${grid}) {
+    width: initial;
+    padding: 0;
+  }
+`;
+
+const TVBulletinWrapper = styled.div`
+  ${bulletinWrapperStyles};
+  padding: ${GEL_SPACING};
```
DenisHdz: Should this padding be applied in Simorgh? I remember we mentioned that components shouldn't have any outer spacings in Psammead.

tochwill:
I've just noticed by testing in storybook that the whole page becomes a link - I think this is because the `before` pseudoelement in the link is set to `position: absolute` with top, bottom, left and right set to 0, and the wrapper is not set to `position: relative`, so these values are applied for the next relative parent, which in this case is the whole page.

packages/components/psammead-bulletin/src/index.jsx
ghost: If we remove the padding from the component, it would mean we have to set a `C_LUNAR` background in simorgh instead of psammead.

DenisHdz: I know, just wondering what were your thoughts on this.

ghost: `background` and the `padding` between the image and the wrapper are part of the component. It makes sense to have it in psammead.

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -16,14 +20,100 @@ import { getPica, getLongPrimer } from '@bbc/gel-foundations/typography';
...
+    padding: ${GEL_SPACING_DBL};
+  }
+`;
+
+const TVImageWrapper = styled.div`
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    grid-column: 1 / span 3;
+    width: ${halfWidthColumnsMaxScaleable};
+  }
+  ${imageWrapperStyles};
```
AlistairGempf, line 76: I feel like this should go at the start of the CSS, as this should hold the "defaults" which can then be overridden by the instance. What do you think?

DenisHdz (changes requested):
radio-bug

In situations like these, where the text is higher than the image, there should be a space between the button and the bottom. 

I have confirmed with UX that it should be 16px. You could apply it to the `RadioTextWrapper`

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -33,39 +123,63 @@ const IconWrapper = styled.span`
...
   ${({ service }) => getSansBold(service)}
   color: ${C_EBON};
   margin: 0; /* Reset */
-  ${({ type }) =>
-    type.toLowerCase() === 'audio'
-      ? `padding: ${GEL_SPACING} ${GEL_SPACING};`
-      : `padding: ${GEL_SPACING} 0;`}
+  padding: ${GEL_SPACING} 0;
+  @media(min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    padding:0 0 ${GEL_SPACING} 0;
```
Bopchy:
```suggestion
    padding: 0 0 ${GEL_SPACING} 0;
```

tochwill:
> radio-bug
> 
> In situations like these, where the text is higher than the image, there should a space between the button and the bottom.
> 
> I have confirmed with UX that it should be 16px. You could apply it to the `RadioTextWrapper`

So we just need 16px margin at all times then? Should the image remain aligned at the top with space below it in this case?

packages/components/psammead-bulletin/src/index.jsx
ghost: @AlistairGempf It should but having it on top means that a width of `50%` will be applied to the image. That is `50%` of the 3 columns it is supposed to span.

Screenshot 2019-11-21 at 21 03 15

packages/components/psammead-bulletin/src/index.jsx
ghost: @Bopchy Fixed.

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -16,14 +20,100 @@ import { getPica, getLongPrimer } from '@bbc/gel-foundations/typography';
...
 
-const BulletinWrapper = styled.div`
+const TVTextWrapper = styled.div`
   @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
-    display: none;
+    grid-column: 4 / span 3;
+    width: ${halfWidthColumnsMaxScaleable};
+    ${({ dir }) =>
+      dir === 'ltr'
+        ? `padding-right: ${GEL_SPACING_DBL};`
```
Bopchy:
```suggestion
        ? `padding-left: ${GEL_SPACING_DBL};`
```
As this is right now it's causing the padding on IE to be on the wrong side
Screenshot 2019-11-22 at 10 50 51

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -16,14 +20,100 @@ import { getPica, getLongPrimer } from '@bbc/gel-foundations/typography';
...
+        : `padding-left: ${GEL_SPACING_DBL};`}
```
Bopchy:
```suggestion
        : `padding-right: ${GEL_SPACING_DBL};`}
```

Bopchy (approved):
Looks good 👍

DenisHdz (approved):
Great work on this 👍

tochwill (approved):

PriyaKR:
Testing:
Noticed few UX related issues
- Summary font size and line height not matching designs across all widths for both radio and TV bulletin
- Line height of the heading not matching the design across all widths
- Radio click to play button has 'Listen Live` according to designs but actual its just `Listen`
- Also the radio component at above 600px without image seems to behave it has image and there is an empty space for the image
radio_bulletin

laurapsn:
@sadickisaac Following UX review, please amend: 

- CTA block to span full width on both TV and radio bulletin components
- change margin between headline and image from 8px to 16px on both TV and radio bulletin components 
- Headline font to change from Sans bold to Serif medium
- Padding between icon and label to be 8px

sadickisaac:
After reviewing the component with UX, here were some of the issues raised.
- The heading should be sans serif medium instead of sans bold
- The CTA should span the whole width for both TV and Radio `Bulletin`
- Space between the heading and Image should be `16px` top and `8px` bottom. This is the padding found on `h3`
- There should be a `8px` space between the icon and the words in the CTA.

ghost:
@PriyaKR I have done a review with UX
- The typography setting for the headline and summary are correct across breakpoints.
- I have updated the storybook to have the Live text.
- The expected behaviour of `Radio` component at above 600px without image is to have it empty without any fallbacks. This has been discussed with Editorial to make sure they add in an image.

PriyaKR:
The summary text font size/line height doesn't seem to match the designs little variations.As @laurapsn has reviewed it and happy it looks good to me.

PR #2668, opened 2019-11-22
Add service variant compatiablity

### LINKED ISSUE #2667: Make withServiceKnobs() variant compatible
Our current implementation for `withServiceKnobs()` makes it hard for us to distinguish services that are variants. They currently only return the service name as: ukChinaSimp, ukChinaTrad, serbianCyr, serbianLat and we are missing zhongwen :/ 
This currently doesn't work well if our simorgh implementation of `<ServiceProvider>`, as it expects `ukChinaTrad` (for example), is not a compatible service name (loadable names: https://github.com/bbc/simorgh/blob/latest/src/app/lib/config/services/loadableConfig.js). 

We need to  update the [text-variant file](https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-storybook-helpers/src/text-variants.js) to have a structure like this for variants: 

```
serbian: {
    lat: {...},
    cyr: {...}
}
```

This would be a breaking change, to the storybook-helper, as we will have to update a lot of logic around simorgh to handle this new structure.

Open to any data structure for handling our data structure, the example above is just what is currently being used on Simorgh. e.g: `src/app/lib/config/services/serbian.js` but any other data structure that is easy to handle is welcome.

**Testing notes**
[Tester to complete]

**Additional context**
More issues will be created from this, once this has been done and the data structure has been agreed on.

thekp:
Resolves #2667

**Overall change:** _Updated our example data for the storybook-helper package by adding variant prop into the storyProps object._

**Code changes:**

- _updated the text-variants file to include a variant and service value for variant service._
- _added logic to the withServiceKnob to pass a variant when it exists_
- _made the code more readable by adding constants at the top of the file_
- _install ramda to root package.json and use path to safely check for values in object_

---

### DIFF
--- a/packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
+++ b/packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
@@ -2,22 +2,25 @@ import React from 'react';
 import { Helmet } from 'react-helmet';
 import { select } from '@storybook/addon-knobs';
 import * as scripts from '@bbc/gel-foundations/scripts';
-import EXAMPLES from './text-variants';
+import EXAMPLES_TEXT from './text-variants';
 
 const DEFAULT_SERVICE = 'news';
-const SERVICES_LIST = Object.keys(EXAMPLES);
+const SERVICES_LIST = Object.keys(EXAMPLES_TEXT);
 
 export default ({
   defaultService = DEFAULT_SERVICE,
   services = SERVICES_LIST,
 } = {}) => storyFn => {
-  const getServiceAndVariant = selectedItem =>
-    EXAMPLES[selectedItem].variant
-      ? {
-          service: EXAMPLES[selectedItem].service,
-          variant: EXAMPLES[selectedItem].variant,
-        }
-      : { service: selectedItem };
+  const hasVariant = selectedItem =>
+    typeof EXAMPLES_TEXT[selectedItem].variant === 'string';
+
+  const getService = selectedItem =>
+    hasVariant(selectedItem)
+      ? EXAMPLES_TEXT[selectedItem].service
+      : selectedItem;
+
+  const getVariant = selectedItem =>
+    hasVariant(selectedItem) ? EXAMPLES_TEXT[selectedItem].variant : null;
 
   const includesSelectedService = selectedItem =>
     services.includes(selectedItem);
@@ -28,14 +31,15 @@ export default ({
     defaultService,
   );
 
-  const { text, script, locale, dir = 'ltr' } = EXAMPLES[selectedItem];
+  const { text, script, locale, dir = 'ltr' } = EXAMPLES_TEXT[selectedItem];
 
   const storyProps = {
     text,
     script: scripts[script],
     locale,
     dir,
-    ...getServiceAndVariant(selectedItem),
+    service: getService(selectedItem),
+    variant: getVariant(selectedItem),
   };
   return (
     <>
--- a/packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
+++ b/packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
@@ -2,20 +2,41 @@ import React from 'react';
 import { Helmet } from 'react-helmet';
 import { select } from '@storybook/addon-knobs';
 import * as scripts from '@bbc/gel-foundations/scripts';
-import LANGUAGE_VARIANTS from './text-variants';
+import EXAMPLES from './text-variants';
+
+const DEFAULT_SERVICE = 'news';
+const SERVICES_LIST = Object.keys(EXAMPLES);
 
 export default ({
-  defaultService = 'news',
-  services = Object.keys(LANGUAGE_VARIANTS),
+  defaultService = DEFAULT_SERVICE,
+  services = SERVICES_LIST,
 } = {}) => storyFn => {
-  const includesService = service => services.includes(service);
-  const service = select(
+  const getServiceAndVariant = selectedItem =>
+    EXAMPLES[selectedItem].variant
+      ? {
+          service: EXAMPLES[selectedItem].service,
+          variant: EXAMPLES[selectedItem].variant,
+        }
+      : { service: selectedItem };
+
+  const includesSelectedService = selectedItem =>
+    services.includes(selectedItem);
+
+  const selectedItem = select(
     'Select a service',
-    services.filter(includesService),
+    services.filter(includesSelectedService),
     defaultService,
   );
-  const { text, script, locale, dir = 'ltr' } = LANGUAGE_VARIANTS[service];
-  const storyProps = { text, script: scripts[script], locale, dir, service };
+
+  const { text, script, locale, dir = 'ltr' } = EXAMPLES[selectedItem];
+
+  const storyProps = {
+    text,
+    script: scripts[script],
+    locale,
+    dir,
+    ...getServiceAndVariant(selectedItem),
+  };
   return (
     <>
       <Helmet htmlAttributes={{ dir }} />
--- a/packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
+++ b/packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
@@ -7,27 +7,27 @@ import TEXT_EXAMPLES from './text-variants';
 
 const DEFAULT_SERVICE = 'news';
 const SERVICES_LIST = Object.keys(TEXT_EXAMPLES);
-const getVariant = selectedItem => path([selectedItem, 'variant']);
-const getService = selectedItem => path([selectedItem, 'service']);
+const getVariant = selectedService => path([selectedService, 'variant']);
+const getService = selectedService => path([selectedService, 'service']);
 const includesService = services => service => services.includes(service);
 
 export default ({
   defaultService = DEFAULT_SERVICE,
   services = SERVICES_LIST,
 } = {}) => storyFn => {
-  const selectedItem = select(
+  const selectedService = select(
     'Select a service',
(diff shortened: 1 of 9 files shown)


sareh (approved):
👍 Looks good. It adds variant support for Serbian/UKChina/Zhongwen. 
Though I also worked on part of it with @thekp, so it'd be good to get other eyes on it too.

FK78 (commented):
```
Branch must update CHANGELOG.md in psammead-storybook-helpers
Branch must update package-lock.json in psammead-storybook-helpers
Branch must update package.json in psammead-storybook-helpers
```

FK78 (approved):

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
```diff
@@ -2,20 +2,41 @@ import React from 'react';
...
 
 export default ({
-  defaultService = 'news',
-  services = Object.keys(LANGUAGE_VARIANTS),
+  defaultService = DEFAULT_SERVICE,
+  services = SERVICES_LIST,
 } = {}) => storyFn => {
-  const includesService = service => services.includes(service);
-  const service = select(
+  const getServiceAndVariant = selectedItem =>
```
jroebu14: Slightly misleading function name as it may not return a variant. The name suggests this function should be split up anyway. Maybe a function called `getService` and then one called `getVariant`.

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
```diff
@@ -2,20 +2,41 @@ import React from 'react';
 import { Helmet } from 'react-helmet';
 import { select } from '@storybook/addon-knobs';
 import * as scripts from '@bbc/gel-foundations/scripts';
-import LANGUAGE_VARIANTS from './text-variants';
+import EXAMPLES from './text-variants';
```
jroebu14: Reading this I'm constantly reminding myself what `EXAMPLES` is. It needs a better name `TEXT_EXAMPLES` or `TEXT_SNIPPETS` or something

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
```diff
@@ -2,20 +2,41 @@ import React from 'react';
...
-  const storyProps = { text, script: scripts[script], locale, dir, service };
+
+  const { text, script, locale, dir = 'ltr' } = EXAMPLES[selectedItem];
+
+  const storyProps = {
+    text,
+    script: scripts[script],
+    locale,
+    dir,
+    ...getServiceAndVariant(selectedItem),
```
jroebu14: what about something like

```js
  service: getService(),
  ...(hasVariant && { variant: getVariant() },
}
```

thekp: 👌 yeah that's easier to read.

thekp: ah slight issue with this is that, depending on if the service has a variant or not. The data structure is slightly different. Our non variant service, we get the service value from the key value but in the instance of a variant we would get it from the value of service within the object.

i.e 
```
news: {
  text: 'this is text'
},
serbianCyr: {
  service: 'serbian',
  variant: 'cyr',
  text: 'this is text'
}
```

thekp: Would this be a good implementation for it:
variant will just be passed into storyProps as null for non variant services
```
  const hasVariant = selectedItem =>
    typeof EXAMPLES_TEXT[selectedItem].variant === 'string';

  const getService = selectedItem =>
    hasVariant(selectedItem)
      ? EXAMPLES_TEXT[selectedItem].service
      : selectedItem;

  const getVariant = selectedItem =>
    hasVariant(selectedItem) ? EXAMPLES_TEXT[selectedItem].variant : null;

  const storyProps = {
    text,
    script: scripts[script],
    locale,
    dir,
    service: getService(selectedItem),
    variant: getVariant(selectedItem),
  };

```

jroebu14: Looks good but a few things I see...

- No need for typeof stuff in hasVariant. Do `Boolean(EXAMPLES_TEXT[selectedItem].variant)` instead
- I'd make `hasVariant` a boolean argument in `getService`
- I'd make `getVariant` simpler by having the conditional logic outside of the function and just returning `EXAMPLES_TEXT[selectedItem].variant` like in my initial comment

also why did you go for `EXAMPLES_TEXT` and not `TEXT_EXAMPLES`?

thekp: > also why did you go for EXAMPLES_TEXT and not TEXT_EXAMPLES?

i miss read the initial comment regarding that, will change to `TEXT_EXAMPLES` now 👍 

and thanks for the other feedback 💯

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
```diff
@@ -1,21 +1,46 @@
...
 export default ({
-  defaultService = 'news',
-  services = Object.keys(LANGUAGE_VARIANTS),
+  defaultService = DEFAULT_SERVICE,
+  services = SERVICES_LIST,
 } = {}) => storyFn => {
-  const includesService = service => services.includes(service);
-  const service = select(
+  const hasVariant = selectedItem =>
+    Boolean(EXAMPLES_TEXT[selectedItem].variant);
```
jroebu14: I'd use path here too

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
```diff
@@ -1,21 +1,45 @@
...
 
 export default ({
-  defaultService = 'news',
-  services = Object.keys(LANGUAGE_VARIANTS),
+  defaultService = DEFAULT_SERVICE,
+  services = SERVICES_LIST,
 } = {}) => storyFn => {
-  const includesService = service => services.includes(service);
-  const service = select(
+  const hasVariant = selectedItem => path([selectedItem, 'variant']);
```
sareh: Given that `getVariant` and `hasVariant` are identical, you can get rid of this function.

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
```diff
@@ -1,21 +1,45 @@
...
-    services.filter(includesService),
+    services.filter(includesService(services)),
     defaultService,
   );
-  const { text, script, locale, dir = 'ltr' } = LANGUAGE_VARIANTS[service];
-  const storyProps = { text, script: scripts[script], locale, dir, service };
+
+  const variant = getVariant(selectedItem)(TEXT_EXAMPLES);
+
+  const service = hasVariant(selectedItem)(TEXT_EXAMPLES)
```
sareh: Then here, you can use:
```suggestion
  const service = variant
```

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
thekp: 🤦‍♂ 
resolved now

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.test.jsx
```diff
@@ -2,7 +2,7 @@ import { render, waitForDomChange } from '@testing-library/react';
 import * as knobs from '@storybook/addon-knobs';
 import * as scripts from '@bbc/gel-foundations/scripts';
 import withServicesKnob from './withServicesKnob';
-import LANGUAGE_VARIANTS from './text-variants';
+import EXAMPLES from './text-variants';
```
EinsteinNjoroge:
```suggestion
import TEXT_EXAMPLES from './text-variants';
```

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.test.jsx
```diff
@@ -12,7 +12,7 @@ it('should correctly configure the default story book dropdown', () => {
 
   expect(knobs.select).toHaveBeenCalledWith(
     'Select a service',
-    Object.keys(LANGUAGE_VARIANTS),
+    Object.keys(EXAMPLES),
```
EinsteinNjoroge:
```suggestion
    Object.keys(TEXT_EXAMPLES),
```

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
```diff
@@ -1,21 +1,43 @@
...
 
 export default ({
-  defaultService = 'news',
-  services = Object.keys(LANGUAGE_VARIANTS),
+  defaultService = DEFAULT_SERVICE,
+  services = SERVICES_LIST,
 } = {}) => storyFn => {
-  const includesService = service => services.includes(service);
-  const service = select(
+  const selectedItem = select(
```
EinsteinNjoroge: 🤔We should give this variable a more descriptive name, `selectedService` perhaps.

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.test.jsx
thekp: goood spot

packages/utilities/psammead-storybook-helpers/src/withServicesKnob.jsx
thekp: yeah selectedService is more clear 👍

jroebu14 (approved):

PR #2710, opened 2019-11-27
GEL Grid - Margins update

sareh:
Part of https://github.com/bbc/simorgh/pull/4493 (Using Grid)

**Overall change:** Removes `enableGelMargins` & `enableNegativeGelMargins` props & adds `margins` prop (object)

This change from a single toggle for GEL Margins to an object was made to enable flexibility - for example, for Simorgh Article pages, we only need margins on groups 0-3. Groups 4 & 5 have fixed widths and so should not have margins applied to them, when used. 
So this usecase can be captured as so: 
```
    margins={{
      group0: true,
      group1: true,
      group2: true,
      group3: true,
      group4: false,
      group5: false,
    }}
```
Or more simply as this, since if you don't pass in a `true` value, the default for that group is `false`. 
```
    margins={{
      group0: true,
      group1: true,
      group2: true,
      group3: true,
    }}
```

**Code changes:**
- Remove  `enableGelMargins` & `enableNegativeGelMargins` 
- Add `margins`
- Update stories
- Update tests
- Update Documentation
- Update calculation for Gutter %s for fallbacks. Removes the negative margins for the gutters implementation, since this causes the page to overflow, since negative margins don't collapse horizontally with positive margins of children elements. 

Note: Fallback bugs relating to the Gutter percentages / width issues in IE11 are captured in this issue: https://github.com/bbc/psammead/issues/2722

---


packages/components/psammead-grid/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-grid",
-  "version": "1.0.0-alpha.15",
+  "version": "1.0.0-alpha.16",
```
EinsteinNjoroge: I'm not clear on how version bumps should be done for alpha components, IMO this should be a `major` change

packages/components/psammead-grid/src/index.jsx
```diff
@@ -335,8 +319,14 @@ Grid.propTypes = {
     group5: number.isRequired,
   }).isRequired,
   enableGelGutters: bool,
-  enableGelMargins: bool,
   enableNegativeGelMargins: bool,
+  margins: shape({
```
EinsteinNjoroge, line 323: We might want to simplify https://github.com/bbc/psammead/blob/margins-update/packages/components/psammead-grid/src/index.jsx#L323-L329 for scenarios where margins are enabled for all breakpoints. Perhaps passing in an optional `enableAllMargins: true` which should set margins to true for all breakpoints

sareh: I agree. Can we do this in a follow-up PR? For all of our Articles/FrontPages/current Simorgh pages, we will need to have margins on for groups 0-3 and off for 4-5, so we could simplify in that way, too.

packages/components/psammead-grid/package-lock.json
sareh: True, since it's pre-version 1.0.0, I'll make it a `1.1.0-alpha.0`.

EinsteinNjoroge (approved):
looks good 👍

jroebu14 (approved):
Looks good

PR #2759, opened 2019-12-03
Change menu button height depending on script in use

AlistairGempf:
Resolves #NUMBER

**Overall change:** Change menu button size depending on script in use - when changing the service in storybook, you'll see the menu icon seem to move place as the button changes size.

**Code changes:**

- Use the script object to change the height and width of the menu button based on the line-height
- Vertically align the svg within the button to ensure it is centred

---

### DIFF
--- a/packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
+++ b/packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
@@ -123,6 +123,10 @@ const iconBorder = css`
 const calculateButtonSide = lineHeight =>
   lineHeight / 16 + NAV_BAR_TOP_BOTTOM_SPACING * 2;
 
+const getButtonDimensions = lineHeight =>
+  `height: ${calculateButtonSide(lineHeight)}rem;
+  width: ${calculateButtonSide(lineHeight)}rem;`;
+
 const MenuButton = styled.button`
   position: relative;
   padding: 0;
@@ -130,16 +134,9 @@ const MenuButton = styled.button`
   border: 0;
   background-color: transparent;
   ${({ dir }) => (dir === 'ltr' ? `float: left;` : `float: right;`)}
-  ${({ script }) => {
-    const sideLength = `${calculateButtonSide(
-      script.pica.groupA.lineHeight,
-    )}rem;`;
-    return (
-      script &&
-      `height: ${sideLength};
-      width: ${sideLength};`
-    );
-  }}
+  ${({ script }) =>
+    script &&
+    getButtonDimensions(script.pica.groupA.lineHeight)}
 
   &:hover,
   &:focus {
@@ -154,16 +151,8 @@ const MenuButton = styled.button`
     visibility: hidden;
   }
   @media (min-width: ${GEL_GROUP_B_MIN_WIDTH}rem) {
-    ${({ script }) => {
-      const sideLength = `${calculateButtonSide(
-        script.pica.groupB.lineHeight,
-      )}rem;`;
-      return (
-        script &&
-        `height: ${sideLength};
-        width: ${sideLength};`
-      );
-    }}
+    ${({ script }) =>
+      script && getButtonDimensions(script.pica.groupB.lineHeight)}
   }
 
   & svg {
--- a/packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
+++ b/packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
@@ -19,7 +19,7 @@ import { getPica } from '@bbc/gel-foundations/typography';
 import { scriptPropType } from '@bbc/gel-foundations/prop-types';
 import { getSansRegular } from '@bbc/psammead-styles/font-styles';
 
-const TOP_BOTTOM_SPACING = 0.75; // 12px
+const MENU_TOP_BOTTOM_SPACING = 0.75; // 12px - must be the same as TOP_BOTTOM_SPACING in ../index.jsx
 
 const getStyles = dir => {
   const direction = dir === 'ltr' ? 'left' : 'right';
@@ -122,7 +122,7 @@ const iconBorder = css`
 // The sideLength of the button should be
 //  line height + top padding + bottom padding
 const calculateButtonSide = lineHeight =>
-  lineHeight / 16 + TOP_BOTTOM_SPACING * 2;
+  lineHeight / 16 + MENU_TOP_BOTTOM_SPACING * 2;
 
 const MenuButton = styled.button`
   position: relative;
--- a/packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
+++ b/packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
@@ -13,7 +13,6 @@ import Helmet from 'react-helmet';
 import {
   GEL_GROUP_3_SCREEN_WIDTH_MIN,
   GEL_GROUP_B_MIN_WIDTH,
-  GEL_GROUP_CD_MIN_WIDTH,
 } from '@bbc/gel-foundations/breakpoints';
 import { getPica } from '@bbc/gel-foundations/typography';
 import { scriptPropType } from '@bbc/gel-foundations/prop-types';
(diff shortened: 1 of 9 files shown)


packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -10,12 +10,16 @@ import {
...
+  GEL_GROUP_3_SCREEN_WIDTH_MIN,
+  GEL_GROUP_B_MIN_WIDTH,
+  GEL_GROUP_CD_MIN_WIDTH,
+} from '@bbc/gel-foundations/breakpoints';
 import { getPica } from '@bbc/gel-foundations/typography';
 import { scriptPropType } from '@bbc/gel-foundations/prop-types';
 import { getSansRegular } from '@bbc/psammead-styles/font-styles';
 
-const MENU_ICON_SIDE_LENGTH = '2.75rem'; // 44px
+const TOP_BOTTOM_SPACING = 0.75; // 12px
```
j-pendlebury:
```suggestion
const TOP_BOTTOM_SPACING = 0.75rem; // 12px
```

packages/components/psammead-navigation/src/DropdownNavigation/index.test.jsx
```diff
@@ -73,6 +75,7 @@ describe('Canonical', () => {
         onOpen={() => {}}
         isOpen
         onClose={() => {}}
+        script={latin}
```
j-pendlebury, line 78: can we have at least one test where script isn't `latin`?

j-pendlebury (commented):
just two things from me, otherwise looks good :)

OlgaLyubin (commented):
Great work 👍 Could you please also add `script` as a prop to Readme for sections "Canonical Menu Button Usage" and "Amp Menu Button Usage"?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
OlgaLyubin: Maybe we should give this a name that indicates that this variable is used for a menu button?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -137,6 +154,34 @@ const MenuButton = styled.button`
     display: none;
     visibility: hidden;
   }
+  @media (min-width: ${GEL_GROUP_B_MIN_WIDTH}rem) {
```
OlgaLyubin, line 153: Just wanted to clarify, why does this media query start from 320px?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -137,6 +154,34 @@ const MenuButton = styled.button`
...
+        script.pica.groupB.lineHeight,
+      )}rem;`;
+      return (
+        script &&
+        `height: ${sideLength};
+        width: ${sideLength};`
+      );
+    }}
+  }
+  @media (min-width: ${GEL_GROUP_CD_MIN_WIDTH}rem) {
```
Bopchy: This should be `max-width`. As the menuButton will be displayed below 600px

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
AlistairGempf: This needs to be a number so that it can be used in the calculation for the height [here](https://github.com/bbc/psammead/pull/2759/files#diff-a23047ff5909e5e673c346beeb894a3eR125).

We could do
```suggestion
const TOP_BOTTOM_SPACING = 0.75; //rem = 12px
```
to make the units clear? Let me know what you think.

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
AlistairGempf: This uses the typography groups in order to match when the line-height changes in the typography script object. These are based on the `GEL_GROUP_[A|B|C|D]` as I understand it, so I have started the media queries on those.

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
AlistairGempf: This makes sure the height of the menu button is the correct height if it ever is visible in `GEL_GROUP_CD`. If we think it's not worth having that "just in case" (and it probably isn't as long as if this button is ever visible at this breakpoint we put this back in) then this whole media query can just be removed.

Let me know what you think whether it's worth just removing this.

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
j-pendlebury: ahh okay, my bad. then I'm happy for this to just stay as it was :)

packages/components/psammead-navigation/src/DropdownNavigation/index.test.jsx
AlistairGempf: I've added Arabic rtl snapshot tests for each bit of the dropdown.

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
AlistairGempf: I've named it `MENU_BUTTON_TOP_BOTTOM_SPACING` and I've added to the comment saying it must match the psammead-navigation/src/index.jsx's `TOP_BOTTOM_SPACING`.

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -10,12 +10,16 @@ import {
...
+  GEL_GROUP_3_SCREEN_WIDTH_MIN,
+  GEL_GROUP_B_MIN_WIDTH,
+  GEL_GROUP_CD_MIN_WIDTH,
+} from '@bbc/gel-foundations/breakpoints';
 import { getPica } from '@bbc/gel-foundations/typography';
 import { scriptPropType } from '@bbc/gel-foundations/prop-types';
 import { getSansRegular } from '@bbc/psammead-styles/font-styles';
 
-const MENU_ICON_SIDE_LENGTH = '2.75rem'; // 44px
+const MENU_BUTTON_TOP_BOTTOM_SPACING = 0.75; // 12px - must be the same as TOP_BOTTOM_SPACING in ../index.jsx
```
tochwill: hmm, could we export that value and use it in here?

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
Bopchy: Ah, ok. I'm for removing it for now as it feels a bit premature to add it before it is actually required. Curious what others think though

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
AlistairGempf: Yep, let's do that.

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
OlgaLyubin: Probably best to remove it for now as it is not being used anyway

OlgaLyubin (approved):
Thanks for the changes, looks good to me 👍

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -115,15 +118,28 @@ const iconBorder = css`
...
 const MenuButton = styled.button`
-  width: ${MENU_ICON_SIDE_LENGTH};
-  height: ${MENU_ICON_SIDE_LENGTH};
   position: relative;
   padding: 0;
   margin: 0;
   border: 0;
   background-color: transparent;
   ${({ dir }) => (dir === 'ltr' ? `float: left;` : `float: right;`)}
+  ${({ script }) => {
```
DenisHdz: Since this code is in two places, can we create a function passing the groupX.lineheight and just call it in here?

```
const getSideLength = groupLineHeight => {
  const sideLength = `${calculateButtonSide(groupLineHeight)}rem;`;
  return `
    height: ${sideLength};
    width: ${sideLength};    
  `;
};
```

```
 ${({ script }) => {
    return script && getSideLength(script.pica.groupA.lineHeight);
  }}
```

Bopchy:
Other than what Denis has mentioned, everything looks good to me 👍

Bopchy (approved):

DenisHdz (approved):

PR #2771, opened 2019-12-04
Use a react button to spread styled-component props

AlistairGempf:
Resolves #NUMBER

**Overall change:** Use a react button to spread styled-component props, preventing styled-components from filtering it out as a non-standard attribute. This is required for the `on` attribute required for the Amp Menu Button. For some reason this only happens on the SSR though, so it doesn't show up as missing on any of the tests or storybook, only once we try and pull it into Simorgh.

**Code changes:**

- Use a react button to spread styled-component props

---


tochwill (approved):

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
```diff
@@ -127,7 +127,10 @@ const getButtonDimensions = lineHeight =>
   `height: ${calculateButtonSide(lineHeight)}rem;
   width: ${calculateButtonSide(lineHeight)}rem;`;
 
-const MenuButton = styled.button`
+// eslint-disable-next-line react/prop-types
+const Button = ({ script, ...props }) => <button type="button" {...props} />;
+
+const MenuButton = styled(Button)`
```
ghost, line 141: Another solution would be to have `attrs` on `MenuButton`
```suggestion
const MenuButton = styled.button.attrs({type: 'button' })
```

AlistairGempf: This does work for the `type` attribute, but doesn't work for the `on` attribute because styled components filters it out as it's not a standard attribute.

For some reason this only happens on the SSR though, so it doesn't show up as missing on any of the tests or storybook, only once we try and pull it into Simorgh.

DenisHdz: Now I understand Drew's question.

DenisHdz (approved):

packages/components/psammead-navigation/src/DropdownNavigation/index.jsx
ghost: Do you mind adding a comment explaining why we are doing this. I think it would be helpful for anyone looking at this piece of code in the future.

ghost (approved):

PR #2785, opened 2019-12-06
Add `id`,  `ampOpenClass` props to Navigation

Bopchy:
Resolves #NUMBER

**Overall change:** _Adds `id`,  `ampOpenClass` props to Navigation component._

**Code changes:**
- _Adds `id`,  `ampOpenClass` props._ 
- _Add test._
---

### DIFF
--- a/packages/components/psammead-navigation/src/index.jsx
+++ b/packages/components/psammead-navigation/src/index.jsx
@@ -217,9 +217,19 @@ NavigationLi.defaultProps = {
   dir: 'ltr',
 };
 
+// ampOpenClass is the class added to the Navigation, and is toggled on tap.
+// It indicates whether the menu is open or not. This overrides the background
+// color of the Navigation
 const StyledNav = styled.nav`
   position: relative;
   ${({ isOpen }) => `background-color: ${isOpen ? C_EBON : C_POSTBOX};`}
+  ${({ ampOpenClass }) =>
+    ampOpenClass &&
+    css`
+      &.${ampOpenClass} {
+        background-color: ${C_EBON};
+      }
+    `}
   border-top: 0.0625rem solid ${C_WHITE};
 
   ${StyledListItem} {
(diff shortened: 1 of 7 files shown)


AlistairGempf (approved):

packages/components/psammead-navigation/src/index.jsx
```diff
@@ -217,9 +217,19 @@ NavigationLi.defaultProps = {
...
+// ampOpenClass is the class added to the Navigation, and is toggled on tap.
+// It indicates whether the menu is open or not. This overrides the background
+// color of the Navigation
 const StyledNav = styled.nav`
   position: relative;
   ${({ isOpen }) => `background-color: ${isOpen ? C_EBON : C_POSTBOX};`}
+  ${({ ampOpenClass }) =>
+    ampOpenClass &&
+    css`
+      &.${ampOpenClass} {
```
DenisHdz: Just wondering if we need this `&`?

Bopchy: The `&` is required because without it we will be referring to the children of the `<StyledNav />`, and not the `<StyledNav />` itself - so the background colour override would not work.

DenisHdz: My understanding is that you want to apply the `background-color: ${C_EBON};` when `ampOpenClass` is passed to the `StyledNav`, right? In that case, I don't think we need `&.${ampOpenClass} {` and just with
```css`background-color: ${C_EBON};`}``` should work?

Bopchy: Yes, you're right. It works without `&.${ampOpenClass}`. Made that change

DenisHdz (approved):

ghost:
What is the `id` mean't to be used for?

AlistairGempf:
@sadickisaac the id is used to add an ID to the navigation, allowing simorgh to target it to toggle a class on it when on amp.

ghost:
Should it then be added in `simorgh` as part of extra props. Because as it is, we are not making use of it in the component It is meant to be used somewhere else in `simorgh`. It therefore makes sense to have it where it is used.
### What I propose
- Have `Navigation` spread extra props like we do on `NavigationUl` or `Brand`

Bopchy:
Had to revert the changes talked about https://github.com/bbc/psammead/pull/2785#discussion_r355347502 because the class override was not working as expected in Simorgh. The styling to override the background colour would not change with the toggled presence of the `ampOpenClass`, but instead would always be present making the Navigation appear as pictured when closed
 
image (2)

Made this change here https://github.com/bbc/psammead/pull/2787/commits/49edf27645bb264c5eb5e61c635da5833a52c8c5 as I thought it would be easier than opening a PR for this small tweak

Bopchy:
https://github.com/bbc/psammead/pull/2785#issuecomment-563192989 is a good point. Taken note of this, and will be making the change to spread extra props in `<Navigation />` in a subsequent PR

**Update**: This has been done here - https://github.com/bbc/psammead/pull/2802

PR #2787, opened 2019-12-06
Update useMediaQuery hook to ensure server/client markup is consistent

tochwill:
Resolves #NUMBER

**Overall change:** The useMediaQuery hook was returning different values on the initial render depending on whether the component was on the server or the client - this was causing the react app to hydrate incorrectly. This ensures the initial value is the same, and is then updated when the component mounts on the client.

**Code changes:**

---

### DIFF
--- a/packages/components/psammead-navigation/CHANGELOG.md
+++ b/packages/components/psammead-navigation/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 6.0.0-alpha.10 | [PR#xxxx](https://github.com/bbc/psammead/pull/xxxx) Update useMediaQuery hook to ensure server/client markup is consistent |
 | 6.0.0-alpha.9 | [PR#2779](https://github.com/bbc/psammead/pull/2779) Merge AMP tap handlers correctly |
 | 6.0.0-alpha.8 | [PR#2778](https://github.com/bbc/psammead/pull/2778) Fix menu border in FireFox high contrast mode |
 | 6.0.0-alpha.7 | [PR#2771](https://github.com/bbc/psammead/pull/2771) Spread the `on` prop (and others) onto the button, and set AMP Nav menu button initial state |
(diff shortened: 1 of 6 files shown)


packages/components/psammead-navigation/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 6.0.0-alpha.10 | [PR#xxxx](https://github.com/bbc/psammead/pull/xxxx) Update useMediaQuery hook to ensure server/client markup is consistent |
```
tochwill:
```suggestion
| 6.0.0-alpha.10 | [PR#2787](https://github.com/bbc/psammead/pull/2787) Update useMediaQuery hook to ensure server/client markup is consistent |
```

AlistairGempf (approved):

OlgaLyubin (approved):

packages/components/psammead-navigation/src/hooks/useMediaQuery.js
```diff
@@ -1,18 +1,18 @@
 import { useState, useEffect } from 'react';
 
 const useMediaQuery = query => {
-  const isClient = typeof window !== 'undefined';
+  const [matches, setMatches] = useState(null);
```
tochwill, line 4: A brief explanation into why this change was made, but I can explain in person on Monday too. Previously, this hook was returning `undefined` on the server render, and `true` or `false` on the initial client render. This meant that the `tabIndex` for navigation links was being set to `null` on the server, and `-1` on the client - so there was a slight difference in markup. This confused react during hydration, so the `tabIndex` was not set to -1, as it needed to be. This change means that when the component initially mounts on the client, matches is still `null`, and when `useEffect` runs on the client, it updates the value to either `true` or `false`. This was the desired effect with the original code, but this ensures that the server and initial client markup matches.

DenisHdz: Thanks for the explanation @tochwill :)

DenisHdz (approved):

PR #2801, opened 2019-12-09
Move media query hooks to Simorgh

AlistairGempf:
Resolves #NUMBER

**Overall change:** Remove the media query hook so that it can be added in simorgh to allow for other actions to be triggered on a media query.

**Code changes:**

- Remove media query hooks (to be added back in simorgh to allow for other parts of the nav to be controlled based on the media query)
- Add an `isScrollable` prop to be set in Simorgh based on the media query hook

---


tochwill (approved):

Bopchy (approved):
LGTM 👍

packages/components/psammead-navigation/src/index.stories.jsx
```diff
@@ -146,6 +147,16 @@ const navigationStory = (
     ? AmpScrollableNavigation
     : CanonicalScrollableNavigation;
 
+  const [isScrollable, setIsScrollable] = useState(false);
```
DenisHdz: Is there any difference setting it by default as `null` or `false`? Just asking because in the hooks we had `null`.

packages/components/psammead-navigation/src/index.stories.jsx
```diff
@@ -146,6 +147,16 @@ const navigationStory = (
...
+
+  useEffect(() => {
+    const mediaQueryList = window.matchMedia('(max-width: 37.5rem)');
+    setIsScrollable(mediaQueryList.matches);
+    mediaQueryList.addListener(event => {
+      setIsScrollable(event.matches);
+    });
```
DenisHdz: With this refactor we don't have the `removeListener` method anymore.

```
    const handler = event => setMatches(event.matches);

    mediaQueryList.addListener(handler);
    return () => mediaQueryList.removeListener(handler);
```

AlistairGempf: Good point, I'll add this back in for the story to prevent the storybook slowing down due to there being a load of extra listeners that aren't being used anymore.

packages/components/psammead-navigation/src/index.stories.jsx
AlistairGempf: The only place it gets used is in `isScrollable && ...` statements, so using `false` will lead to the same results as `null`.

DenisHdz (approved):
LGTM

PR #2860, opened 2019-12-30
update props for most read

thekp:
Related to: https://github.com/bbc/simorgh/issues/4986

**Overall change:** _Update the items props to include a timestamp and an id value._

**Code changes:**

- _Move the prop type validation into mostReadHelper, to make the `src/Item/index.jsx` less messy._
- _Added a random number generator for the key value of each item mapped from `src/List/index.jsx`, when real data is available it would use the actual id provided from the API._
- updated the readme to reflect the changes to props.
- removed some props that were not being used

---


packages/components/psammead-most-read/README.md
```diff
@@ -23,7 +23,7 @@ The `MostRead` component is designed to display the most read articles given a d
 <!-- prettier-ignore -->
 | Argument | Type | Required | Default | Example |
 | -------- | ---- | -------- | ------- | ------- |
-| items | array | yes | N/A | `[{ title: 'This is a headline', href: 'https://www.bbc.com' }]`
+| items | array | yes | N/A | `[{ id: 1, title: 'This is a headline', href: 'https://www.bbc.com', timestamp: null }]` |
```
DenisHdz: Could you put another example instead of null?

packages/components/psammead-most-read/src/testHelpers/itemsHelper.jsx
```diff
@@ -0,0 +1,47 @@
...
+  Array.from({ length: arraySize }, () => getItem(service));
+
+export const itemPropTypes = shape({
+  id: number.isRequired,
+  title: string.isRequired,
+  href: string.isRequired,
+  timestamp: node,
+});
+
+export const itemsPropTypes = arrayOf(
```
DenisHdz: Do we need this? Can't we just do `items: arrayOf(itemPropTypes).isRequired` in `MostReadRankWrapper` and `MostReadList`?

packages/components/psammead-most-read/README.md
rhenshaw56: If we expect `null` values, can we have this as an optional prop instead and default it to `null`

thekp: The reason I've added it as an optional prop here is that we can have 10 items in our array so there could possibly be 10 timestamp values and we would like to display the timestamp for that item if it is greater than 60days (I plan on having this 60days logic in Simorgh and passing the timestamp container a prop here 👀). Because the timestamp is tied to the item, I think to have it as a separate prop will make difficult to handle.

rhenshaw56: thanks @thekp makes sense

DenisHdz (approved):
LGTM 👍

rhenshaw56 (approved):

PR #2869, opened 2020-01-06
update most read prop structure

thekp:
Related to https://github.com/bbc/simorgh/issues/4742
example of how this PR is tied into AMP: https://github.com/bbc/simorgh/pull/4980/files#diff-8d1c24b698c68492c855f7257b908b52R42-R85

**Overall change:** _Update the way we handle the props for the most read items, as we a restricted to having to loop through each item in AMP, we have to handle each item separately, hence this change._

**Code changes:**

- _Updated most-read link to act as a parent wrapper that applies styles independently to each item._
- _split most-read rank into a separate file_
- _updated most-read list to accept children as a prop_
- _Updated tests and storybook to work with new prop structure change._
- update readme

---

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Automated jest tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

PR #2870, opened 2020-01-06
update timestamp prop to not be required

thekp:
Resolves n/a

**Overall change:** raising this PR to fix the issue here: https://github.com/bbc/simorgh/pull/5023#discussion_r362871432

**Code changes:**

- _update timestamp prop to not be required so don't have to pass a value to it._

---

### DIFF
--- a/packages/components/psammead-most-read/CHANGELOG.md
+++ b/packages/components/psammead-most-read/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 0.4.0-alpha.2 | [PR#2860](https://github.com/bbc/psammead/pull/2717) Update timestamp prop to not be required |
 | 0.4.0-alpha.1 | [PR#2860](https://github.com/bbc/psammead/pull/2717) Update props for MostReadItem and update Readme |
 | 0.3.0-alpha.1 | [PR#2717](https://github.com/bbc/psammead/pull/2717) Update MostReadList with UX changes and various fixes |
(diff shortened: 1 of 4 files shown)


packages/components/psammead-most-read/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 0.4.0-alpha.2 | [PR#2860](https://github.com/bbc/psammead/pull/2717) Update timestamp prop to not be required |
```
OlgaLyubin:
```suggestion
| 0.4.0-alpha.2 | [PR#2870](https://github.com/bbc/psammead/pull/2870) Update timestamp prop to not be required |
```

packages/components/psammead-most-read/src/testHelpers/itemsHelper.jsx
```diff
@@ -34,5 +34,5 @@ export const itemPropTypes = shape({
   id: string.isRequired,
   title: string.isRequired,
   href: string.isRequired,
-  timestamp: node.isRequired,
+  timestamp: node,
```
OlgaLyubin, line 37: Should there be a default value for this since it is not required anymore?

EinsteinNjoroge (changes requested):
We need to update the readme as well.

thekp:
after talking on Slack with Einstein, there is no clean way of us defaulting the timestamp value with the current implementation so we are going to push forward with this approach instead: https://github.com/bbc/psammead/pull/2869

PR #2399, opened 2019-10-15
Layout iteration - Leading story promo 1008px+ layout

Bopchy:
Resolves #2208 

**Overall change:** _Implement layout iteration designs for the leading story promo for screen widths of 1008px+._

**Code changes:**

- _Create psammead-leading-story-promo component._

---

- [x] I have assigned myself to this PR and the corresponding issues
- [x] Automated jest tests added (for new features) or updated (for existing features)
- [x] This PR requires manual testing

PR #2929, opened 2020-01-13
Update leading story styling

AlistairGempf:
Resolves #-

**Overall change:** 
Previously, the grid was only applied to larger breakpoints. This meant on smaller breakpoints the story promo was not filling the width of the screen.

Previously:
image
On this branch:
image

**Code changes:**

- Apply the grid styling across all breakpoints
- Add `position: relative` to wrapper css to prevent faux-block-link from taking over the whole page

---


packages/components/psammead-leading-story-promo/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 | Version | Description |
 |---------|-------------|
+| 0.1.0-alpha.7 | [PR#2929](https://github.com/bbc/psammead/pull/2929) Use grid styling across all breakpoints and use position relative |
```
sareh:
```suggestion
| 0.1.0-alpha.8 | [PR#2929](https://github.com/bbc/psammead/pull/2929) Use grid styling across all breakpoints and use position relative |
```

sareh (approved):
👍 Looks good.

DenisHdz (commented):
We were applying the grid only to larger breakpoints because this component is meant to be used just on the new Grid layout above 1008+px, but it kind of makes sense to keep it consistent across all the breakpoints https://github.com/bbc/psammead/pull/2399.

On that note, I have just realized testing it on IE, that there is no separation between the headline and the image so sometimes it can result in the following as there is not fallback for the `grid-column-gap`:

leading-story-promo-ie

AlistairGempf:
@DenisHdz we were thinking it would be a pain to change which component gets rendered depending on breakpoint so we'll probably have to have this consistent across all breakpoints (but we're going to discuss with UX).

Good point on the fallback - I've added that in now using padding in the fallback.

j-pendlebury (approved):
looks good to me 👍

j-pendlebury:
Apparently linting is complaining (which is why Jenkins is being a 💩)

```
packages/components/psammead-leading-story-promo/src/index.jsx

 33:3  ✖  Unknown word   CssSyntaxError
```

j-pendlebury (changes requested):
linting needs fixing

packages/components/psammead-leading-story-promo/src/index.jsx
```diff
@@ -29,6 +30,7 @@ const ImageGridColumns = css`
 
 const ImageGridColumnsFallback = css`
   width: ${fourOfSixColumnsMaxWidthScaleable};
+  ${({ dir }) => (dir === 'ltr' ? 'padding-left' : 'padding-right')}: 0.5rem;
```
DenisHdz:
```suggestion
  ${({ dir }) => (dir === 'ltr' ? 'padding-left' : 'padding-right')}: ${GEL_SPACING};
```

OlgaLyubin (commented):
Looks good! Just needs a README update in the "props" sections since there is a `dir` prop now

packages/components/psammead-leading-story-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -127,9 +139,11 @@ exports[`LeadingStoryPromo should render RTL leading story promo correctly 1`] =
 
 <div
   class="c0"
+  dir="ltr"
```
OlgaLyubin: In a test "LeadingStoryPromo should render RTL leading story promo correctly" you should pass `dir="rtl"` since right now in a snapshot for RTL there is a wrong direction

AlistairGempf: Good spot! Should be done now.

j-pendlebury (approved):

OlgaLyubin (approved):
Looks good!

AlistairGempf:
This will be resolved and superceded by #2937 so I'm going to close this PR.

PR #2935, opened 2020-01-14
Add clock icon for Radio Schedules

### LINKED ISSUE #2932: Radio Schedules: add clock icon to psammead-assets
Add a clock icon to the psammead-assets so that it can be used in Radio Schedules.

Screenshot 2020-01-14 at 11 46 17

Get a clock icon from UX team, optimise the svg using this tool: https://jakearchibald.github.io/svgomg/ . And then convert it from svg to JSX using this one: https://transform.tools/ .

N/A

**Testing notes**
Check the icon in Psammead Storybook.

OlgaLyubin:
Resolves #2932 

**Overall change:** 
Added a clock icon for Radio Schedules to the `psammead-assets`.

**Code changes:**
- Added clock icon.

---


EinsteinNjoroge (approved):
👍

Bopchy (approved):

packages/utilities/psammead-assets/README.md
```diff
@@ -56,7 +56,7 @@ import { coreIcons } from '@bbc/psammead-assets/svgs';
 
 ## Media Icons SVGs
 
-Media icons is an object containing styled SVG icons for video, audio and photogallery. Media icons are sized to work well alongside specific text with typography group `GEL Minion`. They are used in `psammead-media-indicator` component.
+Media icons is an object containing styled SVG icons for video, audio, photogallery, guidance and clock. Media icons are sized to work well alongside specific text with typography group `GEL Minion`. They are used in `psammead-media-indicator` component.
```
simonsinclair: Seeing as we're updating this paragraph, do you think we should remove (or update) the line "They are used in `psammead-media-indicator` component." to better reflect the current use of these icons?

packages/utilities/psammead-assets/src/svgs/mediaIcons.jsx
```diff
@@ -61,17 +70,29 @@ const mediaIcons = {
...
+    <GuidanceIcon viewBox="0 0 32 32" width="32" height="32" {...defaultAttrs}>
       <path d="M32,16A16,16,0,1,1,16,0,16,16,0,0,1,32,16Zm-8.6-.8H16.6v3.1h2.9c-.2,1.5-1.6,2.4-3.7,2.4s-3.7-2.4-3.7-4.7,1-4.7,3.7-4.7,3,.8,3.2,2.2h4.2c-.4-4-3.9-5.8-7.2-5.8-4.8,0-8,3.7-8,8.3s3.1,8.3,8,8.3a5.4,5.4,0,0,0,4.5-2.1l.2,1.7h2.7Z" />
     </GuidanceIcon>
   ),
+  clock: (
+    <ClockIcon
+      viewBox="0 0 13 13"
+      width="13"
+      height="13"
+      aria-hidden="true"
```
simonsinclair: I completely understand why this attribute has been added, because we're constantly wrapping our icons in containers, which have this attribute. 😅 However:
- This affords us more control when it comes to using the icons and choosing what's hidden from assistive technology and what's not.
- If we're adding this to one icon, we should probably add it to the rest for consistency and that's probably a discussion we should have with the wider team.

simonsinclair (commented):
Nice work, Olga - the clock looks great in Storybook. You've even tidied up code in places too.
I left a some comments on a couple of minor things below.

packages/utilities/psammead-assets/README.md
OlgaLyubin: Good point! I will add that guidance is used for media-player. But I am a bit unsure about saying that a clock is used for radio schedules since that component has not been developed yet.

packages/utilities/psammead-assets/src/svgs/mediaIcons.jsx
OlgaLyubin: From my conversation with @greenc05 I understand that all icons should have `aria-hidden="true"` attribute, so I will add them to other media icons too.

greenc05: As long as asssistive tech can't access the SVG I don't think it matters if aria-hidden true is applied to a container wrapping the SVG or directly to the SVG tag, personally I would prefer to apply it directly to the SVG as that's what I've tested and know it works, though if it's better from a code writing point of view applying it to a wrapping container is also fine - no one has flagged any issues from swarms doing it this way

simonsinclair: We could add it at this level. But I think we have to be careful about blindly applying it to every icon. If we have an accessible icon down the line that uses a [title](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) element for example, it would fail to be accessible with this attribute applied. This is where the container > icon pattern serves us well.

greenc05: We don't use the title element anywhere, using visually hidden text has greater support

simonsinclair: If our icons will always be hidden from assistive tech. then this could be a good route to go down.

greenc05: I can't guarantee that we will always hide our icons but that's what I recommend currently...

simonsinclair (approved):

PriyaKR:
The icon looks slightly not circular in firefox browser.If UX are fine i am fine with it.
Screenshot 2020-01-20 at 12 42 15

OlgaLyubin:
@PriyaKR I have checked the icon in FireFox 72.0.1 and was not able to reproduce the issue

Screenshot 2020-01-20 at 14 24 43

PriyaKR:
I am still seeing that issue may be we can merge this and see how it appears on simorgh pages once we use it in radio schedule components.

PR #2937, opened 2020-01-14
Story promo generalisation

sareh:
Resolves #2931

**Overall change:** Generalises the Story Promo 

**Code changes:**

- Remove`topStory` prop from StoryPromo - replace with `promoType`
- Move `Image` and `Text` styles to different files
- Refactor headline styles
- Update headline styling to work for IE (removes nested media queries)
- Add and clean up stories
- Update prop in tests from `topStory` to `promoType`
- Remove obsolete test (top story without image) and update snapshots
- Integrate Leading Story Promo in the `psammead-story-promo` package
- Update README
- Remove `pammed-leading-story-promo` package

---

- [x] I have assigned myself to this PR and the corresponding issues
- [x] Automated jest tests added (for new features) or updated (for existing features)
- [X] This PR requires manual testing

PR #2938, opened 2020-01-15
Add program card component to radio-schedule

### LINKED ISSUE #2924: Radio Schedules: Generalise Bulletin component to allow for programme card designs
The radio schedule programme card contains many of the same elements and behaviours of the bulletin component that is currently being used. Make sure we can make this component generic across both, and if we can, then update the component to allow it to be used in both situations. If we can't, see how we can reuse common elements across both components.

- Ensure we can recolour the svg to blue (to allow for the 'next' state) in a similar manner to the way we recolour svgs at the moment.
- Consider being able to pass the heading as a child, so the component doesn’t need all of the code for all of the states (for example, don’t have booleans for ‘live’ and ‘next’ states, pass in the correct heading from the parent). 
- Have individual storybook stories for different states, and unit tests in Psammead.
- The whole card should be a link, as the bulletin component currently is. this is currently being done for story-promo here: https://github.com/bbc/psammead/blob/latest/packages/components/psammead-story-promo/src/index.jsx#L316-L326 
(we would need to add this code snippet to our link and a position relative to the element that we want to be included in the link area)

This doesn't include the timestamp above the card:

Screenshot 2020-01-14 at 11 13 35

Building a new, separate, radio schedule programme card component.

**Testing notes**
Expected outcome is either generalised component which supports bulletin and radio schedule component or separate one for both.

**Additional context**
See designs here: https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/5dc0724999d4a3572b675557

Screen reader UX can be seen here: https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/5dc0724850731722a4aa767a

Pick this up as a pair for the initial investigation.

EinsteinNjoroge:
Resolves #2924 

**Overall change:** _The radio schedule pages are designed to show program-cards with the different states, brand titles, episode titles, summary, and duration. This PR adds and exports a `ProgramCard` component from the `psammead-radio-schedule` package. You can find the [chromatic storybook builds here](https://www.chromaticqa.com/snapshot?appId=5d28eb5ee163f6002046d6fb&id=5e269bb9526d380020b804dc)._

**Code changes:**
in this approach, we have created an entirely new component that does not extend `psammead-bulletin`.

- _Create and export the program card component._
- _Add snapshot tests and storybook stories._

**Screenshots**
**_live_**
Screen Shot 2020-01-21 at 9 34 16 AM

**_onDemand_**
Screen Shot 2020-01-21 at 9 34 03 AM

**_Next_**
Screen Shot 2020-01-21 at 9 33 53 AM

---


packages/components/psammead-program-card/src/index.jsx
```diff
@@ -0,0 +1,169 @@
...
+  script,
+  heading,
+  summary,
+  date,
+  duration,
+  state,
+  ctaLink,
+}) => (
+  <CardWrapper>
+    <Link href={ctaLink}>
```
tochwill: According to the accessibility requirements here: https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/5dc0724850731722a4aa767a, we should just be wrapping the heading in a link, and using the faux link block pattern that this `Link` component (and the bulletin component) uses to make the whole card clickable.

packages/components/psammead-program-card/src/index.jsx
```diff
@@ -0,0 +1,169 @@
...
+      <TextWrapper dir={dir}>
+        <HeadingWrapper
+          service={service}
+          script={script}
+          {...programStateColors[state]}
+        >
+          {state !== 'onDemand' && (
+            <LabelWrapper
```
tochwill: Hmm, we already have a LiveLabel in the story-promo package, that's also being used in the Bulletin component. The Link component from story-promo is also being used both in the story promo, bulletin, and in here. Should we extend the LiveLabel to include this Next state as well, and pull this shared functionality out of story promo as it's used in 3 places now?

packages/components/psammead-program-card/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@bbc/psammead-program-card",
```
tochwill: I'm wondering if it would be easier for us to create a `psammead-radio-schedule` package that contains the program card, the timestamp, and the wrapping components to begin with, and then split things out into separate packages if it becomes apparent later that we can make them generic and reuse them across new components? Probably something to discuss with the team.

packages/components/psammead-program-card/src/index.jsx
```diff
@@ -0,0 +1,169 @@
...
+              service={service}
+              script={script}
+              {...programStateColors[state]}
+            >
+              {`${state} `}
+            </LabelWrapper>
+          )}
+          {heading}
+        </HeadingWrapper>
+        <DateWrapper service={service} script={script}>
```
tochwill: Sorry it's not particularly clear from the designs but this might not be a date! It's the title of the episode, it just happens that a lot of the examples are a date.

packages/components/psammead-program-card/src/index.jsx
```diff
@@ -0,0 +1,169 @@
...
+`;
+
+const DateWrapper = styled.div`
+  color: ${C_SHADOW};
+  padding: ${GEL_SPACING} 0;
+  ${({ service }) => service && getSansRegular(service)};
+  ${({ script }) => script && getPica(script)};
+`;
+
+const SummaryWrapper = styled(Paragraph)`
```
tochwill: If we're overriding all of the styles, should we just be using a `p` tag here rather than overriding the styles from the `Paragraph` component? The bulletin also uses a paragraph with different styling from the `Paragraph` component.

packages/components/psammead-program-card/src/index.jsx
```diff
@@ -0,0 +1,169 @@
...
+  getSansBold,
+} from '@bbc/psammead-styles/font-styles';
+import {
+  getLongPrimer,
+  getMinion,
+  getPica,
+} from '@bbc/gel-foundations/typography';
+import { Link } from '@bbc/psammead-story-promo';
+
+const programStateColors = {
```
tochwill: Is it worth defining these states as constants and then exporting them from this package too? So then we can make sure we're passing in a valid state that matches what we expect in here.

packages/components/psammead-program-card/src/index.jsx
EinsteinNjoroge: Noted, I'll check with UX to get some designs that are not entirely dates.

packages/components/psammead-program-card/src/index.jsx
EinsteinNjoroge: That approach makes sense, we can export them from this package for now. We might need to move them to a utility file in the future.

packages/components/psammead-program-card/src/index.jsx
EinsteinNjoroge: If we use `LiveLabel` from story-promo, we'll end up overriding all of it's styling since this card's Label uses a different set of colors, margins, and typography.

packages/components/psammead-program-card/src/index.jsx
tochwill: Thinking about it again, we could just use `oneOf` in propTypes to make sure we pass in a valid state! This is done for the `dir` prop, and now in the story promo package as well  (https://github.com/bbc/psammead/pull/2937/files#diff-cfe98725ca3d73fa9b4b1936998b8977R138) - sorry, I didn't think about this when I reviewed before!

packages/components/psammead-program-card/package.json
EinsteinNjoroge: After discussions, we've decided to have a `psammead-radio-schedule` package that exports a `program-card` component among others.

packages/components/psammead-program-card/src/index.jsx
EinsteinNjoroge: Currently, we do not have designs that show an episode title not in a date format. We've made reservations for this by adding necessary padding to accommodate a multiline episode title

packages/components/psammead-radio-schedule/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Radio Schedule Changelog
+
+| Version | Description |
+|---------|-------------|
+| 0.1.0-alpha.1 | [PR#2979](https://github.com/BBC-News/psammead/pull/2979) Initial creation of package. |
+| 0.1.0-alpha.2 | [PR#2938](https://github.com/BBC-News/psammead/pull/2938) Add radio program card component. |
```
thekp, line 5: we should fix the ordering here so it's like this:
```
| 0.1.0-alpha.2 | [PR#2938](https://github.com/BBC-News/psammead/pull/2938) Add radio program card component. |
| 0.1.0-alpha.1 | [PR#2979](https://github.com/BBC-News/psammead/pull/2979) Initial creation of package. |
```

thekp: or do we even need the  0.1.0-alpha.2 🤔

thekp (commented):
looks good to me, just confused about the changelog.

packages/components/psammead-radio-schedule/CHANGELOG.md
EinsteinNjoroge: Thanks for pointing this out, I created `0.1.0-alpha.1` on a separate branch so that guys working on `radio-schedules` could have a common ground, I just pulled that branch into this one.

packages/components/psammead-program-card/src/index.jsx
thekp: since this value might not be a date, is it worth changing it to like `subHeading`?

EinsteinNjoroge: We've already changed this to `TitleWrapper` which can span multiple lines.

thekp: ah yes, I see that now. nice one 💯

thekp (approved):

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,191 @@
...
+        {summary}
+      </SummaryWrapper>
+    </TextWrapper>
+    <ButtonWrapper
+      dir={dir}
+      service={service}
+      script={script}
+      {...programStateConfig[type]}
+    >
+      <IconWrapper dir={dir} {...programStateConfig[type]}>
```
rhenshaw56: why is the `IconWrapper` and `ButtonWrapper` not making use of the `dir` prop, even though it's being passed in to both ??

EinsteinNjoroge: Nice catch.
Initially, we were applying paddings conditionally based on the `dir`. We no longer need to pass this prop to those components

itimirichard: Thanks @rhenshaw56 . This is has been fixed.

rhenshaw56 (approved):

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,956 @@
...
+      class="c2"
+      href="/news/articles/cn7k01xp8kxo"
+    >
+      <h3
+        class="c3"
+      >
+        <span
+          class="c4"
+        >
+          live
```
greenc05: Uppercase? Are we covering the requirement for 'LIVE' to be read out with an english accent here? Also we usually nest the link inside the h3, can we do the same here please?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,956 @@
...
+        >
+          live 
+        </span>
+        Could a computer ever create better art than a human?
+      </h3>
+    </a>
+    <div
+      class="c5"
+    >
+      29/01/1990
```
greenc05: This content should be inside the h3 > Link - Have you seen the screen reader ux?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,956 @@
...
+      The critic, author, poet and TV host was known for his witty commentary on international television.
+    </p>
+  </div>
+  <div
+    class="c7"
+  >
+    <span
+      class="c8"
+    >
+      <svg
```
greenc05, line 261: aria-hidden?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,956 @@
...
+    >
+      29/01/1990
+    </div>
+    <p
+      class="c6"
+    >
+      The critic, author, poet and TV host was known for his witty commentary on international television.
+    </p>
+  </div>
+  <div
```
greenc05, line 255: Just checking if this content will be in the faux block link area? It should be if it's not...

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,956 @@
...
+      class="c2"
+      href="/news/articles/cn7k01xp8kxo"
+    >
+      <h3
+        class="c3"
+      >
+        <span
+          class="c4"
+        >
+          next
```
greenc05: Next items are not links and the card should not be clickable. Also this should be in uppercase?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
greenc05: Same goes for the other examples below...

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,956 @@
...
+        <path
+          d="M10.596.199l-.525.562c1.35 1.35 2.175 3.225 2.175 5.25s-.825 3.9-2.175 5.25l.525.525c1.5-1.462 2.4-3.525 2.4-5.775s-.9-4.312-2.4-5.812zM6.996 1.511l-2.25 2.25H.996v4.5h3.75l2.25 2.25z"
+        />
+      </svg>
+    </span>
+    <span
+      class="c10"
+      dir="ltr"
+    >
+      30:00
```
greenc05: Are we passing this to the h3 > link as visually hidden text? Please see screen reader UX...

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,956 @@
...
+        <path
+          d="M10.596.199l-.525.562c1.35 1.35 2.175 3.225 2.175 5.25s-.825 3.9-2.175 5.25l.525.525c1.5-1.462 2.4-3.525 2.4-5.775s-.9-4.312-2.4-5.812zM6.996 1.511l-2.25 2.25H.996v4.5h3.75l2.25 2.25z"
+        />
+      </svg>
+    </span>
+    <span
+      class="c9"
+      dir="ltr"
+    >
+      30:00
```
greenc05: We should be adding the word 'Duration' (translated) here - Please see screen reader UX... and we need the value to be in the usual format for screen reader users e.g. 30,00

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
EinsteinNjoroge: This already in the link area.

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
EinsteinNjoroge: We're passing the start-time to the H3 visually hidden text, this `duration` is being passed to a different visually hidden text at the bottom of the card

packages/components/psammead-radio-schedule/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+<ProgramCard
+  service="news"
+  script={latin}
+  dir="ltr"
+  brandTitle="This is a brand title"
+  summary="Could a computer ever create better art than a human?"
+  episodeTitle="29/01/1990"
+  duration="30:00"
+  state="live"
```
tochwill, line 58: `state` is a string in the usage example, but an object in the table of props above :)

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,250 @@
...
+  link,
+  translation,
+  brandTitle,
+  episodeTitle,
+  service,
+  script,
+  startTime,
+}) => {
+  const isOnDemand = type === 'onDemand';
+  const hiddenTextProps =
```
tochwill: Could we set the lang here based on the state being `live` rather than relying on the translation?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,250 @@
...
+const ProgramCard = ({
+  dir,
+  service,
+  script,
+  brandTitle,
+  summary,
+  episodeTitle,
+  duration: { durationValue, durationText },
+  startTime,
+  state: { type, translation },
```
tochwill: Could we flatten the props for state and duration, rather than passing in objects? I'm not sure it's too clear what the `translation` and `durationText` props are here. We could pass in the translations we need and use them, and keep the props that indicate the state of the card seperate?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,250 @@
...
+}) => {
+  const isOnDemand = type === 'onDemand';
+  const hiddenTextProps =
+    translation.toLowerCase() === 'live' ? { lang: 'en-GB' } : {};
+
+  const content = (
+    <>
+      {!isOnDemand && (
+        <VisuallyHiddenText {...hiddenTextProps}>
+          {`${translation}, `}
```
tochwill: I'm not really sure what `translation` is at this point! Maybe rename to what the translation is for?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,250 @@
...
+  link,
+}) => (
+  <CardWrapper>
+    <TextWrapper>
+      <HeadingWrapper
+        service={service}
+        script={script}
+        {...programStateConfig[type]}
+      >
+        {renderHeaderContent({
```
tochwill, line 194: What do people think about exporting the Heading and composing it into the card component as a child or a prop? It's nice to avoid 'prop drilling' (https://kentcdodds.com/blog/prop-drilling), but it looks like here the rendering of the header has just been separated into a function to make this component more readable, so may be overkill for now.

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,250 @@
...
+      service={service}
+      script={script}
+      {...programStateConfig[type]}
+    >
+      <IconWrapper {...programStateConfig[type]}>
+        {mediaIcons.audio}
+      </IconWrapper>
+      <DurationWrapper dir={dir}>
+        <VisuallyHiddenText>
+          {`${durationText} ${durationValue.replace(/:/g, ',')}`}
```
tochwill: Hmm... I'm wondering if we want to pass the duration in as a number and then format it in the component (so we could format in one way for presentation, and another for the screen reader text). Another thing I haven't considered before is do we format durations in the same way for all services? A question for Jon and the UX team maybe?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
EinsteinNjoroge: @tochwill The intention is to have the `lang` being set to `en-GB` only if we don't have a localized translation for the word `live`. 
Screen Shot 2020-01-24 at 2 40 53 PM

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
EinsteinNjoroge: Translation refers to the localised state type, I'll update to reflect that.
example: The live state in Arabic would look like`{ type: 'live', translation: 'مباشر'}`

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
EinsteinNjoroge: Sure, this makes sense.

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
EinsteinNjoroge: Updated

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
EinsteinNjoroge: You're right, the separation here was just to make the code more readable and to apply the `link` conditionally.

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,244 @@
...
+import { Link } from '@bbc/psammead-story-promo';
+import { oneOf, shape, string } from 'prop-types';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+import VisuallyHiddenText from '@bbc/psammead-visually-hidden-text';
+
+const CardWrapper = styled.div`
+  padding-top: ${GEL_SPACING};
+  background-color: ${C_WHITE};
+  position: relative;
+  border: 1px solid transparent;
```
Bopchy:
```suggestion
  border: 0.0625rem solid transparent;
```
Suggesting this for consistency - as through out psammead we tend to use `rem`

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,244 @@
...
+const ButtonWrapper = styled.div`
+  ${({ service }) => service && getSansRegular(service)};
+  ${({ script }) => script && getMinion(script)};
+  padding: ${GEL_SPACING};
+  background-color: ${({ backgroundColor }) => backgroundColor};
+  color: ${({ durationColor }) => durationColor};
+  border-top: 1px solid transparent;
+`;
+
+const IconWrapper = styled.span.attrs({ 'aria-hidden': 'true' })`
```
Bopchy: I think having `{ 'aria-hidden': 'true' }` specified here is repetitive, as it has been [specified on the media icon already](https://github.com/bbc/psammead/blob/b070cf7480941681c11834e6c64ad13f2efb2cc0/packages/utilities/psammead-assets/src/svgs/mediaIcons.jsx#L58)

Bopchy (commented):
Looks really good 👍. Left some comments

tochwill (approved):
If you haven't already, could you add a ticket to address the duration formatting that we discussed in slack? Thanks :)

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+    class="c1"
+  >
+    <h3
+      class="c2"
+    >
+      <a
+        class="c3"
+        href="/news/articles/cn7k01xp8kxo"
+      >
+        <span
```
greenc05, line 213: This should have the same implementation as we do for 'live' elsewhere? Does it but we just haven't got the exact code here? Live should be on the page twice, one version in uppercase that is aria-hidden and another version in Sentence case (e.g. 'Live, ') that is visually hidden with the lang attribute, as per https://bbc.github.io/psammead/?path=/story/components-storypromo-storypromo--regular-promo

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+          class="c4"
+        >
+          Could a computer ever create better art than a human?, 13:00, 29/01/1990
+        </span>
+        <span
+          aria-hidden="true"
+          class="c5"
+        >
+          LIVE
```
greenc05: why is this here with another headline below?? Please implement like we do in story promo

greenc05: Please can we have one headline and one episode title? we need to remember that our content should still be useable when CSS isn't available

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
greenc05: will the start time also be in this?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+      </svg>
+    </span>
+    <span
+      class="c11"
+      dir="ltr"
+    >
+      <span
+        class="c4"
+      >
+        duration 30,00
```
greenc05: Lets not forget to use sentence case please e.g. Duration 30,00

greenc05: And lets also use the time html element?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
greenc05: Ok, this is the duration, cool, not sure that requires visually hidden CSS?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+        class="c3"
+      >
+        next, 
+      </span>
+      <span
+        class="c3"
+      >
+        Could a computer ever create better art than a human?, 13:00, 29/01/1990
+      </span>
+      <span
```
greenc05: This isn't needed

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+        aria-hidden="true"
+        class="c4"
+      >
+        NEXT 
+      </span>
+      Could a computer ever create better art than a human?
```
greenc05: Headline and episode just once please as per the above similar comment

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+  <div
+    class="c1"
+  >
+    <h3
+      class="c2"
+    >
+      <span
+        class="c3"
+      >
+        next,
```
greenc05: Sentence case please... e.g. 'Next' rather than 'next'

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
greenc05: 30:00 should only be in the code once, we then might need mitigate possible text splitting in voiceover with role text

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
greenc05: we then might need mitigate possible text splitting in voiceover with role text

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+      </svg>
+    </span>
+    <span
+      class="c10"
+      dir="ltr"
+    >
+      <span
+        class="c3"
+      >
+        duration 30,00
```
greenc05: same as above

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+      class="c2"
+    >
+      <a
+        class="c3"
+        href="/news/articles/cn7k01xp8kxo"
+      >
+        <span
+          class="c4"
+        >
+          Could a computer ever create better art than a human?, 13:00, 29/01/1990
```
greenc05: As per above please

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+      </svg>
+    </span>
+    <span
+      class="c10"
+      dir="ltr"
+    >
+      <span
+        class="c4"
+      >
+        duration 30,00
```
greenc05: As per above

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1038 @@
...
+      >
+        <span
+          class="c4"
+        >
+          مباشر, 
+        </span>
+        <span
+          class="c4"
+        >
+          لماذا يخجل البعض من اسم قريته في مصر؟, 13:00, 29/01/1990
```
greenc05: Is the comma placement correct here?

packages/components/psammead-radio-schedule/README.md
```diff
@@ -0,0 +1,88 @@
...
+  durationLabel="duration"
+  state="live"
+  stateLabel="Live"
+  link="https://bbc.com/arabic/articles/c1er5mjnznzo"
+/>;
+```
+
+### When to use this component
+
+This component is to be used on the `radio` pages.
```
OlgaLyubin: Won't radio schedules be located on front pages?

greenc05:
The html that I'm expecting to see here is as follows:

Default/standard version:
```
<h3>
    <a href="">
        <span role="text">
           <span>Brand title</span><span class="VisuallyHiddenText">, 30:00, </span>
           <span>Episode title</span>
        </span>
    </a>
</h3>
```

Live version:
```
<h3>
    <a href="">
        <span role="text">
            <span aria-hidden="true">LIVE</span>
            <span lang="en-GB" class="VisuallyHiddenText">Live, </span>
            <span>Brand title</span><span class="VisuallyHiddenText">, 30:00, </span>
            <span>Episode title</span>
        </span>
    </a>
</h3>
```

Next version: (Might not need role text for the 'Next' version - would have to test - Update: Tested and we need role text here too)
```
<h3>
    <span role="text"> 
        <span>NEXT</span><span class="VisuallyHiddenText">, </span>
        <span>Brand title</span><span class="VisuallyHiddenText">, 30:00, </span>
        <span>Episode title</span>
    </span>
</h3>
```

And for the Duration:
```
<time>
    <span class="VisuallyHiddenText">Duration 30,00</span>
    <span aria-hidden="true">30:00</span>
</time>
```

greenc05:
And with no CSS I'm expecting those code examples to look like this...

Screen Shot 2020-01-27 at 16 09 14

packages/components/psammead-radio-schedule/README.md
EinsteinNjoroge: I'm not sure about this, maybe @tochwill can help.

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
EinsteinNjoroge: This placement follows the `[Brand title], [00:00], [Episode title]` convention.

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+  >
+    <h3
+      class="c2"
+    >
+      <a
+        class="c3"
+        href="/news/articles/cn7k01xp8kxo"
+      >
+        <span
+          class=""
```
greenc05, line 214: empty class??

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+            LIVE 
+          </span>
+          <span
+            class="c5"
+            lang="en-GB"
+          >
+            Live, 
+          </span>
+          <span
+            class=""
```
greenc05: empty class?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+      </svg>
+    </span>
+    <time
+      class="c11"
+      dir="ltr"
+    >
+      <span
+        class="c5"
+      >
+         Duration 30,00
```
greenc05, line 284: Do you know if we have the translation for 'duration' yet?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+  class="c0"
+>
+  <div
+    class="c1"
+  >
+    <h3
+      class="c2"
+    >
+      <span
+        class=""
```
greenc05, line 475: class?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+        role="text"
+      >
+        <span
```
greenc05, line 478: We don't need this aria hidden span, we only have 'live' twice to get the screen reader to announced the word 'live' correctly, this isn't needed here.

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+        >
+          NEXT 
+        </span>
+        <span
+          class="c4"
+        >
+          Next, 
+        </span>
+        <span
+          class=""
```
greenc05: class?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+  >
+    <h3
+      class="c2"
+    >
+      <a
+        class="c3"
+        href="/news/articles/cn7k01xp8kxo"
+      >
+        <span
+          class=""
```
greenc05, line 742: class?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+          role="text"
+        >
+          <span
+            class=""
```
greenc05: class?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1106 @@
...
+  >
+    <h3
+      class="c2"
+    >
+      <a
+        class="c3"
+        href="/arabic/articles/c1er5mjnznzo"
+      >
+        <span
+          class=""
```
greenc05, line 1022: class?

greenc05 (commented):
Thank you for making those changes @EinsteinNjoroge Just a couple of minor comments. Also please could you confirm that the spacing with no CSS is exactly the same as per the screen shot I added above? https://github.com/bbc/psammead/pull/2938#issuecomment-578821994

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
EinsteinNjoroge: We don't have translations for `duration` at the moment. The component allows for a localised duration label as demonstrated here https://github.com/bbc/psammead/pull/2938/files/ff393e2ef5532811206763b31a2af8d2f45ad0cc#diff-a46b8e623fc5362ce7634f4c72437b91R1095

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
EinsteinNjoroge: @greenc05 this is expected since we're using styled-components. You can find [similar snapshots here](https://github.com/bbc/psammead/blob/latest/packages/components/psammead-story-promo/src/__snapshots__/index.test.jsx.snap#L1252-L1255)

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,260 @@
...
+const HeadingWrapper = styled.h3`
+  ${({ service }) => service && getSerifMedium(service)};
+  ${({ script }) => script && getPica(script)};
+  color: ${({ headerTextColor }) => headerTextColor};
+  margin: 0; /* Reset */
+`;
+
+const HeadingContentWrapper = styled.span.attrs({ role: 'text' })``;
+
+const StyledSpan = styled.span``;
```
OlgaLyubin: What is the reason for having an empty `StyledSpan` styled component instead of just using a `<span>` where needed?

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,1111 @@
...
+  >
+    <h3
+      class="c2"
+    >
+      <span
+        class=""
+        role="text"
+      >
+        <span
+          aria-hidden="true"
```
greenc05: we don't want this to be hidden

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
EinsteinNjoroge: Fixed

packages/components/psammead-radio-schedule/README.md
tochwill: Hey - this is going to be on front pages

tochwill:
I'm going to pull this back into code review quickly and we can discuss the UX review process with alpha components in refinement!

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,261 @@
...
+  },
+  onDemand: {
+    backgroundColor: C_EBON,
+    headerTextColor: C_EBON,
+    titleColor: C_SHADOW,
+    durationColor: C_WHITE,
+  },
+};
+
+const sentenceCase = text =>
```
tochwill: Is this just used to capitalise the duration label? Would we be able to just pass in the translation string like this, we do that elsewhere :)

tochwill: Ah it's also used below for the live label text. We have to do exactly the same in the TV and Radio bulletin I believe? Can we follow the same approach in both places?

packages/components/psammead-radio-schedule/README.md
greenc05: It's also going to be on other pages that the media team are working on...

tochwill: Is there any value in specifying Simorgh pages in the README files? As these components shouldn't be tied to a particular Simorgh use case really?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
EinsteinNjoroge: We're using this for the visually-hidden state label as well,

EinsteinNjoroge: @tochwill we're using this on the **state label** as well as the **duration label**. If the state is live, we need to display localised  `LIVE` and have `Live` as a hidden text on the card. The Radio bulletin [has a hardcoded hidden-text](https://github.com/bbc/simorgh/blob/2b82268bbf23ebf961cdc31e47b59f2068b93647/src/app/containers/Bulletin/index.jsx#L89) `Live` regardless of the service.

tochwill: I think the requirement _should_ be the same for both - if we're hardcoding Live for all cases on the Bulletin and have the requirement to only do that if the translation is 'Live' for the radio schedule, we should check that this should be the case, I imagine they should be consistent. At the moment it looks like we're always displaying 'LIVE' as the visible text for the bulletin component too.

Let's check the following for bulletin and radio schedules are true:

Scenario 1 (LTR services):
- Show a visible 'LIVE' label in english
- Show visually hidden text with 'Live' read in an english accent

Scenario 2 (RTL services):
- Show a visible live label, in capitals, in the service language.
- Show visually hidden text with the translation of live read in the service language.

This is what I understood, but as you point out, currently in Simorgh we never pass a live label translation for bulletins even in RTL languages, so it defaults to 'LIVE', and the visually hidden text is hardcoded as 'Live'.

tochwill: Stupid question, but do our right to left scripts contain capital letters? I don't think arabic does?

tochwill: Another question, can we pass the translation of 'Live' in all cases in this format, and use `text-transform: uppercase` in the styles for the visible live label?

greenc05: "Show visually hidden text with the translation of live read in the service language." - this is not required for services that do not have 'live' in english. Though note, in this case the diplayed 'LIVE' (translated) could not have aria-hidden on it...

tochwill: Just had a conversation with Catharine:
- We only need visually hidden text for ‘Live’ in English as a hack to stop screen readers reading out ‘live’ incorrectly, we don’t need it for other languages.
- We _don’t_ want to read out ‘Live’ in English when we are using a different language, so it looks like the bulletin component is currently incorrect. We have no RTL languages using this component yet, so this won’t have impacted anyone.
- We can pass in translation strings as ‘Live’ and ‘Duration’, not need the sentenceCase function, and use CSS (or toUpperCase as you are) to display the make it ‘LIVE’.

EinsteinNjoroge: > Another question, can we pass the translation of 'Live' in all cases in this format, and use `text-transform: uppercase` in the styles for the visible live label?

This will cause UX and accessibility issues in `no-CSS `browsers

EinsteinNjoroge: I'm gonna move that function to a helper and use it in `tests` and `stories`

tochwill: Could you not just pass the translations strings in sentence case? You could still use toUpperCase for the visible label

packages/components/psammead-radio-schedule/README.md
tochwill:
```suggestion
This component is to be used initially on the `front pages`, and media pages.
```

tochwill: I don't think we need this section? But it's in all of the READMEs.

packages/components/psammead-radio-schedule/package.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/bbc/psammead/issues"
+  },
+  "homepage": "https://github.com/BBC-News/psammead/blob/latest/packages/components/psammead-radio-schedule/README.md",
+  "dependencies": {},
+  "peerDependencies": {
+    "react": "^16.12.0"
+  }
```
tochwill, line 32: Do we need to add this?

```
"publishConfig": {
    "tag": "alpha"
  }
```

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -0,0 +1,254 @@
...
+        <>
+          <LabelWrapper
+            service={service}
+            script={script}
+            {...labelWrapperProps}
+            {...programStateConfig[state]}
+          >
+            {`${stateLabel.toUpperCase()}`}
+          </LabelWrapper>
+          <VisuallyHiddenText {...hiddenTextProps}>
```
tochwill, line 149: I think this is good to merge into the alpha as long as we raise tickets to:

- Create the timestamp correctly, using the same timestamp that is used in the start time indicator (will be easier to do when they're both merged in)
- Get requirements for duration formatting and refactor that in here based on what they are
- Make sure the visually hidden live label has the same logic in here, story promos, bulletins (and anywhere else it will be used). We can merge it in as is to the alpha and then pull it out and share it across each?

EinsteinNjoroge: https://github.com/bbc/psammead/issues/3036
https://github.com/bbc/psammead/issues/3037

PR #2981, opened 2020-01-20
Add aria label to item

thekp:
Resolves n/a

**Overall change:** _Adds an aria label that includes the rank and title of the item_
image

**Code changes:**

- _export serviceNumerals_
- _add aria label to mostReadItemWrapper_
- _pass listIndex, service, and title prop to ItemWrapper so that we can set the rank as a prefix the title in the aria label_

---


Bopchy:
Needs a CHANGELOG.md, package-lock.json and package.json update

packages/components/psammead-most-read/src/Item/__snapshots__/index.test.jsx.snap
```diff
@@ -311,6 +311,7 @@ exports[`MostReadItemWrapper should render ltr correctly with 10 items 1`] = `
 
 <div>
   <li
+    aria-labelledby="1 Could a computer ever create better art than a human?"
```
tochwill, line 314: Hey! I'm not sure this is the correct usage of the `aria-labelledby` tag - that tag should be used to associate an element with another element that acts as its label, using the ID of the labelling element. What problem are you solving here by using this tag?

tochwill: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

thekp:
we decided this PR was not necessary.

Issue #2926, opened 2020-01-13
Radio Schedules: Clock icon and timestamp styling


tochwill:
**Is your feature request related to a problem? Please describe.**
Off the back of https://github.com/bbc/psammead/issues/2925, export a component from Psammead with the clock icon, timestamp and horizontal line as seen in the [designs](https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/5dc0724999d4a3572b675557), which can be composed inside a wrapping container alongside the radio schedule card component.

**Testing notes**
Expected according to the zeplin designs.

Dev insight: Will there be any potential regression? etc

- [ ] This feature is expected to need manual testing.

**Additional context**
See designs here: https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/5dc0724999d4a3572b675557

tochwill:
Dependent on #2925

EinsteinNjoroge:
Closed in https://github.com/bbc/psammead/pull/2988

PR #2988, opened 2020-01-22
Display start time of programme with clock and timestamp

### LINKED ISSUE #2925: Radio Schedules: Display start time of programme
Use Psammead’s TimestampContainer to display the start time of the programme, which will be provided as a 13 character UTC unix timestamp.

We will have a component for the radio schedule data that is being created in https://github.com/bbc/psammead/issues/2924. Above this, we will have a separate export, containing the start time of the programme. These will then be composed together inside a wrapper component. We can use the timestamp container as is to begin with, and have a separate issue to update the styling to reflect the designs. 

**Testing notes**
Verify the timestamp is local to the corresponding service and doesn't include any styling changes.

**Additional context**
See designs here: https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/5dc0724999d4a3572b675557

OlgaLyubin:
Resolves #2925 an https://github.com/bbc/psammead/issues/2926

**Overall change:**
Imported and used `TimestampContainer` to get a start time of the program as a 13 character UTC unix timestamp and display it. Imported and styled clock icon and added a horizontal line. Made StartTime a separate export within the radio schedule package.

**Code changes:**
- Use a `TimestampContainer` to display start time of the radio program.
- Style clock icon and and horizontal line
- Add snapshots

---


packages/components/psammead-radio-schedule/startTime/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
mmcculloch, line 1: Is it normal to deal with sub-components this way or should it be placed as a top level component? E.g. `@bbc/psammead-radio-schedule-starttime`

OlgaLyubin: This is the way we have done it previously e.g. in navigation https://github.com/bbc/psammead/tree/latest/packages/components/psammead-navigation/scrollable so that is why I follow this approach

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+import styled from 'styled-components';
+import { number, string, shape, oneOf } from 'prop-types';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+import { C_RHINO, C_PEBBLE } from '@bbc/psammead-styles/colours';
+import { getMinion } from '@bbc/gel-foundations/typography';
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+import { coreIcons } from '@bbc/psammead-assets/svgs';
+import TimestampContainer from '@bbc/psammead-timestamp-container';
+
+const StyledStartTime = styled.div`
```
EinsteinNjoroge: Since this is just a styled div, I think we should just call it a `StyledWrapper`

packages/components/psammead-radio-schedule/startTime/package.json
mmcculloch: Ah cool, I was hunting around and couldn't find other examples. That's fine if it's the standard approach.

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+  display: flex;
+  align-items: center;
+`;
+
+const StyledClock = styled.span`
+  display: flex;
+  align-items: center;
+  > svg {
```
EinsteinNjoroge, line 24: Looking at the designs, the clock shouldn't have a margin to the right, we should reset the default margin on the `svg` and apply some padding to the StyledClock.

OlgaLyubin: Good point, done

EinsteinNjoroge: Sorry @OlgaLyubin , looking at the designs I can't seem to find where `const CLOCK_SIDE_PADDING = '0.51875rem'` is coming from

OlgaLyubin: Screenshot 2020-01-24 at 10 43 57

This 8.3px as far as I understand correctly

OlgaLyubin: That value also seemed a bit strange to me, I will double check with UX that it's supposed to be 8.3px not 8px

OlgaLyubin: Okay so Laura said that 8.3 was a Zeplin glitch so I changed the padding value to 8px

j-pendlebury (approved):
LGTM 👍

packages/components/psammead-radio-schedule/src/index.stories.jsx
```diff
@@ -59,3 +63,38 @@ stories.add(
...
+          locale={locale}
+          script={script}
+          service={service}
+        />
+      );
+    },
+    { notes },
+  )
+  .add(
+    'Start time',
```
hotinglok:
```suggestion
    'default',
```
I think this should be the first story displayed.

packages/components/psammead-radio-schedule/src/index.stories.jsx
```diff
@@ -59,3 +63,38 @@ stories.add(
     ),
   { notes },
 );
+
+storiesOf('Components|RadioSchedule/StartTime', module)
+  .addDecorator(withKnobs)
+  .addDecorator(withServicesKnob())
+  .add(
+    'Timestamp',
```
hotinglok:
```suggestion
    'start time text',
```
Not sure if this story is needed since it isn't much different from the actual Timestamp story. If this is to be kept, I think this should be called something else since it is different as it is styled for this component. Not sure on the name though!

hotinglok:
LGTM overall, just some small suggestions in the stories.

DenisHdz (changes requested):
You will also have to bump the `Psammead` version since you are updating its dependencies.

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+export const StartTimestamp = ({
+  timestamp,
+  timezone,
+  locale,
+  script,
+  service,
+}) => {
+  return (
+    <StyledTimestamp script={script} service={service} aria-hidden="true">
+      <TimestampContainer
```
tochwill, line 79: I've noticed that the programme card is expecting the start time as a string, as it needs to read out the start time for screenreaders. Maybe what we should do is create the TimestampContainer outside of the components and pass it in to both the start time indicator and the programme card, so they both get the same timestamp to work with. I think we should merge this in as it is first though and then pull out the common parts and consider the best way of adding their styling as a separate PR.

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+
+ClockIcon.propTypes = {
+  dir: oneOf(['ltr', 'rtl']),
+};
+
+ClockIcon.defaultProps = {
+  dir: 'ltr',
+};
+
+const HorizontalLine = styled.div`
```
tochwill: Would this work as an `::after` pseudoelement on the StyledTimestamp component?

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+import { number, string, shape, oneOf } from 'prop-types';
+import { GEL_SPACING_HLF, GEL_SPACING } from '@bbc/gel-foundations/spacings';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+import { C_RHINO, C_PEBBLE } from '@bbc/psammead-styles/colours';
+import { getMinion } from '@bbc/gel-foundations/typography';
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+import { coreIcons } from '@bbc/psammead-assets/svgs';
+import TimestampContainer from '@bbc/psammead-timestamp-container';
+
+const StyledWrapper = styled.div`
```
tochwill: I think just `Wrapper` is ok!

packages/components/psammead-radio-schedule/src/index.stories.jsx
OlgaLyubin: Good point, that story was redundant so I removed it and only left the Start time (default) one.

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
OlgaLyubin: Good point, after this PR and https://github.com/bbc/psammead/pull/2938 are merged we should address this issue

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
OlgaLyubin: T̶r̶i̶e̶d̶ ̶t̶h̶a̶t̶ ̶a̶n̶d̶ ̶i̶t̶ ̶d̶o̶e̶s̶ ̶n̶o̶t̶ ̶w̶o̶r̶k̶ ̶p̶r̶o̶p̶e̶r̶l̶y̶ ̶i̶f̶ ̶a̶d̶d̶e̶d̶ ̶a̶s̶ ̶a̶ ̶p̶s̶e̶u̶d̶o̶e̶l̶e̶m̶e̶n̶t̶  This is now updated and works as a pseudoelement.

packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,269 @@
...
+  -webkit-align-items: center;
+  -webkit-box-align: center;
+  -ms-flex-align: center;
+  align-items: center;
+  padding-right: 0.25rem;
+  padding-left: 0.5rem;
+}
+
+.c1 > svg {
+  color: #5A5A5A;
```
greenc05, line 48: Is this going to need fill: currentColor; as well??

packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,269 @@
...
+    class="c3"
+  >
+    <time
+      class="c4"
+      datetime="2019-08-27"
+    >
+      14:54
+    </time>
+  </span>
+  <div
```
greenc05: what's this div for??

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
greenc05: Will that also help with making the whole card including the start time clickable?

packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap
OlgaLyubin: Technically svg should already have `fill: currentColor` (https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-assets/src/svgs/coreIcons.jsx#L24) . But I've noticed that in some cases e.g. in Bulletin we add that value again when we are using the icon. So I can add this property again if needed?

packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap
OlgaLyubin: This is for the horizontal line after the timestamp

packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap
greenc05: I can't say if it's needed without testing it...

packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap
greenc05: And is that valid html?  
```
<div class="" dir="" />
```

OlgaLyubin: I have tried using an `<hr />` element in order to avoid having an empty div with a border, but `<hr />` did not work well with FireFox high contrast mode. I will look into other ways how this can be implemented

packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap
OlgaLyubin: I checked the icon in developer tools and the svg has property `fill: currentColor`

Screenshot 2020-01-28 at 11 28 03

It is also in the snapshot file here: https://github.com/bbc/psammead/blob/dbd13f5a5582ec41740e0307d933c5818054dab4/packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap#L8

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
OlgaLyubin: @greenc05 Yes, it will. We will make the start time clickable in our next PR

packages/components/psammead-radio-schedule/src/StartTime/__snapshots__/index.test.jsx.snap
OlgaLyubin: This is now updated and the line is added to an `::after` pseudoelement instead of an additional `div`.

packages/components/psammead-radio-schedule/src/StartTime/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+    ${({ service }) => service && getSansRegular(service)}
+  }
+
+  &::after {
+    content: '';
+    border-top: 0.0625rem solid ${C_PEBBLE};
+    top: ${({ script }) => 0.5 + script.minion.groupA.lineHeight / 2 / 16}rem;
+    ${({ dir }) =>
+      dir === 'ltr' ? `margin-left: 0.625rem;` : `margin-right: 0.625rem;`}
+    width: 100vw;
```
AlistairGempf: image
This is causing the screen to be horizontally scrollable (note the scrollbar on the bottom) - the `:after` pseudo-element is the full width of the screen, then the clock and time are also added on top of that full width. This should be possible to remedy with flex but I don't know the css off the top of my head.

AlistairGempf: From a quick investigation, putting `width: 100%` on the `StyledTimestamp` and changing the `width: 100vw` to `width: 100%` on the `:after` should solve this.

OlgaLyubin: Good spot, fixed now 👍

AlistairGempf (approved):
Looks good.

mmcculloch (approved):

DenisHdz (approved):
Looks good 👍

PR #2991, opened 2020-01-22
Revert 2836 - Add AV embed timeout handling to MP

simonsinclair:
**Overall change:** This is a manual revert of https://github.com/bbc/psammead/pull/2836, which added AV embed timeout handling to Media Player. We had to revert this due to a strange interaction with Simorgh's e2e tests, which could be related to a bug in Cypress. See below for additional context. Until we can offer this issue more time, this functionality will be reverted.

**Code changes:**

- Revert changes made in https://github.com/bbc/psammead/pull/2836.
- Bump package to 2.6.0.
- Add changelog entry.

**Additional context:**

- Attaching an on `load` event handler to Canonical MP iframe causes the first e2e test that accesses the iframe's `contentWindow` property to fail. Subsequent tests will pass. It does not matter on the order of these tests - the first one will always fail.
- A possible related open issue with Cypress: https://github.com/cypress-io/cypress/issues/4973.
- When we can afford more time to this problem, it might be worth replicating this issue in a slim repository that mimics our setup and use it to raise a bug report with Cypress.

---


j-pendlebury (approved):
LGTM 👍

rhenshaw56 (approved):

PR #2181, opened 2019-09-19
Simplifies stories

samora:
Resolves #2157 

**Overall change:** Simplifies stories. Remove dependencies which are not important to demonstrating usage.

**Code changes:**

- Remove story promo's dependency on `Timestamp`. 

---

- [x] I have assigned myself to this PR and the corresponding issues
- [x] Automated jest tests added (for new features) or updated (for existing features)
- [x] This PR requires manual testing

Issue #2157, opened 2019-09-17
Stop building complex stories in psammead packages


dr3:
**Is your feature request related to a problem? Please describe.**
In psammead we have a number of stories, like [story promo](https://github.com/bbc/psammead/blob/latest/packages/components/psammead-story-promo/src/index.stories.jsx) where we setup extremely complicated story designs.

This is just mimicking the layout in simorgh, IMO we should keep stories for psammead as simple as possible, just highlighting the bare bones components. Adding `VisuallyHiddenText`, `MediaIndicator` & `Timestamp` to the story of `Story Promo` gives a false representation of what the `story promo` component provides. 

**Describe the solution you'd like**
Slim down our existing stories, isolating them to only contain what they export. We should discuss if we should rely on Simorgh or a seperate collection of stories, to give a representation of what you _can_ build with psammead.

**Testing notes**
[Tester to complete]

Dev insight: Will Cypress tests be required or are unit tests sufficient? Will there be any potential regression? etc

- [ ] This feature is expected to need manual testing.

**Additional context**

ghost:
I think it is important that we decide on when we do accessibility checks for a component. If a component has a requirement for `VisuallyHiddenText` it makes sense to include that in the story in order to facilitate isolated accessibility checks on the component.

DenisHdz:
I am happy simplifying them, as long as all the disciplines know that when something is missing in Psammead Storybook, they should check the Simorgh one to see all the functionality. 

Designers and a11y leads like to see them in Psammead during the development of a new component and also in conjunction with other ones, as happened with the navigation and the brand, which ends up adding all the requirements of a component in its story.

sareh:
I think it's a visibility issue. Having a `storypromo - default` story that is *just* that promo, without media indicator etc, is very important. 

This component has props for four nodes - two required, `image` & `info`, two optional, `mediaIndicator` & `topStory`

The `storypromo - default` should include the simplest possible required nodes for `image` and `info` - a single image and a single text component - for example a heading or paragraph (`psammead-headings` or `psammead-paragraph`)

Then we should have clear additional examples that show how it'd look like with the extra nodes passed in:

```
storypromo - standard
storypromo - standard - with media indicatior
storypromo - top story
storypromo - top story - with media indicator
```

The examples should be the simplest possible. So for the `info` node, we shouldn't have a heading and paragraph and timestamp. 

[...]

PR #3027, opened 2020-01-29
Leading story promo stylings on mobile and tablet

### LINKED ISSUE #3024: Use agreed LeadingStory layouts
Following discussion with UX, we have updated the layouts for `StoryPromo`s of type `'leading'`.

On all breakpoints, the DOM ordering will be `Info` component (containing Headline, Summary, Timestamp) followed by the `Image` component. Internally this means the `TextGridItem` followed by `ImageGridItem`.

| Breakpoint | `TextGridItem` columns | `ImageGridItem` columns | Total columns/line |
|------------|-------------------------|---------------------------|-------------------|
| > 1007px | 2 | 4 | 6 |
| 600px - 1007px | 3 | 3 | 6 |
| < 600px | 6 | 6 | 6 |

This will also need the relevant fallbacks added.

**Testing notes**
[Tester to complete]

**Additional context**
Tablet: 
image

Mobile (note the padding under the timestamp is not correct, this was a quick prototype):
image

AlistairGempf:
Resolves #3024

**Overall change:** Puts stylings on the leading story promo that are correct for mobile and tablet

**Code changes:**

- Add required columns for mobile and tablet on leading story promo
- Add necessary fallbacks for the above

---


sareh (approved):
👍 Looks good at all breakpoints.

OlgaLyubin (commented):
Looks great on all breakpoints and on IE too. The only thing is that in storybook when you choose e.g. Arabic service, correct language for headline is displayed but it doesn't work the same for timestamps. It just replaces "2 January 1970" with "January 1970 2"

packages/components/psammead-story-promo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-story-promo",
-  "version": "4.0.0-alpha.5",
+  "version": "4.0.0-alpha.6",
```
DenisHdz: I think `@bbc/psammead-timestamp-container` should be added as a `devDependency` since we are using it now in the stories and it is not in the root `package.json`.

packages/components/psammead-story-promo/src/TextStyles.jsx
```diff
@@ -4,6 +4,7 @@ import {
   GEL_GROUP_3_SCREEN_WIDTH_MIN,
   GEL_GROUP_4_SCREEN_WIDTH_MIN,
   GEL_GROUP_5_SCREEN_WIDTH_MIN,
+  GEL_GROUP_3_SCREEN_WIDTH_MAX,
```
DenisHdz: Can we put this in order?

packages/components/psammead-story-promo/src/TextStyles.jsx
```diff
@@ -104,6 +129,9 @@ const TextGridItem = styled.div`
       padding: initial;
       ${({ promoType }) => textGridStyles[promoType]}
     }
+
+    ${({ promoType }) =>
```
DenisHdz, line 136: Shall we add this condition in the `ImageGridItem` and apply a `padding-top` to the image instead? I remember having [this](https://github.com/bbc/psammead/issues/2157) conversation in the past about trying to simplify the stories and the [outcome](https://github.com/bbc/psammead/pull/2181/files) of it was to remove the `TimeStamp` from the story.

AlistairGempf: The headline already has padding-bottom applied to it, so I targeted the timestamp specifically to add padding to it to prevent double padding being applied when there is no timestamp. Initially I wanted to use the `:last-child` selector to apply it ensuring there's always 8px of padding between the text and image, but I couldn't get that to work.

In terms of the story, good point. I'd forgotten that conversation had happened. I can absolutely see why building complex stories is not wanted, in this case I thought it would be the easiest way to check I was adding it properly. I'll remove it now.

DenisHdz: Since we won't have a story showing how the padding-bottom applied to the `time` looks like, do you think it would be useful to add a short comment above the `leadingPromoTimestampPadding` or is it auto explanatory?

packages/components/psammead-story-promo/package.json
DenisHdz: We don't need this now :)

packages/components/psammead-story-promo/src/TextStyles.jsx
AlistairGempf: A comment is a good idea - it's not clear even with the story why it's applied only to the `timestamp`.

package.json
```diff
@@ -66,6 +66,7 @@
     "@bbc/psammead-styles": "^4.2.0",
     "@bbc/psammead-test-helpers": "^3.1.2",
     "@bbc/psammead-timestamp": "^2.2.23",
+    "@bbc/psammead-timestamp-container": "^2.6.20",
```
DenisHdz: We don't need this anymore

packages/components/psammead-story-promo/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
+| 4.0.0-alpha.7 | [PR#3027](https://github.com/bbc/psammead/pull/3027) Use mobile and tablet styings |
```
AlistairGempf:
```suggestion
| 4.0.0-alpha.7 | [PR#3027](https://github.com/bbc/psammead/pull/3027) Use mobile and tablet stylings |
```

DenisHdz (approved):
Great work 👍

packages/components/psammead-story-promo/src/index.stories.jsx
```diff
@@ -125,6 +125,7 @@ const generateStory = ({
       type={mediaType}
       alsoItems={alsoItems}
       promoHasImage={displayImage}
+      locale={locale}
```
OlgaLyubin: I don't think you need `locale` here anymore since timestamp got deleted

PriyaKR:
The grid layout changes looks good to me couldn't verify the padding-bottom below the timestamp on mobile as the timestamp is not available for this simple leading story.

PR #3029, opened 2020-01-29
Inline Media Indicator

### LINKED ISSUE #3019: Update Media Indicator component
Currently the 'index also' version of the media indicator is fragmented and not easily reusable. 

We suggest having a single export within the `psammead-media-indicator` package that outputs the following example markup, which can be used as a direct child of a link or span, as appropriate.

```
<span aria-hidden="true">
  <svg ... focusable="false" aria-hidden="true" >
    ...
  </svg>
</span>
<span role="text">
  <VisuallyHiddenText>video, </span>
  <span>{children}<span>
</span>
```
Note - the first span should have padding-left/padding-right (according to dir) of `GEL_SPACING` so the icon is not too close to the text. 

This is needed for the Desktop layout work. After completed, the Index Alsos within layouts can be updated to use this InlineMediaIndicator variant, since it includes the visually hidden text required for accessibility.

Have an export called `InlineMediaIndicator`. 
Usage

```
import { InlineMediaIndicator } from '@bbc/psammead-media-indicator';
```

By not making the change here, we will be duplicating work in several places in Simorgh.

**Testing notes**
[Tester to complete]

Dev insight: Readmes need updating, unit tests/snapshots should be added / storybook examples should be added.

**Additional context**

image

j-pendlebury:
Resolves https://github.com/bbc/psammead/issues/3019

**Overall change:** 
Add prop to MediaIndicator to be able to use this component inline with text.

**Code changes:**

- Add `isInline` prop to MediaIndicator
- Update stories for each type - video, audio, photogallery
- Add knob to stories for inline MediaIndicator to toggle inline state, and to add extra text if required
- Add tests for inline media indicator for LTR and RTL service (only for video since the behaviour is the same for other types)

#### Index Also with `isInline=true` on News
image

#### Index Also with `isInline=true` on Persian
image

#### IE11
image
image

(with Persian, the behaviour is different on IE11 than Chrome, but I checked on `latest` and it's the exact same behaviour)

---


jblaut:
image
image

Chrome looks fine, IE however needs fixing.

sareh:
@jblaut That IE11 issue is to do with the text itself not being in a RTL script. When I replaced it with arabic text, it displayed correctly:
Screenshot 2020-01-30 at 15 25 04

I will open an issue to fix the Storybook test for StoryPromo with index alsos, using the Storybook helper to use the correct localised text for the index alsos. It'd make testing index alsos reflective of reality.

packages/components/psammead-media-indicator/src/index.jsx
```diff
@@ -42,54 +49,44 @@ const TimeDuration = styled.time`
   margin: 0 ${GEL_SPACING_HLF};
 `;
 
-const IndexAlsosMediaIndicator = styled.span`
```
DenisHdz, line 45: We will have to also update the `IndexAlsos` straight after to use the `MediaIndicatorWrapper` with the `isInline` prop instead since this `IndexAlsosMediaIndicator` won't exist anymore.

packages/components/psammead-media-indicator/src/index.jsx
```diff
@@ -42,54 +49,44 @@ const TimeDuration = styled.time`
...
+);
 
 MediaIndicator.propTypes = {
   datetime: string,
   duration: string,
   type: oneOf(['video', 'audio', 'photogallery']),
   topStory: bool,
   service: string.isRequired,
-  indexAlsos: bool,
+  isInline: bool,
```
ryanmccombe, line 83: We should include a small update to the README if we're adding or removing props 😃

packages/components/psammead-media-indicator/src/index.jsx
j-pendlebury: Created [Psammead](https://github.com/bbc/psammead/issues/3039) and [Simorgh](https://github.com/bbc/simorgh/issues/5340) issues 👍

packages/components/psammead-media-indicator/src/index.jsx
j-pendlebury: All done 👍

Bopchy (approved):

packages/components/psammead-media-indicator/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
+| 3.0.0 | [PR#3029](https://github.com/bbc/psammead/pull/3029) Add prop for displaying media indicator inline |
```
sareh:
```suggestion
| 3.0.0 | [PR#3029](https://github.com/bbc/psammead/pull/3029) Add prop `isInline` for displaying media indicator inline. Remove boolean prop `indexAlsos` since it should be replaced with `isInline` |
```

packages/components/psammead-media-indicator/src/index.jsx
```diff
@@ -42,54 +49,44 @@ const TimeDuration = styled.time`
...
-      <FlexWrapper>
-        {mediaIcons[type]}
-        {duration && datetime && (
-          <TimeDuration dateTime={datetime}>{duration}</TimeDuration>
-        )}
-      </FlexWrapper>
-    </MediaIndicatorWrapper>
-  );
+      {duration && datetime && (
+        <TimeDuration dateTime={datetime}>{duration}</TimeDuration>
```
sareh, line 70: Just had a UX review of the Desktop layout, and the UX designers said that when we use the media indicator inline, the duration shouldn't be displayed.  
There are two ways to do this. One is to couple that logic here in Psammead - so when you have `isInline`, you don't render the duration component.  The other way is to handle this where this component is being used.  I think this depends on whether we will *ever* have the duration being displayed when in the 'inline' view. I'll just double check with UX about this.

j-pendlebury: I'm happy to include this in this PR, but will wait to make sure

OlgaLyubin (commented):
There is one test that still has an `indexAlsos` prop instead of `isInline`

https://github.com/bbc/psammead/blob/009af95bddd8158f12d9a195cf78070136bcec9b/packages/components/psammead-media-indicator/src/index.test.jsx#L13

packages/components/psammead-media-indicator/src/index.jsx
```diff
@@ -42,54 +49,44 @@ const TimeDuration = styled.time`
...
+      )}
+    </FlexWrapper>
```
sareh, line 73: Can we pass in `children` into the flex wrapper just after the duration? This will enable us to have text that is aligned with the Media Indicator icon, without having to add more divs/spans in our markup.

packages/components/psammead-media-indicator/src/index.jsx
sareh: @jblaut Thanks. Confirmed with UX that for now, it can be coupled to the `isInline` prop since that is the general usecase.

OlgaLyubin (commented):
This is not a mistake at all, but maybe we should make "example text" lie within the white block? Just for appearance

Screenshot 2020-02-03 at 11 12 01

j-pendlebury:
> This is not a mistake at all, but maybe we should make "example text" lie within the white block? Just for appearance
> 
> Screenshot 2020-02-03 at 11 12 01

@OlgaLyubin yeah I thought about it, but the `example text` is only there to show how the component looks with text inline and is not part of it. I like the separation between the two

packages/components/psammead-media-indicator/src/index.stories.jsx
```diff
@@ -52,6 +52,20 @@ storiesOf('Components|MediaIndicator/Video', module)
...
+  .add(
+    'inline media indicator',
+    ({ service }) => (
+      <>
+        <MediaIndicator
+          type="video"
+          service={service}
+          isInline={boolean('inline?', true)}
+        />
+        {text('extra text', 'example text')}
```
sareh: Now that we have `children` as a prop, we should pass in the example text as a child node. 
e.g. `<MediaIndicator type="video"....>{text('extra text', 'example text')}</MediaIndicator>`

OlgaLyubin (commented):
Older stories ("audio without duration", "video without duration" and "photogallery") have `indexAlsos` prop. Other than that everything looks great

j-pendlebury:
> Older stories ("audio without duration", "video without duration" and "photogallery") have `indexAlsos` prop. Other than that everything looks great

all fixed now 👍

OlgaLyubin (approved):
Looks good 👍

PriyaKR:
Looks good to me.

PR #3061, opened 2020-02-04
Create Byline component

### LINKED ISSUE #3056: Create new Byline component
CPS story assets have a byline displaying the author and their title/role. We need to create a Byline component that will render the name and the title of a given author.

There are no dedicated UX designs for the byline, we should work off of what we have on our existing pages. A byline can have an optional image, so the component must render with an image is one is provided (and this should have an empty alt attribute as that would be the same as the authors name and we don't need to repeat that for users).

Byline without an image:
image

Byline with an image:
image

**Testing notes**
The optional avatar does not need to be factored into testing this component, yet.

**Additional context**

simonsinclair:
Resolves https://github.com/bbc/psammead/issues/3056.

### Overall change
Create `@bbc/psammead-byline`, which displays an author's full name, role, and optional avatar.

### Code changes

- Add README.md
- Add CHANGELOG.md
- Add index.jsx
- Add index.test.jsx
  - And relevant snapshots.
- Add index.stories.jsx

### API

| Argument          | Type      | Required | Default   | Example                        |
| ----------------- | --------- | -------- | --------- | ------------------------------ |
| service           | string    | Yes      | N/A       | `'news'`                       |
| avatar | object    | No       | `null`    |  _See below._                  |
| name              | string    | Yes      | N/A       | `'John Smith'`                 |
| title             | string    | Yes      | N/A       | `'Art editor'`                 |

#### Avatar
| Argument          | Type      | Required | Default   | Example                          |
| ----------------- | --------- | -------- | --------- | -------------------------------- |
| src               | string    | Yes      | N/A       | `'http://www.bbc.co.uk/img.jpg'` |
| alt               | string    | No       | `null`    | `'A person wearing glasses.'`    |

---


packages/components/psammead-byline/src/index.jsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import { string, shape } from 'prop-types';
+import styled from 'styled-components';
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+import { GEL_SPACING_DBL } from '@bbc/gel-foundations/spacings';
+
+const AVATAR_DIAMETER = '64px';
```
FK78: I believe we use REMs instead of pixels in Psammead, this would be 4rem.

simonsinclair: Thanks, Fahad. This was a draft value used to demo the component in refinement today. I will change it before I put it into review. 😄

andrewscfc (approved):

FK78 (approved):

packages/components/psammead-byline/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,137 @@
...
+  .c2 {
+    font-size: 0.875rem;
+  }
+}
+
+<div
+  class="c0"
+>
+  <img
+    class="c1"
```
greenc05, line 57: missing alt=""??

simonsinclair: This is correct, because `alt` is unset in this test. I will add another case where it isn't.

simonsinclair: Good spot! I'll fix that.

greenc05: thanks

PR #3062, opened 2020-02-05
Media Indicator - remove spacing on the low-level component

DenisHdz:
Resolves #NUMBER

**Overall change:** 
After passing the `inLine` prop added in [this](https://github.com/bbc/psammead/pull/3029) PR to the `Media Indicator` used in the `Index Alsos`, we've realized that the `MediaIndicatorWrapper` styles are not suitable.

Before:
Screenshot 2020-02-05 at 10 00 26

After:
Screenshot 2020-02-05 at 09 59 51

For this reason, we've decided to remove the spacing on the `Media Indicator` so we can add them at a top-level in a separate PR (Story Promo/Index Alsos).

**Code changes:**
- Remove `padding` and `height` from the `MediaIndicatorWrapper`
- Remove the Time element from the Media Indicator
- Update stories
- Remove `topStory` test

---


packages/components/psammead-media-indicator/src/index.jsx
```diff
@@ -2,40 +2,24 @@ import React from 'react';
...
 import { string, oneOf, bool, node } from 'prop-types';
 import { C_WHITE, C_EBON } from '@bbc/psammead-styles/colours';
-import {
-  GEL_SPACING_HLF,
-  GEL_SPACING,
-  GEL_SPACING_QUAD,
-} from '@bbc/gel-foundations/spacings';
-import { GEL_GROUP_1_SCREEN_WIDTH_MAX } from '@bbc/gel-foundations/breakpoints';
-import { GEL_MINION } from '@bbc/gel-foundations/typography';
+import { GEL_SPACING_HLF } from '@bbc/gel-foundations/spacings';
```
sareh: Seems like we still are using this GEL_SPACING_HLF,

When looking at where it's used, it's applied to the Time element
https://github.com/bbc/psammead/blob/302a5857b799b4f2cd192a0ccfe468ba5bd86487/packages/components/psammead-media-indicator/src/index.jsx#L33

Makes me think that this should be passed in as a Child, rather than be within this component.  
We already have the `children` prop.

packages/components/psammead-media-indicator/src/index.jsx
```diff
@@ -2,40 +2,24 @@ import React from 'react';
...
+  ${({ script }) => script && getMinion(script)};
 
   ${({ isInline }) =>
     isInline &&
     css`
       display: inline-block;
       vertical-align: middle;
     `}
-
-  ${({ topStory }) =>
```
sareh, line 31: We can remove the `topStory` prop now.

packages/components/psammead-media-indicator/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
+| 4.0.0 | [PR#3062](https://github.com/bbc/psammead/pull/3062) Remove spacing on the low-level component |
```
sareh: 👍 Making it as a major version is good. We should update the Readme - the `topStory` prop being removed, that we expect users of this component to add their own spacing where relevant. 
That children passed include a Timestamp passed in. We shouldn't need the DateTime component defined here, with the specific spacing, when spacing may change depending on context.  https://github.com/bbc/psammead/blob/f8537de4a42cedfee4f1135c965946303dc3c88f/packages/components/psammead-media-indicator/src/index.jsx#L70

packages/components/psammead-media-indicator/src/index.jsx
sareh: A consequence of the change would be that other components that use `psammead-media-indicator` would need to be updated. 

I’ve looked at Psammead and Simorgh to see how much work this would be.
Psammead:
Stories and tests/test helpers in these two components, that can be carried out in one PR
```
packages/components/psammead-grid/src/index.stories.jsx:
packages/components/psammead-story-promo/src/index.stories.jsx:
packages/components/psammead-story-promo/src/index.test.jsx:
packages/components/psammead-story-promo/testHelpers/IndexAlsosContainer.jsx:
```
(Note, the top-level dependency in psammead would need to be updated)

In Simorgh, the changes would be to these containers - being carried out as part of the Desktop redesign: https://github.com/bbc/simorgh/pull/5139
```
src/app/containers/StoryPromo/IndexAlsos/index.jsx:
src/app/containers/StoryPromo/MediaIndicator/index.jsx:
```

OlgaLyubin (commented):
There is still a `top story photogallery` story even though `topStory` styles/prop got removed

packages/components/psammead-media-indicator/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
+| 4.0.0 | [PR#3062](https://github.com/bbc/psammead/pull/3062) Pass Time element as a child and remove spacing froom the `MediaIndicatorWrapper`|
```
OlgaLyubin:
```suggestion
| 4.0.0 | [PR#3062](https://github.com/bbc/psammead/pull/3062) Pass Time element as a child and remove spacing from the `MediaIndicatorWrapper`|
```

packages/components/psammead-media-indicator/README.md
```diff
@@ -32,19 +29,14 @@ The `MediaIndicator` component provides a 'play', 'audio' or 'camera' icon as we
 
 The typical use-case of this component is on top of images within promos for articles that contains a video asset at the top of the page. It indicates to the user that the link is to a video and how long the video is in duration.
 
-For top story promos, we should pass the `topStory` prop to the `Media Indicator` to keep the same padding, otherwise this will be modified under 400px.
-
 Top story promos can also have related stories links called `Index Alsos`. When an Index Also item is a video, audio or photogallery, an icon should be displayed with the content using the `MediaIndicator` component passing the `isInline` prop.
 
+We expect users to add their own spacing around this component where relevant.
```
OlgaLyubin: Maybe it's worth rephrasing this a bit since "users" sounds like users of the website, not developers

packages/components/psammead-media-indicator/src/index.stories.jsx
```diff
@@ -26,39 +37,28 @@ storiesOf('Components|MediaIndicator/Video', module)
   .add(
     'video with duration',
     ({ service }) => (
-      <MediaIndicator
-        duration={text('duration', '2:15')}
-        datetime={text('datetime', 'PT2M15S')}
-        type="video"
-        service={service}
-      />
+      <MediaIndicator type="video" service={service} isInline>
```
sareh:
```suggestion
      <MediaIndicator type="video" service={service}>
```

OlgaLyubin (approved):
Everything looks good, `inline` toggle works properly now 👍

sareh (approved):
👍 Changes look good. 
Looking at the ChromaticQA snapshots, there is a visual difference in the height for the 'inline' version of the MediaIndicator - this is expected.

Issue #3066, opened 2020-02-05
Include Front Page/Article Page stylings in MostRead


hotinglok:
**Is your feature request related to a problem? Please describe.**
When integrating MostRead into pages, it has some issues adhering to the layout of the page because of the pre-existing grid. 
![image](https://user-images.githubusercontent.com/53564281/73841635-e792f300-4812-11ea-9410-39e1c68d0b30.png)
![image](https://user-images.githubusercontent.com/53564281/73841795-3d679b00-4813-11ea-8e9f-9271c823d862.png)

**Describe the solution you'd like**
Add the styles into Psammead Most Read and apply them here to be pulled into Simorgh so it can adhere to the pre-existing grids on the pages.

**Describe alternatives you've considered**
N/A

**Testing notes**
[Tester to complete]

Dev insight: Will there be any potential regression? etc

- [ ] This feature is expected to need manual testing.

**Additional context**

PR #3071, opened 2020-02-06
Update MostRead README

mmcculloch:
README update for  #3066

**Overall change:** Add missing README content

**Code changes:**

- README update, no code change
---

### DIFF
--- a/packages/components/psammead-most-read/README.md
+++ b/packages/components/psammead-most-read/README.md
@@ -72,24 +72,6 @@ const items = [
 />;
 ```
 
-The `className` prop allows additional styles to be applied to the components `MostReadSection` wrapper. This is intended to be passed by `styled-components`.
-
-```jsx
-import styled from 'styled-components';
-
-const StyledMostRead = styled(MostRead)`
-  margin: 0 auto;
-`;
-
-<StyledMostRead
-  items={items}
-  script={script}
-  service="arabic"
-  header="الأكثر قراءة"
-  dir="rtl"
-/>;
-```
-
 ## MostReadList
 
 ### Props
--- a/packages/components/psammead-most-read/README.md
+++ b/packages/components/psammead-most-read/README.md
@@ -10,7 +10,7 @@ The `MostRead` component is designed to display the most read articles given a d
 
 `npm install @bbc/psammead-most-read`
 
-## Components (WIP)
+## Components
 
 ## MostRead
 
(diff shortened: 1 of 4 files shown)


packages/components/psammead-most-read/README.md
```diff
@@ -71,6 +72,24 @@ const items = [
 />;
 ```
 
+The `className` prop allows additional styles to be applied to the components `MostReadSection` wrapper. This is intended to be passed by `styled-components`.
+
+```jsx
```
DenisHdz: Sorry for the misunderstanding, but I meant to give me an example of how the `className` is passed and used, not to add it to the README. 

Still, I can not understand how the `className` is used in this example. This just shows how to apply styles to the `MostRead` using styled-components. Is that what you are trying to achieve?

mmcculloch: Hmm yes the intention is that it's never used directly, styled-components will pass a className in this case. There's nothing stopping us from passing a className manually though.

mmcculloch: Example of the PlayButton doing this as well

https://github.com/bbc/psammead/blob/latest/packages/components/psammead-play-button/src/index.jsx#L73

mmcculloch: I've removed the example since it's unnecessary, but as it was showing, when `styled` is used on a component
```jsx
const StyledMostRead = styled(MostRead)`
  background-color: white;
`;
```
A className is generated and then passed in to it, in this case `MostRead`.

mmcculloch: Example from docs: https://styled-components.com/docs/basics#styling-any-component

DenisHdz: Thanks for the explanations @mmcculloch. I'm still not sure if we need to pass a `className` to be able to style the component from somewhere else. 

There are few places in Simorgh and Psammead where we just import the component and wrap it with `styled` when we want to expand/override its styles and from what I'm aware they work fine without the `className`.

E.g. 
https://github.com/bbc/simorgh/blob/latest/src/app/containers/Headings/index.jsx#L16
https://github.com/bbc/simorgh/blob/latest/src/app/containers/RadioPageMain/index.jsx#L17

DenisHdz: I've done some testing and now I understand the functionality of the `className` https://codesandbox.io/s/gracious-rhodes-lilfy?fontsize=14&hidenavigation=1&theme=dark

Basically, it is needed when you want to style any component which **has not** been defined using styled-components, and probably those components which we extend their styles in our codebase at the moment, have been defined using `styled-components`, therefore my confusion.

thekp (approved):

packages/components/psammead-most-read/README.md
mmcculloch: Yeah that makes sense. The others are defined by styled-components so automatically have this functionality. Especially Headline which is a fairly basic styled element.

DenisHdz (approved):
LGTM 👍

PR #3075, opened 2020-02-06
Media Indicator - Pass script

DenisHdz:
Resolves #NUMBER

**Overall change:** 
Pass script to Media Indicator to apply the right typography.

**Code changes:**
- Pass `script` prop to the `MediaIndicator` component
- Update stories
- Update tests

---

### DIFF
--- a/packages/components/psammead-media-indicator/CHANGELOG.md
+++ b/packages/components/psammead-media-indicator/CHANGELOG.md
@@ -3,7 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 | ------- | ----------- |
-| 4.0.1 | [PR#XXX](https://github.com/bbc/psammead/pull/XXX) Pass `script` to the `MediaIndicator`|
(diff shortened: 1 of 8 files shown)


sareh (approved):
👍 Thanks for this fix. When looking locally in Storybook, I see the script-specific font-size and line-height applied.

e.g. http://localhost:8180/?path=/story/components-mediaindicator-video--video-with-duration
Screenshot Media indicator with duration text

thekp (commented):
https://github.com/bbc/psammead/pull/3075/files#diff-9caff7f606acc564ecc5cea4ea7173e6R37-R41

in this example ^ should we add an example of importing scripts and passing it as a prop, since it is a required prop?

DenisHdz:
> https://github.com/bbc/psammead/pull/3075/files#diff-9caff7f606acc564ecc5cea4ea7173e6R37-R41
> 
> in this example ^ should we add an example of importing scripts and passing it as a prop, since it is a required prop?

Done :)

thekp (approved):

PR #3074, opened 2020-02-06
Format hidden start time in program-card for Radio Schedules

OlgaLyubin:
Resolves #3036

**Overall change:** 
Formatted the visually hidden start time inside the program card component.
Hidden start time can be seen by checking out developer tools:
Screenshot 2020-02-06 at 12 22 35

For RTL version:
Screenshot 2020-02-10 at 12 28 52

It is also announced by the screen reader correctly:
Screenshot 2020-02-06 at 12 23 45

**Code changes:**
- Used `formatUnixTimestamp` for transforming the timestamp
- Passed `timezone` and `locale` props to program card

---

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Automated jest tests added (for new features) or updated (for existing features)
- [x] This PR requires manual testing

PR #3078, opened 2020-02-06
Export timestamp-container utilities

OlgaLyubin:
Resolves #NUMBER

**Overall change:** 
Export `timestampUtilities` from `@bbc/timestamp-container` so that `formatUnixTimestamp ` can be used for Radio Schedules in https://github.com/bbc/psammead/pull/3074.

**Code changes:**
- Exported `timestampUtilities` by creating a `package.json`
- Added export to README

---


DenisHdz (approved):
LGTM, just wondering if we should create a `utilities` folder inside the `psammead-timestamp-container` which includes the `timestampUtilities.js` and `timestampUtilities.test.js` to keep the files more organized.

OlgaLyubin:
> LGTM, just wondering if we should create a `utilities` folder inside the `psammead-timestamp-container` which includes the `timestampUtilities.js` and `timestampUtilities.test.js` to keep the files more organized.

Good point, that would be a cleaner structure. Will update the PR

packages/containers/psammead-timestamp-container/src/index.jsx
```diff
@@ -6,7 +6,7 @@ import {
   isValidDateTime,
   unixTimestampToMoment,
   formatUnixTimestamp,
-} from './timestampUtilities';
+} from './Utilities/timestampUtilities';
```
DenisHdz: Usually, the utility folders are lowercase, do you mind making this change?

OlgaLyubin: I have updated the imports and renamed the folder, but for some reason GitHub doesn't register the change from `Utilities` to `utilities` and it did not update the name even though locally everything is correct. Currently looking into it.

EinsteinNjoroge (approved):
LGTM

packages/containers/psammead-timestamp-container/src/index.jsx
OlgaLyubin: This is done now :) Git doesn't like case sensitivity

DenisHdz (approved):
LGTM

PR #3102, opened 2020-02-10
Fix double digit column spacing in MostRead

### LINKED ISSUE #3101: MostRead Items have strange spacing for most services
**Describe the bug**
For most services, it seems as if recent changes have caused some strange spacing for certain items in most services. This mostly has to do with the fifth item in a list of ten which has specific spacing in the largest breakpoint to accommodate the double digit rank of the last item.

**To Reproduce**

1. Go to storybook online or on localhost.
2. Go to MostRead and see off spacing on news and many other services.

**Expected behaviour**
Consistent spacing for all items in MostRead.

**Screenshots**
image

hotinglok:
Resolves #3101 & #3103

**Overall change:** 
Changed values in `doubleDigitOverride` and refactored `ItemWrapper` in order to fix odd spacing for ~most~ all services.
~Some services now have more problems and will most likely be dealt with in a separate PR.~
~Services which have more complex spacing issues are mentioned here: https://github.com/bbc/psammead/issues/3103~
~Will be worked on in another PR.~

**Code changes:**
- Updated `doubleDigitOverride` with a new default which covers 14 services.
- Updated values in order to fix spacing.
- Refactored `ItemWrapper` to include a non-auto default.
- Updated snapshots and comments.

---


packages/components/psammead-most-read/src/testHelpers/doubleDigitOverride.js
```diff
@@ -1,34 +1,34 @@
 // These default measurements are to be used for 2nd column minimum width.
+// Services which use this default:
+// afaanoromoo, afrique, amharic, gahuza, hausa, hindi, igbo, indonesia,
+// kyrgyz, serbian, somali, swahili, ukchina, zhongwen (14)
 export const doubleDigitDefault = {
   group0: '2rem',
```
DenisHdz, line 6: Could you replace some of these `rem`s with existing Psammead spacing values? https://github.com/bbc/psammead/blob/latest/packages/utilities/gel-foundations/src/spacings.js

tochwill:
A conversation to have once we've gone live with most read and can spend some time having a rethink, but I think it's worth taking a step back and seeing what we can achieve without this overrides file, I think it could become quite hard to maintain going forwards!

packages/components/psammead-most-read/src/testHelpers/doubleDigitOverride.js
hotinglok: I'm not sure if it's worth replacing the values in this override file because it might look odd with half the values being in the psammead spacings and half very specific. I've left it as it is currently because I think it might be more readable with all rem values but I'd like some more thoughts on this.

tochwill: These values seem so specific to keeping things exactly aligned (I'm assuming based on trial and error?) that I think we should keep this as is, and then take another look at a solution that doesn't require this overrides file at all

tochwill: (In another PR, once we've released this)

thekp (approved):

packages/components/psammead-most-read/src/testHelpers/doubleDigitOverride.js
DenisHdz: That sounds sensible 👍

DenisHdz (approved):

PR #3109, opened 2020-02-11
 Media Indicator - Pass dir

DenisHdz:
Resolves #NUMBER

**Overall change:** 
 Pass direction to the Media Indicator to apply the right padding.

**Code changes:**
- Pass `dir` prop to the `MediaIndicator` component
- Update stories
- Update tests

Before:
padding

After:
padding-fixed

---

### DIFF
--- a/packages/components/psammead-media-indicator/src/__snapshots__/index.test.jsx.snap
+++ b/packages/components/psammead-media-indicator/src/__snapshots__/index.test.jsx.snap
@@ -50,6 +50,7 @@ exports[`MediaIndicator should render audio indicator correctly 1`] = `
 <div
   aria-hidden="true"
   class="c0"
+  dir="ltr"
 >
   <div
     class="c1"
@@ -123,6 +124,7 @@ exports[`MediaIndicator should render photogallery correctly 1`] = `
 <div
   aria-hidden="true"
   class="c0"
+  dir="ltr"
 >
   <div
     class="c1"
@@ -182,6 +184,7 @@ exports[`MediaIndicator should render video by default 1`] = `
 <div
   aria-hidden="true"
   class="c0"
+  dir="ltr"
 >
   <div
     class="c1"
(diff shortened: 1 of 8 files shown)


packages/components/psammead-media-indicator/src/index.stories.jsx
```diff
@@ -91,12 +102,13 @@ storiesOf('Components|MediaIndicator/Audio', module)
...
+    ({ script, service, dir }) => (
       <>
         <MediaIndicator
           type="audio"
           script={script}
           service={service}
+          dir={dir}
           isInline={boolean('inline?', true)}
         />
         <StyledHeadline script={script} service={service} promoHasImage={false}>
```
sareh, line 112: Could you please also update the text 'example text' to use the localised text from Storybook helpers? It'd particularly be useful when looking at RTL services. https://github.com/bbc/psammead/blob/4f6e4e9f621d3c917cfbeeb7b7727468b703f608/packages/components/psammead-media-indicator/src/index.stories.jsx#L116

packages/components/psammead-media-indicator/src/__snapshots__/index.test.jsx.snap
```diff
@@ -396,6 +401,7 @@ exports[`MediaIndicator should render video indicator correctly when inline on R
 <div
   aria-hidden="true"
   class="c0"
+  dir="ltr"
```
OlgaLyubin: This should be "rtl" since it is for the test `MediaIndicator should render video indicator correctly when inline on RTL 1`. It's getting the default value because `dir='rtl'` is not passed to that test

DenisHdz: True, forgot to pass it also in the tests 👍

OlgaLyubin (commented):
Should also add `dir` prop to README's table of props

OlgaLyubin (approved):
Looks good 👍

sareh (approved):

PR #3152, opened 2020-02-17
(POC) MostRead ItemWrapper Refactor

hotinglok:
Resolves #NUMBER

**Overall change:** 
Potential way of refactoring `ItemWrapper` and `doubleDigitOverrides` to make it slightly more maintainable. 
Currently, the `doubleDigitOverrides` exists in order to add more spacing to all items in the second column except the tenth in the two column breakpoints as well as increase the spacing for all items except the fifth and the tenth in the largest breakpoint in order to match the spacing design requirements in zeplin. These tiny adjustments for each service were also mainly made because of the reliance on `auto` being the default value for the `min-width` leading to a very messy and hard to maintain file.

The method in this PR utilises the `ItemWrapper` a bit more by doing the reverse. Rather than adding specific padding to all items except the tenth to match it, it sets values that all items adhere to so that the wrapper is actually used as a wrapper. For single column break points, the function hasn't changed. For two column layouts, values have been set for all items in either the first or second column. For the last break point, the fifth and tenth items have the same spacing as each other which is different from every other item.

Another major change made is that some services have fonts decidedly 'thinner' than most others (only chosen via the original `doubleDigitOverrides` values which were <2rem). These fonts have their own values to compensate for the glyphs.

This reduces the total number of individual objects in this file from 30 to 5 and utilises GEL_SPACING values where possible which makes it easier to maintain.

**Code changes:**
- All changes only made in `Rank` and `doubleDigitOverride`
- Added `doubleDigitCheck` to decide if single digit or double digit values are to be used.
- Added `fontWeight` to decide if thin or default values are to be used.
- Added values such as `group0_column` in order to specify values for second column items.

---

### DIFF
--- a/packages/components/psammead-most-read/src/Item/__snapshots__/index.test.jsx.snap
+++ b/packages/components/psammead-most-read/src/Item/__snapshots__/index.test.jsx.snap
@@ -168,13 +168,13 @@ exports[`MostReadItemWrapper should render ltr correctly with 10 items 1`] = `
 
 @media (min-width:37.5rem) and (max-width:80rem) {
   .c3 {
-    min-width: auto;
+    min-width: 2.5rem;
   }
 }
 
 @media (min-width:80rem) {
   .c3 {
-    min-width: 2rem;
+    min-width: 2.5rem;
   }
 }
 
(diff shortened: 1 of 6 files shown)


hotinglok:
Closing until the refactor discussions begin taking place.

hotinglok:
Closing due to updated PR for it here: https://github.com/bbc/psammead/pull/3201

PR #3160, opened 2020-02-19
Uniformise the visually-hidden state label in Bulletins and Radio Schedules

### LINKED ISSUE #3037: Uniformise the visually-hidden state label in story promos, bulletins, and radio-Schedules
The visually-hidden `live` state label is implemented differently in these three components, we should refactor this to have a more uniform implementation.

The requirements for what should be shown are as follows:

- We only need visually hidden text for ‘Live’ in English as a hack to stop screen readers reading out ‘live’ incorrectly, we don’t need it for other languages.
- We _don’t_ want to read out ‘Live’ in English when we are using a different language, so it looks like the bulletin component is currently incorrect. The requirement is we use the english word 'Live' for LTR languages, or the translation for the service in RTL languages. We have no RTL languages using this component yet, so this won’t have impacted anyone.

**Bulletins**
Screen Shot 2020-01-30 at 3 42 37 PM

**Story promos**
Screen Shot 2020-01-30 at 3 41 54 PM

**Radio schedules**
Screen Shot 2020-01-30 at 4 03 35 PM

For flexibility, we should consider passing in two props - `offscreenLiveText` `liveText` and if we don't pass in `offscreenLiveText` there should not be an associated VisuallyHiddenText component used.

**Testing notes**
Regression test the live label working as expected as a part of this refactor.

**Additional context**

OlgaLyubin:
Resolves #3037

**Overall change:** 
Refactored visually-hidden live state label since it is implemented differently in Radio Schedules and Bulletins.

**Code changes:**
- Passed offscreenLiveText prop so that VisuallyHiddenText would not be rendered in case offscreenLiveText is not provided.

---


packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -136,6 +136,7 @@ const renderHeaderContent = ({
   startTime,
   timezone,
   locale,
+  offscreenLiveText,
```
EinsteinNjoroge: I think the `offscreenLiveText` is the same as the `stateLabel`, why don't we use that instead?

packages/components/psammead-radio-schedule/src/__snapshots__/index.test.jsx.snap
```diff
@@ -1846,11 +1827,6 @@ exports[`RadioSchedule should render rtl radio schedules correctly 1`] = `
             >
               NEXT
             </span>
-            <span
-              class="c13"
-            >
-              ,
```
EinsteinNjoroge: Looking at the accessibility notes 5,6 and 7 on [these designs ](https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/5dc0724850731722a4aa767a), the visually hidden text should be `[State], [Brand title], [00:00], Episode [Episode title]`. Removing this comma will cause the `state` to be read in one breath with the `Brand title`.

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
OlgaLyubin: Changed it:)

packages/components/psammead-radio-schedule/src/__snapshots__/index.test.jsx.snap
OlgaLyubin: Good point, done now

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -202,12 +202,12 @@ const Bulletin = ({
             <span role="text">
               {offScreenText && (
                 <VisuallyHiddenText lang={lang}>
-                  {`${offScreenText},`}
+                  {`${offScreenText}, `}
```
OlgaLyubin, line 205: This is added so that in CSS disabled mode state is displayed with spaces instead of how it was before:
Screenshot 2020-02-21 at 11 49 26

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -164,7 +164,7 @@ const renderHeaderContent = ({
             {`${stateLabel.toUpperCase()}`}
           </LabelWrapper>
           <VisuallyHiddenText {...hiddenTextProps}>
```
sareh, line 166: We don't want to have a `<VisuallyHiddenText>,</VisuallyHiddenText>` if the state label is not 'Live', so we should do the check prior to the component.
```
{ stateLabel === 'Live' && <VisuallyHiddenText {...hiddenTextProps}>{` ${stateLabel}, `}</VisuallyHiddenText> }
```

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -164,7 +164,7 @@ const renderHeaderContent = ({
             {`${stateLabel.toUpperCase()}`}
```
sareh, line 164: Uppercase does not exist for several scripts, e.g arabic. We should check that it makes sense for all scripts when carrying out a UX review.

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
OlgaLyubin: I had this condition previously but changed it because of https://github.com/bbc/psammead/pull/3160#discussion_r381312079 . Seems that we need that comma for separating state and brand title even if the Program is not live

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
OlgaLyubin: Looks like when it is a script that does not support capital letters, they are just not getting updated e.g. arabic text just stays the same as it is. But will check for other services too

packages/components/psammead-radio-schedule/src/testHelpers/helper.jsx
```diff
@@ -57,8 +58,8 @@ export const renderProgramCard = ({
       startTime={startTime}
       state={state}
       link={articlePath}
-      durationLabel={durationLabel}
-      stateLabel={stateLabel}
+      durationLabel={dir === 'rtl' ? 'المدة الزمنية' : durationLabel}
```
DenisHdz: Can this be defined above instead of inside the `prop`?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -164,7 +164,7 @@ const renderHeaderContent = ({
...
           </LabelWrapper>
           <VisuallyHiddenText {...hiddenTextProps}>
-            {isLive && ` ${stateLabel}`}
+            {stateLabel === 'Live' && ` ${stateLabel}`}
```
Bopchy, line 167: Looks good to me. Just wondering if there is any difference between using `isLive` and `stateLabel === 'Live' `?

OlgaLyubin: `isLive` is used because a comma will be rendered for live programs for every service. And `stateLabel === 'Live'` is used so that the `stateLabel` is rendered only for some services that need "Live" pronounced correctly by a screen reader. For example, in RTL services for a live card `stateLabel === 'Live'` will be `false` because the label will be translated but `isLive` will be true

Bopchy (approved):

packages/components/psammead-radio-schedule/src/testHelpers/helper.jsx
```diff
@@ -19,7 +19,7 @@ const getSchedule = (service, withLongSummary) => {
   return stateTypes.map((state, index) => ({
     id: index,
     state,
-    stateLabel: sentenceCase(state),
+    stateLabel: service === 'arabic' ? 'مباشر' : sentenceCase(state),
```
thekp, line 25: LGTM overall. Just something small... for consistency, as below we can define these props (durationLabel/stateLabel) above.

OlgaLyubin: Defined `durationLabel` above, but this cannot be done for `stateLabel` as it uses a `state` parameter from mapping

thekp: makes sense, thanks Olga 💯

thekp (approved):
thanks for the update 💯

paruchurisilpa:
Checked Bulletin and Radio Schedule for <VisuallyHiddenText> in Storybook. For Radio Schedules LTR services that are live  have a `<VisuallyHiddenText>` with “Live” label and a comma. Non live programs and live RTL services should  have a <VisuallyHiddenText>  with a comma. Checked for Bulletins  with no-CSS mode, it is rendered normally ..

paruchurisilpa:
@PriyaKR can you please pick up the accessibility side of things for this PR please, even the frontpage regression is done..

PriyaKR:
This can be merged as accessibility swarm would be done once this PR gets merged and available on environments.

PriyaKR:
Looks good to me.Regression tested the component on pages and also the tests run ok locally.

PR #3181, opened 2020-02-26
Alias radio schedule to be a list

thekp:
Resolves n/a

**Overall change:** _Alias the grid components in radio schedules as `<ul>` for the parent grid and `<li>` the children._

**Code changes:**

- _Pass the forwardAs prop in components._
- documentation in styled components: https://styled-components.com/docs/api#as-polymorphic-prop
image

---

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Automated jest tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

PR #3184, opened 2020-02-27
Change display of program cards in group3 & update stories to include 4 cards

### LINKED ISSUE #3134: Radio Schedules: Hide oldest on Demand card on Group 3
The v1 designs for radio schedules do not include the carousel. Instead, we stack 4 cards on breakpoints lower than group 3, show 3 cards in a grid and hide the 3rd card on group 3, and show all 4 in a grid on breakpoints greater than group 3.

- Add logic to be able to hide a card between these breakpoints, based on a prop we can pass in within Radio Schedules.

- Update the Psammead storybook stories to show 4 programmes.

**Testing notes**
At viewport 600-1007px only three cards should be shown in the radio schedule component.

**Additional context**

hotinglok:
Resolves https://github.com/bbc/psammead/issues/3134

**Overall change:** 
Changes the grid layout of the component in group 3 from having the 4th card overflow to another row to having a 2x2 layout instead. Also updated stories to have 4 cards rather than 3.
Originally we were to hide a card so the layout would be:
image
Now UX have agreed on this layout so we don't hide any information just on one breakpoint:
image

**Code changes:**

- Changed `group3` of `programCardGridProps` from 2 to 3 in order to make 2x2 layout.
- Just added `onDemand` twice to `stateTypes` to add another card easily to stories and changed the order to be correct.
- Updated tests to use just the 3 unique state types.
- Added roles to each grid from https://github.com/bbc/psammead/pull/3181#issuecomment-593337683
---

### DIFF
--- a/packages/components/psammead-radio-schedule/CHANGELOG.md
+++ b/packages/components/psammead-radio-schedule/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 0.1.0-alpha.15 | [PR#3184](https://github.com/bbc/psammead/pull/3184) Change display of program cards in group3 & update stories to include 4 cards |
 | 0.1.0-alpha.14 | [PR#3164](https://github.com/bbc/psammead/pull/3164) Talos - Bump Dependencies - @bbc/psammead-timestamp-container |
 | 0.1.0-alpha.13 | [PR#3153](https://github.com/bbc/psammead/pull/3153) Talos - Bump Dependencies - @bbc/psammead-timestamp-container |
 | 0.1.0-alpha.12 | [PR#3132](https://github.com/bbc/psammead/pull/3132) Render correct hidden `startTime` in RadioSchedule component |
--- a/packages/components/psammead-radio-schedule/src/index.jsx
+++ b/packages/components/psammead-radio-schedule/src/index.jsx
@@ -1,5 +1,9 @@
 import React from 'react';
 import styled from 'styled-components';
+import {
+  GEL_GROUP_3_SCREEN_WIDTH_MIN,
+  GEL_GROUP_3_SCREEN_WIDTH_MAX,
+} from '@bbc/gel-foundations/breakpoints';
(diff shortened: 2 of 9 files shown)


packages/components/psammead-radio-schedule/CHANGELOG.md
```diff
@@ -3,7 +3,11 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+<<<<<<< HEAD
```
OlgaLyubin: Left this after conflict resolving 🙈

packages/components/psammead-radio-schedule/src/testHelpers/helper.jsx
```diff
@@ -11,7 +11,7 @@ export const sentenceCase = text =>
     .charAt(0)
     .toUpperCase() + text.substring(1);
 
-export const stateTypes = ['onDemand', 'live', 'next'];
+export const stateTypes = ['live', 'onDemand', 'onDemand', 'next'];
```
OlgaLyubin, line 16: Since now we have two `onDemand`s here, in `ProgramCard` test `should render correctly for ${state}` where we go through each state type, we will have two identical snapshots for 'onDemand'. Can you please fix that to avoid repetition?

packages/components/psammead-radio-schedule/src/index.jsx
```diff
@@ -102,7 +102,7 @@ const programGridProps = {
     group0: 4,
     group1: 4,
     group2: 6,
-    group3: 2,
+    group3: 3,
```
EinsteinNjoroge, line 109: Might be useful to add a screenshot in the description for this change

packages/components/psammead-radio-schedule/src/testHelpers/helper.jsx
hotinglok: Nice catch, fixed now

packages/components/psammead-radio-schedule/CHANGELOG.md
hotinglok: ohno

OlgaLyubin (approved):
Thank you for the changes, looks good 👍

EinsteinNjoroge (approved):
LGTM

PR #3199, opened 2020-03-03
Update most read numerals

thekp:
Resolves n/a

**Overall change:** _Use Latin numerals for Arabic and Urdu_

**Code changes:**

- _remove EasternArabic numerals for Arabic and Urdu._
- _Updating double-digit spacing for Arabic and Urdu._

URDU:
image

ARABIC: 
image

---

- [x] I have assigned myself to this PR and the corresponding issues
- [ ] Automated jest tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

PR #3201, opened 2020-03-03
Refactor RankWrapper in MostRead

hotinglok:
Resolves #NUMBER

**Overall change:** 
Updated version of the POC here: https://github.com/bbc/psammead/pull/3152, Overall change is regarding the ItemWrapper and how this PR utilises it more by applying the spacing in `doubleDigitOverrides` to all ranks rather than just targeting the 5th or 10th element. More description can be found in the original POC.

** Additional Context**
- `doubleDigitOverrides` was originally a file filled unique spacing values for most services due to the spacing requirement of how there needed to be more spacing between the rank and the link if there was 10 items due to there it being a double-digit numeral. 
- This spacing (just giving the wrapper around the rank a `min-width`) was applied to ranks either just in the second column excluding the 10, just the 5th or 10th item, or all of the ranks depending on the breakpoint and if there was even a double-digit numeral present. In hindsight, this was a very bad approach that was very unmaintainable as evident when we run into bugs such as https://github.com/bbc/psammead/pull/3199. The explanation for all the values is also non-existent besides the fact that they were chosen by looking by eye. 

**Why this POC makes it a little better**
- Instead of targeting specific items with varying conditions, this firstly refactors the `ItemWrapper` to always target items in groups so they will definitely be the same. For example, if there is a double-digit number present and there's two columns, it will apply the same spacing to all ranks in the first column and apply the new extra spacing to all items in the second column. Similarly, the fifth and tenth item in the multi-column layout will also always be the same whilst the other 8 items have a different spacing.
- This uses GEL_SPACING values besides in two places (one of which can still be considered a GEL_SPACING if it is added to the file).
- This POC reduces 98% of the unique values which were present in the previous iteration. This is significantly more maintainable even if this doesn't completely eliminate this file.

**How to review and compare changes**
- Opening up the branch and seeing MostRead in storybook, you can inspect any `li` and inspect the `StyledItem` part of it. If you hover over the styles, you can see that all the padding lines up with each other unlike what you might see in:https://github.com/bbc/psammead/pull/3199
Example below:
image
- The things to look out for are any odd out of place items, the fact that the same spacing is now applied to all items given certain criteria should mean that there are never any instances where this spacing looks out of place or one seems out of line with the others for all breakpoints.

**Code changes:**
- All changes only made in `Rank` and `doubleDigitOverride`
- Added `doubleDigitCheck` to decide if single digit or double digit values are to be used.
- Added `fontWeight` to decide if thin or default values are to be used.
- Added values such as `group0_column` in order to specify values for second column items.
- Add `letter-spacing: -4px` to Japanese as the double-digit is to wide.

---


packages/components/psammead-most-read/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 3.0.0 | [PR#3201](https://github.com/bbc/psammead/pull/3201) Refactor ItemWrapper in MostRead |
```
thekp:
```suggestion
| 3.0.0 | [PR#3201](https://github.com/bbc/psammead/pull/3201) Refactor RankWrapper in MostRead |
```

thekp:
**Note:**

The logic in this PR for setting min-width will mean there is a bias for some services. For example, this PR is optimised for the numerals for `News` however for a service like Zhongwen, the space between rank and headline text will be a bit larger 🙈 

News: 
image
Zhongwen: 
image

I have discussed this problem with UX and it has been agreed that this is ok: as long as all items align with each other with reasonable spacing. 
Another problem was Japanese's digit `10` was too wide, the solution for this was to add some specific css for japanese.

packages/components/psammead-most-read/src/Rank/index.jsx
```diff
@@ -37,53 +54,80 @@ const listHasDoubleDigits = ({ numberOfItems }) => numberOfItems >= 9;
 const columnIncludesDoubleDigits = (props, supportsGrid) =>
   isOnSecondColumn(props, supportsGrid) && listHasDoubleDigits(props);
```
mmcculloch: Should this function 
```js
const listHasDoubleDigits = ({ numberOfItems }) => numberOfItems >= 9;
```
Be more like
```js
const listHasDoubleDigits = numberOfItems => numberOfItems > 9;
```
Was thinking a few things:
1. Might not make sense to take an object where there's only one value used.
2. Would it not be greater than 9 rather than equal to or greater than 9?

packages/components/psammead-most-read/src/utilities/rankMinWidth.js
```diff
@@ -0,0 +1,59 @@
...
+  group3: GEL_SPACING_QUAD,
+  group5: GEL_SPACING_QUAD,
+};
+
+// If numberOfItems >= 10, extra spacing needs to be accounted for.
+export const doubleDigitDefault = {
+  group3: GEL_SPACING_QUAD,
+  group5: GEL_SPACING_QUAD,
+  // These values are used to align the rank when a double digit exists in the column
+  group0_1_column: GEL_SPACING_QUIN,
```
mmcculloch: Could these be `group1WithOneColumn`, `group3WithTwoColumns`, `group5WithFiveColumns` ? The current names don't give away the intent very well, and maybe plain english makes more sense (and keeping to camelcase convention).

thekp: Yep that's much better

mmcculloch (approved):
Would be good to not rely on hardcoded lists of services, but this is an improvement on what's currently there. :)

EinsteinNjoroge (commented):
Overall looks good, will require a UX review.
@lokh01 have you viewed these changes in IE?

thekp:
@EinsteinNjoroge 

> have you viewed these changes in IE?

Still looks like it works on IE (browserstack): 
image

Chromatic changes:
image



> Overall looks good, will require a UX review.

I had a UX review yesterday with Laura. Commented some notes here: https://github.com/bbc/psammead/pull/3201#issuecomment-595162799

EinsteinNjoroge (approved):

PriyaKR:
Looks good to me. @thekp do you want to check the persian screenshot with UX before you merge this PR?
Screen Shot 2020-03-09 at 11 49 19

thekp:
> Looks good to me. @thekp do you want to check the persian screenshot with UX before you merge this PR?

Thanks for testing Priya.

I have confirmed this issue with Laura from UX. She is for us to fix forward on this extra-large spacing, as all the items are still aligned but the spacing is just not optimal :( 
Created a bug issue for this: https://github.com/bbc/psammead/issues/3233 🐛

thekp:
image
I did not see this before on IE 😭

thekp:
This PR was not the cause of the IE bug above, it is already occurring on the latest version of Most Read. A bug issue has been raised for it here: https://github.com/bbc/psammead/issues/3234

I'm going to merge this PR as it cleans up the code most read a bit.

PR #3218, opened 2020-03-05
Create Psammead Include

FK78:
Resolves https://github.com/bbc/simorgh-infrastructure/issues/882

**Overall change:** _Create the psammead-include component._

**Code changes:**

- _Created `psammead-include` component using `npm run create:package`_
- _Update tests for `psammead-include`_
- _Update story for `psammead-include`_

---


packages/components/psammead-include/src/index.jsx
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable react/no-danger */
+import React from 'react';
+import { string } from 'prop-types';
+
+const Include = ({ html }) => {
```
HarveyPeachey, line 5: As the only thing this component is doing is using `dangerouslySetInnerHTML` with a single prop of HTML should we tie it with the naming convention of Include? I feel like the naming is very specific to something which is internal to the BBC.

packages/components/psammead-include/README.md
```diff
@@ -0,0 +1,53 @@
...](https://snyk.io/test/github/bbc/psammead/badge.svg?targetFile=packages%2Fcomponents%2Fpsammead-include%2Fpackage.json)](https://snyk.io/test/github/bbc/psammead?targetFile=packages%2Fcomponents%2Fpsammead-include%2Fpackage.json) [![Dependency Status](https://david-dm.org/bbc/psammead.svg?path=packages/components/psammead-include)](https://david-dm.org/bbc/psammead?path=packages/components/psammead-include) [![peerDependencies Status](https://david-dm.org/bbc/psammead/peer-status.svg?path=packages/components/psammead-include)](https://david-dm.org/bbc/psammead?path=packages/components/psammead-include&type=peer) [![Storybook](https://raw.githubusercontent.com/storybooks/brand/master/badge/badge-storybook.svg?sanitize=true)](https://bbc.github.io/psammead/?path=/story/include--containing-image) [![GitHub license](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](https://github.com/bbc/psammead/blob/latest/LICENSE) [![npm version](https://img.shields.io/npm/v/@bbc/psammead-include.svg)](https://www.npmjs.com/package/@bbc/psammead-include) [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](https://github.com/bbc/psammead/blob/latest/CONTRIBUTING.md)
```
HarveyPeachey, line 1: Information in here needs filling out

packages/components/psammead-include/src/index.jsx
FK78: @andrewscfc will be using this component for his IDT1 work.

andrewscfc: I will be using it but one thing I would say is much of the heavy lifting is being done in simorgh in a container i.e. fetching the include from the endpoint. That brings us back to the discussion about whether we should have a psammead component at all for this, maybe just 'dangerously set' the html in the container and don't bother with a component?

The story demonstrates some of the value as it takes a chromatic snapshot of a real include successfully loading, it encapsulates the knowledge that embedding an include just involves dangerously setting the html.

Let's see what this looks like with the simorgh PR and have another think?

HarveyPeachey: Maybe having it as a component in Simorgh? This component does a very similar thing https://github.com/bbc/simorgh/blob/latest/src/app/components/IfAboveIE9Comment/index.jsx. 

I think you're right though using it in a Simorgh PR and thinking about it later

FK78:
After further discussion, the team has come to the conclusion that the `Include` component isn't needed in Psammead as it can be added to Simorgh and maintained without inconvenience.

PR in Simorgh: https://github.com/bbc/simorgh/pull/5767

PR #3224, opened 2020-03-06
Create Live Label component

### LINKED ISSUE #3203: Create Live Label component
We currently use Live labels with the associated visually hidden text and aria hidden attribute in three different places in our code base. This leads to updates to usage needing to be made in multiple places - story promo, bulletin promo and usage within the parent application. 
E.g. See all of the PRs linked to this issue: https://github.com/bbc/simorgh/issues/5749

We should have a single reusable component that handles all of this logic and can be used in various contexts. 

It should be a self-contained package that can be used on pages in English as well as those not in English. Passing in translation when needed and applying relevant aria roles if necessary. 

We could choose to keep it as it currently is, with implementation duplicated in multiple places. 

**Testing notes**
[Tester to complete]

**Additional context**

DenisHdz:
Resolves #3203

**Overall change:** 
Create Live Label component

**Code changes:**
- Create a new alpha component for Live Label

---


packages/components/psammead-livel-label/src/index.test.jsx
```diff
@@ -0,0 +1,31 @@
...
+      service="news"
+      ariaHidden
+      withOffScreenText
+      offScreenText="Watch Live"
+    />,
+  );
+
+  shouldMatchSnapshot(
+    'should correctly render for RTL service',
+    <LiveLabel service={arabic} ariaHidden withOffScreenText />,
```
OlgaLyubin:
```suggestion
    <LiveLabel service={arabic} dir="rtl" ariaHidden withOffScreenText />,
```

packages/components/psammead-livel-label/src/index.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+LiveLabel.propTypes = {
+  service: string.isRequired,
+  dir: oneOf(['rtl', 'ltr']),
+  ariaHidden: bool,
+  liveText: string,
+  withOffScreenText: bool,
+  offScreenText: string,
+  lang: string,
+  children: node,
```
OlgaLyubin, line 53: `children` are set as a required prop in the readme but optional here

DenisHdz: It is not required. Wrongly copied from another README.

packages/components/psammead-livel-label/src/index.stories.jsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import styled from 'styled-components';
+import { storiesOf } from '@storybook/react';
+import { withKnobs } from '@storybook/addon-knobs';
+import { withServicesKnob } from '@bbc/psammead-storybook-helpers';
+import { Headline, Link } from '@bbc/psammead-story-promo';
+import LiveLabel from './index';
+import notes from '../README.md';
+
+const Wrapper = styled.div`
```
OlgaLyubin, line 10: Why do we need this `Wrapper`? Live label with a headline seems to be displayed in the same way if you remove it

DenisHdz: I've added it to avoid the whole story view to be clickable.

OlgaLyubin (approved):
Looks great 👍 Just needs snapshots to be updated

hotinglok (commented):
~Looks good, just wondering if there's anything we could do about the slight shift down once there is also a headline. When you click between `with custom offscreen text` and `with headline`, the component shifts down a bit. I think this is because the default line-height is `1.15rem` whereas the headline with Reith makes it `1.25rem`. Just a minor nitpick though.~

Quoted from Denis on slack because of github issues:
> I understand your point, but I personally think it's not an issue because stories can be independent from each other and as you said, the headline is using a different line-height so it is expected to happen.
We could just pass some text wrapped in a span as a children, but I would prefer to keep the headline as it is a more realistic user case.

That's a reasonable explanation, and it was just a very minor thing anyways so LGTM.

mmcculloch (approved):

hotinglok (approved):

PR #3247, opened 2020-03-12
Updating Serbian time stamp formats

### LINKED ISSUE #3246: Update Serbian timestamp wordings
Some of the wording around timestamps is grammatically off in Serbian.

Update the following in the `psammead-locales` package:

xMinutesAgo:  
which currently is:
lat: Pre minuta: {x}
cyr: Пре минута: {x}
but it would be grammatically better if it were:
lat: Pre {x} minuta
cyr: Пре {x} минута

For the following however, it's more complex, as there is a need to use a different word for "Hours" when the number of hours is less than 5, versus the word used when the number of hours is 5 or more.

Investigate how this can be achieved with moment.js as we are using it currently - see https://stackoverflow.com/questions/38367038/format-relative-time-in-momentjs for an example.

to fix xHoursAgo which currently is:
lat: Pre sati: {x}
cyr: Пре сати: {x}
it would need to be:
for x < 5 :
lat: Pre {x} sata
cyr: Пре {x} сата
and for x >= 5 :
lat: Pre {x} sati
cyr: Пре {x} сати

N/A

**Testing notes**
Include this scenario in testing - published timestamp greater than and less than 5 hours.

**Additional context**
https://github.com/bbc/psammead/tree/latest/packages/utilities/psammead-locales/moment

jimjohnsonrollings:
Resolves #3246  

**Overall change:**
Updates Serbian timestamps to be more better grammar

**Code changes:**

- Updates the Serbian word order for 'x minutes' and 'y hours', and the plural logic around 'hours', in both latin and cyrillic
- Updates tests to match
- Bumps version number of package

---


packages/utilities/psammead-locales/moment/sr.js
```diff
@@ -10,8 +10,12 @@ moment.updateLocale('sr', {
...
     past: 'Pre %s',
     m: '1 minuta',
-    mm: 'minuta: %d',
+    mm: '%d minuta',
     h: '1 sata',
-    hh: 'sati: %d',
+    hh: function(number) {
+      return number < 5
+        ? String(number).concat(' sata')
+        : String(number).concat(' sati');
```
jimjohnsonrollings: eslint didn't like my first two approaches here, which were:
`number + ' sata'`
and
``${number} sata``

sareh: This should work
```
`${number} sata`
```

jimjohnsonrollings: It did not.  Sorry github removed the ` symbols in my comment. Eslint did not want me to use that format.

jimjohnsonrollings:
Code Climate isn't happy, but I think it's unhappy in a way we are happy to ignore

packages/utilities/psammead-locales/moment/sr.js
tochwill: Hmm. The `moment` folder has its own linting rules that specify we can only use es5 in there. Does anyone know if this is a limitation of moment.js? But there is also a linting rule defined that says we have to use templating instead of concatenation, which is why it won't let you use either. So either we allow es6, but I'm sure it's been disabled for a reason, or we allow string concatenation (in the moment linting config, not everywhere).

DenisHdz: For context, these linting rules were added in [this](https://github.com/bbc/psammead/pull/2080) PR, (https://github.com/bbc/psammead/pull/2080/files#r323281429)

DenisHdz:
```suggestion
        : String(number).concat(' ', 'sati');
```

jimjohnsonrollings: What difference would that make?

DenisHdz: It should be the same, but I think it is safer than adding a space character to the string.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat

jimjohnsonrollings: Updated per your recommendation :)

tochwill:
Just taken a look - I do think we need to update the linting rules rather than just using whatever we can get to work, we can set `'prefer-template': 0` in the eslint file in the moment folder to be able to just use string concatenation for now.

packages/components/psammead-grid/src/__snapshots__/index.test.jsx.snap
```diff
@@ -40,7 +40,7 @@ exports[`Grid component should render Grid with Grid items 1`] = `
   }
 }
 
-@media (min-width:63rem) and (max-width:79.9375rem) {
+@media (min-width:63rem) and (max-width:80rem) {
```
DenisHdz: Howcome did these snapshots get updated in this PR? Also, they are wrong since we've updated the `GEL_GROUP_4_SCREEN_WIDTH_MAX` a few weeks ago to be `79.9375rem; // 1279px` instead of  `80rem; // 1280px`.

Try running `npm ci` to install the latest versions of the packages on your branch.

jimjohnsonrollings: I'm trying that but not having much luck yet...

DenisHdz: I've done it for you :)

DenisHdz (approved):
LGTM

packages/components/psammead-grid/src/__snapshots__/index.test.jsx.snap
jimjohnsonrollings: Thanks :D

sareh (approved):
:+1: Looks great.

tochwill (approved):

PriyaKR:
The hours wording change looks good to me but there are no examples for minutes in storybook.

PriyaKR:
Looks good to me.
Screenshot 2020-03-18 at 12 25 58
Screenshot 2020-03-18 at 12 26 07

PR #3252, opened 2020-03-13
Android talkback issue in Navigation

### LINKED ISSUE #2999: Scrollable Navigation - Android Talkback announces link twice when navigation scrolls
**Describe the bug**
If a navigation item begins off screen, and focusing that item causes it to be brought onto the screen, talkback announces the text within that item twice (once before the link is brought into view, and once after).

**To Reproduce**

1. Go to 'localhost:7080/afrique'
2. Using talkback, scroll through the scrollable navigation until you reach an item that is off screen.

**Expected behaviour**
The item should only be announced once.

**Screenshots**

**Testing notes**

**Additional context**

mmcculloch:
Resolves #2999 

**Overall change:** 
When using Android Talkback and moving between nav items `scroll-behavior: smooth;` was causing talkback to repeat the item 2-3 times because of the scroll speed. This change reverts it back to auto, with a comment to indicate it shouldn't be set to smooth.

A few other things that I checked and discovered
- This style is only applied in mobile view, so disabling it has no effect on Desktop.
- This style is not implemented in Safari, so there is no different in iOS Safari accessibility
- This style only has an effect when tabbing with accessibly settings, but generally it is considered back accessibility practice to enable it https://stackoverflow.com/questions/51152749/smooth-scrolling-while-navigating-with-tab-key

**Code changes:**

- CSS change to change `scroll-behavior: smooth;` to `scroll-behavior: auto;`

---

### DIFF
--- a/packages/components/psammead-navigation/CHANGELOG.md
+++ b/packages/components/psammead-navigation/CHANGELOG.md
@@ -3,7 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
-| 6.0.6 | [PR#3177](https://github.com/bbc/psammead/pull/3252) Remove `scroll-behavior: smooth;` to from StyledScrollableNav |
+| 6.0.6 | [PR#3252](https://github.com/bbc/psammead/pull/3252) Remove `scroll-behavior: smooth;` from StyledScrollableNav |
 | 6.0.5 | [PR#3177](https://github.com/bbc/psammead/pull/3177) Add `cursor: pointer` to Menu button |
 | 6.0.4 | [PR#3135](https://github.com/bbc/psammead/pull/3135) Talos - Bump Dependencies - @bbc/gel-foundations |
 | 6.0.3 | [PR#3129](https://github.com/bbc/psammead/pull/3129) Talos - Bump Dependencies - @bbc/gel-foundations |
--- a/packages/components/psammead-navigation/CHANGELOG.md
+++ b/packages/components/psammead-navigation/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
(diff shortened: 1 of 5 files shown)


packages/components/psammead-navigation/src/ScrollableNavigation/index.jsx
```diff
@@ -15,7 +15,9 @@ const StyledScrollableNav = styled.div`
   @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
     white-space: nowrap;
     overflow-x: scroll;
-    scroll-behavior: smooth;
+
+    /* Avoid using smooth scrolling as it causes accessbility issues */
```
DenisHdz:
```suggestion
    /* Avoid using smooth scrolling as it causes accessibility issues */
```

DenisHdz (approved):
Thanks for investigating 👍

packages/components/psammead-navigation/src/ScrollableNavigation/index.jsx
mmcculloch: doh, I'm blaming my mac keyboard on this :)

packages/components/psammead-navigation/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 6.0.6 | [PR#3177](https://github.com/bbc/psammead/pull/3252) Remove `scroll-behavior: smooth;` to from StyledScrollableNav |
```
j-pendlebury:
```suggestion
| 6.0.6 | [PR#3252](https://github.com/bbc/psammead/pull/3252) Remove `scroll-behavior: smooth;` from StyledScrollableNav |
```

j-pendlebury (approved):
Just one thing, otherwise looks great 👍

PR #3297, opened 2020-03-24
Use Live Label in Radio Schedules, and pass in translations

### LINKED ISSUE #3296: Live and Next Labels are incorrect in Radio Schedules
**Describe the bug**
Currently, the 'Live' label, and the 'Next' label are always being displayed in English.

Instead of passing in a 'stateLabel' that is part of the schedule data, we should instead just pass all the translations as props, from the service config, and let the component decide which to show based on the state.

It appears that currently in Simorgh, we are passing 'next' or 'live' as the state label to be used for the string that is shown to the user, and uppercasing it in the Psammead component. We should remove this. We should pass the translation strings in as they appear in config, and then render them.

We are also setting visually hidden text incorrectly for when 'Live' is being displayed in english:

Screenshot 2020-03-24 at 15 37 54

This is because there is a check in the Psammead component that the state label is the string 'Live', which it is not going to be. We should again remove this check, and explicitly pass in the visually hidden text.

**To Reproduce**

1. Go to a

**Expected behaviour**
If the programme is live, we show the 'LIVE' label in the service language.
If 'LIVE' is in english, we set this to aria hidden and render the word 'Live' as visually hidden text.
If the programme is on next, we show the 'NEXT' label in the service language.

**Screenshots**

**Testing notes**

**Additional context**

tochwill:
Resolves #3296

**Overall change:**
Live and Next labels are displaying incorrectly in radio schedules. This allows us to pass in translation strings, and reuses the existing live label component.

**Code changes:**
- Passes in Live label and Next label as translations.
- Uses the psammead live label component.

---


packages/components/psammead-radio-schedule/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 0.1.0-alpha.22 | [PR#xxxx](https://github.com/bbc/psammead/pull/xxxx) Use LiveLabel, and pass in translations for next and live |
```
tochwill:
```suggestion
| 0.1.0-alpha.23 | [PR#3297](https://github.com/bbc/psammead/pull/3297) Use LiveLabel, and pass in translations for next and live |
```

tochwill:
Fixing lockfiles...

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -500,16 +491,16 @@ exports[`ProgramCard should render correctly for next 1`] = `
...
           class="c3"
+          dir="ltr"
         >
           NEXT
         </span>
         <span
           class="c4"
         >
-          ,
+          ,
```
sareh: Is this span needed? Can the comma be added in the span above instead?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -153,26 +159,24 @@ const renderHeaderContent = ({
...
-            {`${stateLabel.toUpperCase()}`}
-          </LabelWrapper>
-          <VisuallyHiddenText {...hiddenTextProps}>
-            {stateLabel === 'Live' && ` ${stateLabel}`}
-            {`,`}
-          </VisuallyHiddenText>
+          <NextLabel service={service} script={script} dir={dir}>
+            {nextLabel}
+          </NextLabel>
+          <VisuallyHiddenText>{`, `}</VisuallyHiddenText>
```
sareh: Do we really need this visually hidden. Span with just a comma?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -128,20 +133,21 @@ const programStateConfig = {
...
   script,
   startTime,
   timezone,
   locale,
+  dir,
 }) => {
-  const isOnDemand = state === 'onDemand';
   const isLive = state === 'live';
-  const hiddenTextProps = stateLabel === 'Live' ? { lang: 'en-GB' } : {};
+  const isNext = state === 'next';
```
sareh, line 148: This update reduces the logic here and makes this much cleaner.

In future, I'd rather we pull out this Live/Next logic from the program card entirely and pass it all in as a component, to make it reusable. 
For example, also export the NextLabel as its own component. Pass in everything within the `<HeadingContentWrapper>` as a heading prop instead.

I'm happy with this PR being an incremental improvement and we can raise another issue to discuss refactoring further.

sareh (commented):
Thanks for this. I've added a few comments below.

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -51,6 +51,15 @@ exports[`ProgramCard should render correctly for live 1`] = `
   margin: 0;
 }
 
+.c4 {
+  font-family: ReithSans,Helvetica,Arial,sans-serif;
+  font-weight: 700;
+  font-style: normal;
+  color: #B80000;
+  display: inline-block;
+  margin-right: 0.5rem;
```
tochwill, line 60: The live label component has this much margin at all times. It looks a little much to me when in use here, in the bulletins, and in normal story promos, but I think it would be good to use the live label component in here now and then take another look - we can change it everywhere in one place then :D

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -24,6 +24,7 @@ import { Link } from '@bbc/psammead-story-promo';
 import { oneOf, shape, string, number } from 'prop-types';
 import { scriptPropType } from '@bbc/gel-foundations/prop-types';
 import VisuallyHiddenText from '@bbc/psammead-visually-hidden-text';
+import LiveLabel from '@bbc/psammead-live-label';
```
sareh, line 27: It's great that we're reusing the psammead-live-label component here. ✨

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -153,26 +159,24 @@ const renderHeaderContent = ({
...
+            {nextLabel}
+          </NextLabel>
+          <VisuallyHiddenText>{`, `}</VisuallyHiddenText>
         </>
       )}
-      <span>
-        {!isOnDemand && ` `}
-        {brandTitle}
-      </span>
+      <span>{brandTitle}</span>
```
tochwill: This span can go! (The space before is now added by the label)

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
tochwill: This comes from the visually hidden text (I'll reply to that point below)

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
tochwill: I left this as was, it matches the screen reader UX but I agree (I think you're saying this!) we're adding way too much complexity across the board hiding and showing elements.

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
tochwill: I think that would be a good shout, and can write up a ticket for that.

packages/components/psammead-radio-schedule/src/testHelpers/helper.jsx
```diff
@@ -88,6 +88,8 @@ export const renderRadioSchedule = ({
     timezone={timezone}
     script={script}
     service={service}
+    nextLabel={dir === 'rtl' ? 'مباشر' : 'NEXT'}
```
DenisHdz: Can we move this into a constant as done above with the `renderProgramCard`?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
tochwill: The Live Label does not add this extra hidden comma unless it is within the visually hidden text - I will do the same here and we can decide how to proceed when we move this into its own component.

packages/components/psammead-radio-schedule/src/testHelpers/helper.jsx
tochwill: Sure thing

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -153,26 +159,24 @@ const renderHeaderContent = ({
 
   const content = (
     <HeadingContentWrapper>
-      {!isOnDemand && (
+      {isLive && (
+        <LiveLabel
+          service={service}
+          dir={dir}
+          liveText={liveLabel}
```
DenisHdz, line 166: Technically, we could use the default prop for `liveText` when is in English.

tochwill: I think I'd rather just pass it in every time - it makes it a bit less complex at this level!

DenisHdz (approved):
LGTM 👍

sareh (approved):
Looks good. 
Screenshot live

Screenshot next with styles

I checked with CSS turned off for this and the spacing was correct too. Would need to review the offscreen text with UX. But happy to get this merged in and review it on the test environment. 

Screenshot next without styles

packages/components/psammead-radio-schedule/src/ProgramCard/__snapshots__/index.test.jsx.snap
```diff
@@ -226,22 +212,23 @@ exports[`ProgramCard should render correctly for live 1`] = `
...
           <span
-            aria-hidden="true"
-            class="c4"
-          >
-            LIVE
-          </span>
-          <span
-            class="c5"
-            lang="en-GB"
+            role="text"
```
greenc05, line 215: not sure we need 2 x role text??

PR #3294, opened 2020-03-24
Script link - remove wrapping span

DenisHdz:
Partially resolves https://github.com/bbc/psammead/issues/3290

**Overall change:**
Remove `span` from the displayed text.

**Code changes:**
- Remove surrounding `span` from children
- Remove visually hidden text from the story
- Update README

---

- [x] I have assigned myself to this PR and the corresponding issues
- [x] Automated jest tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

Issue #3290, opened 2020-03-23
Script link - Remove visually hidden text


greenc05:
**Describe the bug**
The script link has visually hidden text with english translations on:

https://www.bbc.com/serbian/lat 
https://www.bbc.com/serbian/cyr 
https://www.bbc.com/ukchina/simp 
https://www.bbc.com/ukchina/trad

**To Reproduce**
Steps to reproduce the behaviour:

1. Go to https://www.bbc.com/serbian/cyr
2. Inspect the script link
3. You will see ```<span class="VisuallyHiddenText-sc-1yjwwa5-0 jFfEvX"> Cyrillic</span>```

**Expected behaviour**
The script link should not have visually displayed text as per the UX screen reader spec. This also means that the displayed text should not have an aria hidden attribute, this will also need removing ```<span aria-hidden="true">Ћир</span>```

**Screenshots**
<img width="1220" alt="Screen Shot 2020-03-23 at 11 44 16" src="https://user-images.githubusercontent.com/1061401/77313978-bd32c180-6cfc-11ea-999b-8210e4cfb5b7.png">

**Testing notes**

- [x] This bug fix is expected to need manual testing. When using AT, the script link should just read out the displayed text.

sareh:
@greenc05 So just to clarify:
Before
```
<a href="/serbian/cyr" data-variant="cyr">
  <span>
    <span aria-hidden="true">Ћир</span>
    <span class="VisuallyHiddenText"> Cyrillic</span>
  </span>
</a>
```
After:
```
<a href="/serbian/cyr" data-variant="cyr">Ћир</a>
```

And these changes should be seen on serbian lat/cyr, and ukchina and zhongwen simp/trad too.

sareh:
The Psammead component needs to have its Readme updated, Storybook file and Test file updated as well as not surrounding the index in a `<span></span>` Then in Simorgh we can make sure we are not passing in the visually hidden text component.

PR #3299, opened 2020-03-25
Script Link - add wrapping span

DenisHdz:
Resolves #NUMBER

**Overall change:** 
In [this](https://github.com/bbc/psammead/pull/3294) PR we've removed the `span` that wrapped in the children from the `ScriptLink` as a consequence of [this](https://github.com/bbc/psammead/issues/3290) issue. However, the styles were applied to that `span` and in order to make it look as it was, it will require time to try to apply them directly to the `StyleLink`. For this reason, 
we've decided to put the `span` back.

Still, we'll make sure not to pass the `visually hidden text` and the `<span aria-hidden="true">` as children from Simorgh as done in [here](https://github.com/bbc/simorgh/pull/5962).

**Code changes:**
- Wrap children in a styled `span`
- Use a `StyledSpan` instead of using `> span`
- Remove repated `padding: 0 ${GEL_SPACING};` above `599px`

---

### DIFF
--- a/packages/components/psammead-script-link/CHANGELOG.md
+++ b/packages/components/psammead-script-link/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 1.0.13 | [PR#3294](https://github.com/bbc/psammead/pull/3294) Wrap children in a styled `span` |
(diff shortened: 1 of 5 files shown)


tochwill (approved):

Bopchy (approved):
Looks good 👍

sareh (approved):
:+1: Looks good. Thanks for updating to use styled-components.

PR #3311, opened 2020-03-26
Add SSR support for components using Loadable

simonsinclair:
No issue.

### Overall change:
Components, like `psammead-social-embed`, that use Loadable to dynamically load dependencies are failing SSR within Simorgh, because they haven't been [transpiled using `@loadable/babel-plugin`](https://loadable-components.com/docs/babel-plugin/#transformation). 

### Code changes:

- Install `@loadable/babel-plugin`
- Bump Psammead and package versions
- Add changelog entries

---

- ~~[ ] This PR requires manual testing~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 2.0.112 | [PR#3311](https://github.com/bbc/psammead/pull/3311) Install @loadable/babel-plugin to support components using Loadable. |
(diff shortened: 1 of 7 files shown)


andrewscfc (approved):

jamesdonoh (approved):
The general approach makes sense, but do we know which components should be affected by this change? It occurs to me that it could have quite far-reaching effects that might not become apparent until each component is next built and released.

simonsinclair:
> The general approach makes sense, but do we know which components should be affected by this change? It occurs to me that it could have quite far-reaching effects that might not become apparent until each component is next built and released.

This change only affects components that import Loadable. The additional code generated in transpilation is required by SSR in Simorgh and is [backwards compatible](https://loadable-components.com/docs/babel-plugin/#transformation) with the client.

PR #3321, opened 2020-03-30
Fix radio schedule duration

### LINKED ISSUE #3303: Radio Schedules - Duration markup
**Describe the bug**
Currently duration is read out twice - `datetime="PT29M30S"` 29 mintues, 30 seconds and the visually hidden text `Duration 29,30 `

**To Reproduce**

1. Go to `www.test.bbc.com/hausa?renderer_env=live`
2. Use VoiceOver on iOS (iPhone) to navigate to the Radio Schedules component
3. Hear duration read out twice. 

**Expected behaviour**

**Screenshots**

- OS: Mac version 10.13.6
- Browser Safari version 12.1
- Version 12.1

**Testing notes**

- [x] This bug fix is expected to need manual testing with all AT.

**Additional context**
We'd like to simplify this as much as possible without compromising accessibility for screen reader users. 
e.g we should test a simpler layout with all our supported Assistive Tech devices. 
```
<time dir="ltr" datetime="PT29M30S"><span class="VisuallyHiddenText-sc-1yjwwa5-0 jFfEvX"> Duration 29,30 </span><span aria-hidden="true" class="DurationTextWrapper-sc-6oo8lp-10 dfaGSJ">29:30</span></time>

### PR DESCRIPTION
**Pull request by thekp**
Resolves #3303

**Overall change:** _Fixed duration being read out twice on iphone voice over._

**Code changes:**

- _Updated the visually hidden text to be in a single string, as previously it was causing the user to have to swipe right twice on iphone voice over in order to hear the duration text._

---

### DIFF
--- a/packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
+++ b/packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
@@ -249,9 +249,13 @@ const ProgramCard = ({
       <IconWrapper {...programStateConfig[state]}>
         {mediaIcons.audio}
       </IconWrapper>
-
       <DurationWrapper dir={dir} dateTime={duration}>
-        <VisuallyHiddenText>{` ${durationLabel} `}</VisuallyHiddenText>
+        <VisuallyHiddenText>
+          {` ${durationLabel} ${formatDuration(
+            duration,
+            getDurationFormat(duration, ','),
+          )} `}
+        </VisuallyHiddenText>
         <DurationTextWrapper>
           {formatDuration(duration, getDurationFormat(duration))}
         </DurationTextWrapper>
--- a/packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
+++ b/packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
@@ -249,11 +249,9 @@ const ProgramCard = ({
       <IconWrapper {...programStateConfig[state]}>
         {mediaIcons.audio}
       </IconWrapper>
+
       <DurationWrapper dir={dir} dateTime={duration}>
-        <VisuallyHiddenText>
-          {` ${durationLabel} `}
-          {`${formatDuration(duration, getDurationFormat(duration, ','))} `}
-        </VisuallyHiddenText>
+        <VisuallyHiddenText>{` ${durationLabel} `}</VisuallyHiddenText>
         <DurationTextWrapper>
           {formatDuration(duration, getDurationFormat(duration))}
         </DurationTextWrapper>
--- a/packages/components/psammead-radio-schedule/CHANGELOG.md
+++ b/packages/components/psammead-radio-schedule/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx [hunk 1]
```diff
@@ -249,11 +249,9 @@ const ProgramCard = ({
       <IconWrapper {...programStateConfig[state]}>
         {mediaIcons.audio}
       </IconWrapper>
+
       <DurationWrapper dir={dir} dateTime={duration}>
-        <VisuallyHiddenText>
-          {` ${durationLabel} `}
-          {`${formatDuration(duration, getDurationFormat(duration, ','))} `}
```
DenisHdz, line 257: @greenc05 Do you remember by any chance if we had to add the duration within the visually hidden text?

greenc05: Yes we need that, as it's in a different format with the comma, like our media durations, as that is less likely to read out as a time format e.g. 2 o'clock

greenc05:
Did you investigate any other solutions @thekp @itimirichard

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
DenisHdz: In that case, shall we apply `aria-hidden` to the visual duration instead, so it is not read out twice?

greenc05: I thought that's what we had already done?

greenc05: I actually think this bug is being caused by some of our CSS

greenc05: As the bug isn't present wheh there is no CSS...

DenisHdz: True, the `DurationTextWrapper` has it already.
`const DurationTextWrapper = styled.span.attrs({ 'aria-hidden': 'true' })``;`

itimirichard: Yes we already have that @DenisHdz

itimirichard:
Yes @greenc05 , we tried replacing the `time` element with a span but that didn't work, the duration was still read out twice and also it was read out as number and not time (so for `PT30M`, instead of `thirty minutes` being read out, `three zero, zero zero` was read out.

greenc05:
The bug isn't present here http://2bitstudio.co.uk/code-tests/world-service/time-2.html Same HTML but the only CSS being used (I believe) is the VH CSS, so it must be being caused by some of the other CSS we have used in the component... Note that the duration will read twice due to the time tag when swiping left to right once, but this is a different bug we are investigating... On test you have to do 2 left to right swipes, and a duration is read out for each swipe, this should only be one swipe as per my test page above...

greenc05:
Great work, thanks @thekp @itimirichard

EinsteinNjoroge (approved):
Looks good

DenisHdz (approved):
Tested it with Talkback on the Android phone and the duration is read out just once. 
Thanks for the fix 👍

PriyaKR:
LGTM tested voiceover on ios and mac os.

PR #3349, opened 2020-04-08
Social Embed - resolve a11y issues and release

### LINKED ISSUE #3341: Resolve Social Embed a11y issues and remove alpha tag
### Is your feature request related to a problem? Please describe.

A11y
- [x] p1. EC - missing skip link and end message on the no response message/fallback - TBC if this is a requirement for v.1
- [x] p2. CG - External labelling, should use an aria-label rather than VH text - this is what we do for inline links, because if CSS is off then 'external' looks odd/might not make sense/alter the editorial meaning etc. It would be good to keep our approach consistent...

UX
- [x] p2. NH - Boarder around the focus state for the skip link should be Ebon: #222222. CG - Though we keep the default outline please e.g. blue line in Chrome
- [x] p2. DO - Social embed fallback - Link states are incorrect (currently browser default, should be https://zpl.io/awewv7j)

Source: https://docs.google.com/spreadsheets/d/1pdQt4kE9DsJTBN7BZrNa-0LP-CnK5-qfzKvJ5gHhLTw/edit#gid=0

**Testing notes**
[Tester to complete]

**Additional context**

simonsinclair:
Resolves https://github.com/bbc/psammead/issues/3341.

### Overall change
Following an [a11y swarm](https://docs.google.com/spreadsheets/d/1pdQt4kE9DsJTBN7BZrNa-0LP-CnK5-qfzKvJ5gHhLTw/edit#gid=0&range=H30) on 06/04/20, prepare Social Embed for release by resolving [a11y & UX issues](https://github.com/bbc/psammead/issues/3341).

### Follow-up a11y swarm
- [x] In **VoiceOver iOS** on the fallback, the link is two swipes rather than one when navigating via links.
- [x] In **VoiceOver Mac OS** on the fallback, 'external' is read out before the link text. This should be last.

### Code changes
- Resolve a11y & UX issues
- Bump package for release
- Update snapshots
- Add changelog entry

---

- ~~[ ] This PR requires manual testing~~


packages/components/psammead-social-embed/src/Notice/index.jsx
```diff
@@ -36,8 +32,12 @@ const Wrapper = styled.div`
     ${({ service }) => getSansBold(service)}
     text-decoration: none;
 
-    > span {
-      ${visuallyHiddenStyle}
+    &:visited {
+      color: ${C_METAL};
+    }
+
+    &:hover {
```
greenc05: focus too please?

packages/components/psammead-social-embed/src/Notice/index.jsx
```diff
@@ -57,14 +57,22 @@ const Notice = ({
...
-      <a href={linkHref}>
-        {detokenise(linkText, dictionary)}
-        {linkTextSuffixVisuallyHidden && (
-          <span>{detokenise(linkTextSuffixVisuallyHidden, dictionary)}</span>
-        )}
+      <a
+        href={linkHref}
+        aria-label={
+          linkTextSuffixVisuallyHidden &&
+          `${detokenisedLinkText}${detokenise(
```
chris-hinds: is it worth pulling this out to its own variable like `detokenisedLinkText ` on L60 ? I think it would improve the readability a bit.

rhenshaw56 (approved):

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
```diff
@@ -2,6 +2,54 @@
...
+  padding: 0.75rem;
+  position: absolute;
+  -webkit-text-decoration: none;
+  text-decoration: none;
+  top: 0;
+  z-index: 10;
+}
+
+.c1:not(:focus):not(:active) {
+  position: absolute !important;
```
greenc05: do we defiantly need that !important?

packages/components/psammead-social-embed/src/Notice/index.jsx
simonsinclair: We use the default focus state. Please refer to Storybook.

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
greenc05: and the one below? What is this CSS and .c3 for?

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
```diff
@@ -2,6 +2,54 @@
...
+  color: #222222;
+  display: block;
+  left: 0;
+  line-height: 1;
+  padding: 0.75rem;
+  position: absolute;
+  -webkit-text-decoration: none;
+  text-decoration: none;
+  top: 0;
+  z-index: 10;
```
greenc05, line 25: Is this a random number? why 10?

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
```diff
@@ -59,17 +116,34 @@ exports[`AmpSocialEmbed should render a notice when the provider is unsupported
...
-  <p>
-    Sorry but we're having trouble displaying this content
-  </p>
   <a
-    href="embed-url"
+    class="c1"
+    href="#skip-unknown-content"
   >
-    View content on unknown
+    Skip unknown content
```
greenc05, line 125: Is this the text we are actually using? (Translated)

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
```diff
@@ -59,17 +116,34 @@ exports[`AmpSocialEmbed should render a notice when the provider is unsupported
...
+    <small>
+      Warning: BBC is not responsible for third party content
+    </small>
+  </div>
+  <p
+    class="c3"
+    id="skip-unknown-content"
+    tabindex="-1"
+  >
+    End of unknown content
```
greenc05, line 147: Is this the text we are actually using? (Translated)

packages/components/psammead-social-embed/src/Notice/index.jsx
simonsinclair: Yes, good suggestion!

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
simonsinclair: No, this is test data. Please refer to Storybook.

packages/components/psammead-social-embed/src/Notice/index.jsx
greenc05: the default outline yes, but we would also like text underline too please (as the default outline doesn't always pass contrast depending on the browser)

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
simonsinclair: As above.

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
simonsinclair: z-index values tend to be arbitrary. Here we're ensuring the skip link always sits above the oEmbed.

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
simonsinclair: Yes, `!important` is... important here to ensure this rule is _never_ overridden.

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
greenc05: Not sure how to see that on storybook now? Could you give me an example please? "Skip Twitter content"?

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
greenc05: Could you give me an example please? "End of Twitter content"?

greenc05:
This will need another dev a11y swarm testing specifically what we've changed with all our supported AT (if you haven't already done that) before the alpha label is removed please @simonsinclair

12 (dismissed):
LGTM, though needs an a11y swarm.

packages/components/psammead-social-embed/src/Notice/index.jsx
simonsinclair: `detokenise(text, dictionary)` expects `text` to be a string, and because `linkTextSuffixVisuallyHidden` is an optional prop, using a local variable might result in less readable code. Which do you prefer?

```js
const detokenisedLinkTextSuffix =
  linkTextSuffixVisuallyHidden &&
  detokenise(linkTextSuffixVisuallyHidden, dictionary);
// ...
aria-label={
  detokenisedLinkTextSuffix &&
  `${detokenisedLinkText}${detokenisedLinkTextSuffix}`
}
```

```js
aria-label={
  linkTextSuffixVisuallyHidden &&
  `${detokenisedLinkText}${detokenise(
    linkTextSuffixVisuallyHidden,
    dictionary,
  )}`
}
```

chris-hinds: Personally I prefer Option 1 :D. However this is not about preference, what you have works and works well! Let's merge this.

chris-hinds (approved):
👍

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
greenc05: @lizcameron has kindly reminded me what this will say on the fallback, same as the enriched embed - Thanks both

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
greenc05: @lizcameron has kindly reminded me what this will say on the fallback, same as the enriched embed - Thanks both

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
greenc05: Yes !important will do that and that is the reason for my question, as users should be able to override all our styles if they so wish to load their own CSS... So I'm wondering exactly why it is that we need to apply that here? Why might it be overridden? I believe the CSS still works without that looking on storybook at the example?

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
greenc05: Ok, was just checking... thanks

packages/components/psammead-social-embed/src/index.jsx
```diff
@@ -23,23 +23,25 @@ export const CanonicalSocialEmbed = ({
...
+    if (isSupportedProvider && oEmbed)
+      return <CanonicalEmbed provider={provider} oEmbed={oEmbed} />;
 
-  return <Notice service={service} provider={provider} {...fallback} />;
+    return <Notice service={service} provider={provider} {...fallback} />;
+  };
+
+  return (
+    <SkipLinkWrapper service={service} provider={provider} {...skipLink}>
+      {renderCanonicalContent()}
```
andrewscfc: Why do we need to declare a function and call it here rather than just rendering in the `<SkipLinkWrapper>` tag?

packages/components/psammead-social-embed/src/__snapshots__/index.test.jsx.snap
greenc05: If the .c1 and .c3 CSS declarations are to achieve visually hidden CSS, if it's helpful this is the CSS that we usually use for that: https://github.com/bbc/psammead/blob/797d78a9f289d4e09daaabd1039566f0fa7a7666/packages/components/psammead-visually-hidden-text/src/index.jsx#L4

simonsinclair: I've refactored this to make use of `psammead-visually-hidden-text`'s styles.

packages/components/psammead-social-embed/src/index.jsx
simonsinclair: This was keeping semi-complex control logic out of JSX to improve readability, but it obviously wasn't helping, so I've refactored it. [Hopefully this makes more sense](https://github.com/bbc/psammead/blob/06c32718ad201f7943bb32dae4452d6b8ba2830d/packages/components/psammead-social-embed/src/index.jsx#L61-L78).

packages/components/psammead-social-embed/src/utilities/index.js
```diff
@@ -37,15 +34,15 @@ export const dictionaryFactory = ({ provider }) => ({
 });
 
 /**
- * We require control over _when_ these styles are applied and
- * therefore cannot use @bbc/psammead-visually-hidden-text.
+ * These styles are lifted from @bbc/psammead-visually-hidden-text, which we
+ * cannot use directly as we require control over _when_ they are applied.
```
andrewscfc, line 38: Could we export the style string from visually hidden text and import it here: https://github.com/bbc/psammead/blob/latest/packages/components/psammead-visually-hidden-text/src/index.jsx#L3..L11 ? Just pretty much certain if visually hidden text changes this won't

andrewscfc (approved):
Just one comment relating to DRY, I don't want to hold this up for that, just might be a bit nicer. Maybe a follow-up PR?

packages/components/psammead-social-embed/src/utilities/index.js
simonsinclair: Yeah, I believe we discussed updating `@bbc/psammead-visually-hidden-text` to cater for more use cases in the swarm review. I've raised [an issue here](https://github.com/bbc/psammead/issues/3361).

chris-hinds (approved):

PR #3393, opened 2020-04-17
Add a detokeniser utility to Psammead

### LINKED ISSUE #3375: Pass radio schedule 'duration' translation in as top level prop to component, instead of adding it to the programme data.
Currently, the translation for `duration` is passed into the radio schedules component as part of the programme information (https://github.com/bbc/psammead/blob/latest/packages/components/psammead-radio-schedule/src/index.jsx#L54). This should instead be passed in as a top level prop, like `liveLabel` and `nextLabel`. This will allow us to remove adding 'Duration' in when we format the radio schedules data in simorgh, where it is currently being set to the hard-coded English string `'Duration'`.

This duration translation will be passed in from Simorgh in a format that will allow us to interpolate the duration into the translation (for example, `Duration, %duration%` in english`). This is currently being done in social embeds here: https://github.com/bbc/psammead/blob/latest/packages/components/psammead-social-embed/src/utilities/index.js#L21-L34 - we should decide on where this should be done now (for example a psammead utility) that we are doing this in multiple places outside of social embeds, and then use this approach to handle our case.

```
‘%duration%’: formatDuration(duration, getDurationFormat(duration))
```

We should also update the component in psammead to handle this interpolation and update usage examples in storybook.

When picking this up, we can have another discussion to answer remaining questions on the format.

Accept `durationLabel` as a prop at the top level of the radio schedules component, instead of destructuring it from the programme data.

**Testing notes**
[Tester to complete]
Ensure this is covered by assertion based unit tests.

**Additional context**

OlgaLyubin:
Partially resolves #3375

**Overall change:** 
Added a `psammead-detokeniser` utility that returns a string that has had its tokens replaced.

**Code changes:**
- Add a `psammead-detokeniser` utility
- Add tests

---


packages/utilities/psammead-detokeniser/README.md
```diff
@@ -0,0 +1,54 @@
...
+| --------- | ---- | -------- | ------- | ------- |
+| text | string | yes | N/A | `'Foo %token%'` |
+| dictionary | object | yes | N/A | { `'%token%': 'Bar'` } |
+
+## Usage
+
+<!-- Description of the utility usage -->
+
+```
+import Detokeniser from "@bbc/psammead-detokeniser"
```
sareh:
```suggestion
import detokenise from "@bbc/psammead-detokeniser"
```

DenisHdz: Can we have an example of how this is gonna be used?

packages/utilities/psammead-detokeniser/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Description
+
+The `Detokeniser` utility replaces % delimiter with spaces to be fed into a dictionary.
```
simonsinclair: - What do we mean by "spaces" here? 
- Can we rephrase "% delimiter" with "% delimited tokens"?
- Can I suggest we also mention that this function operates on strings?

packages/utilities/psammead-detokeniser/README.md
```diff
@@ -0,0 +1,54 @@
...
+```jsx
+npm install @bbc/psammead-detokeniser --save
+```
+
+## Props
+
+| Argument  | Type | Required | Default | Example |
+| --------- | ---- | -------- | ------- | ------- |
+| text | string | yes | N/A | `'Foo %token%'` |
+| dictionary | object | yes | N/A | { `'%token%': 'Bar'` } |
```
simonsinclair, line 18: Can I suggest we have a section on Dictionaries here, which explains how they're structured and ultimately used with this function? A 'real world' example could be helpful.

packages/utilities/psammead-detokeniser/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@bbc/psammead-detokeniser",
+  "version": "1.0.0",
+  "main": "dist/index.js",
+  "module": "esm/index.js",
+  "sideEffects": false,
+  "description": "Replaces % delimiter with spaces to be fed into a dictionary",
```
simonsinclair: As a reminder, if we tweak the readme description, this should follow.

packages/utilities/psammead-detokeniser/src/index.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Returns a string that has had its tokens replaced.
+ * @param {String} text A word or words containing tokens.
+ * @param {Object} dictionary An object which maps keys as tokens to values.
+ */
```
simonsinclair, line 5: Could this help us construct a more accurate description of this utility?

simonsinclair (commented):
It's super to see this being used elsewhere. Thanks for migrating it.
I've left a couple of questions below.

packages/utilities/psammead-detokeniser/README.md
sareh: It'd be good to add a usage example. 

Looking at the usage in psammead-social-embed [here in the notice](https://github.com/bbc/psammead/blob/0af9a91f48f9da226b892e44d1f9239c5c38653f/packages/components/psammead-social-embed/src/Notice/index.jsx#L61-L65) and [here in the skip link](https://github.com/bbc/psammead/blob/be3d50ed96f9b315d87f7bc5caedc08cbfecb6b5/packages/components/psammead-social-embed/src/SkipLinkWrapper/index.jsx#L53-L64), it seems like we need to use both the detokenise and the dictionaryFactory. Can you please add `dictionaryFactory` as an export in this package?

simonsinclair: I know it's not immediately obvious, but it might be worth noting that `dictionaryFactory` is specific to the Social Embed implementation and not something that needs to be migrated.

N.B. When called with `provider`, `dictionaryFactory` constructs an object that is used as a dictionary for Social Embeds. This is an example of a 'dynamic dictionary' - when your definitions can change depending on the value of something else. Other implementations will return different dictionaries or may simply use an object literal (or static dictionary). If further clarity is needed here, I'd be happy to answer questions in a call.

OlgaLyubin: @sareh Thank you for reviewing 👍 We did not include a `dictionaryFactory` in this package as we thought that one for `social-embeds` and `radio-schedules` will be different. 
e.g. in `social-embeds` `dictionaryFactory` uses a `getProviderName` function so in order to relocate `dictionaryFactory` into `psammead-detokeniser` package we would need to include that function as well + make a dictionary factory version for radio schedules
Is this something you think we should do or should we have separate versions of `dictionaryFactory` in social embeds and radio schedule?

simonsinclair (approved):
I look forward to using this in `psammead-social-embed`.

j-pendlebury (approved):
LGTM 👍

itimirichard (approved):

PriyaKR:
LGTM.

packages/utilities/psammead-detokeniser/README.md
sareh: @simonsinclair @OlgaLyubin Thank you both for explaining. Sorry I missed your comments on this yesterday. Now that I understand the usage of this better, I agree that this way you've set it up is exactly what we need. Thanks for the PR & for reviewing it.

PR #3403, opened 2020-04-21
Update timestamp utilities - localisation

### LINKED ISSUE #3308: Update all timestamps to use service numerals
In the radio schedules component, we are using the numerals as defined by the service:

Screenshot 2020-03-25 at 15 48 07

In media content, we are using western arabic numerals everywhere:

Screenshot 2020-03-25 at 15 50 23

We have spoken within the team, and agree that we should be consistent. Within our team, and after conversations with Jon and Jim, we would prefer to use the service numerals. We need to discuss this with other teams, and ensure that we keep this consistent across Simorgh.

Decide which numerals to use, and then use them.

**Testing notes**
[Tester to complete]

**Additional context**

### LINKED ISSUE #3373: Radio Schedule duration is formatted differently on server render vs client render
**Describe the bug**
The server rendered duration in the radio schedule section does not respect the service locale, whereas the client rendered duration does. As we do not currently set the locale when formatting the duration timestamp, this is probably due to some other code running on the client updating the global moment locale.

We can update the `formatDuration` function in `psammead-timestamp-container` to accept the locale as an argument, and then ensure the local moment locale is set:

```
return moment.utc(durationInMilliseconds).format(format);
```

becomes
```
return moment.utc(durationInMilliseconds).locale(locale).format(format)
```

Check how this affects the work being done as part of https://github.com/bbc/psammead/issues/3358.

Another point to note is that the `,` delimiter that has been used to format the visually hidden duration is not respected for all locales - it is replaced with a `.` for example on the afrique site, or `،` for persian. We can check this for all services - `,` may not be a valid character for this script so we should be careful about forcing this behaviour, and let moment format these in a way that is better suited (we can check the screen reader output). Is a separate visually hidden timestamp the right approach?

**To Reproduce**

1. Go to the persian homepage: localhost:7080/persian
2. View radio schedules section
4. You will see the numerals rendered with western arabic numerals on the server, which then change to the service numerals on the client.

**Expected behaviour**
The server render of the duration timestamp should respect the locale settings, and match the client render.

**Screenshots**

**Testing notes**

**Additional context**

sareh:
Resolves #3373 
Part of #3308 

**Overall change:** Breaking change in utilities
Change function **formatDuration** to take in locale and destructure arguments
old usage: `formatDuration('PTH1M30', 'mm,ss')`
new usage: `formatDuration({ duration: 'PTH1M30', format: 'mm,ss', locale: 'my' })`

For Bengali, Burmese, Nepali, Pashto, Persian we need to use the service numerals, which are defined in the locale definition. By passing in the locale to this function we should have the correct numerals.

Change function **unixTimestampToMoment** and replace with **localisedMoment** which takes in locale
old usage: `unixTimestampToMoment(timestamp)` 
new usage: `localisedMoment({ locale, timestamp })`

This is used in the Timestamp container for services that need alt calendars (e.g. Pashto and Persian) - ensuring they're using localised numerals that have been defined in the locales. 

Change function **formatUnixTimestamp** to destructure arguments
old usage: `formatUnixTimestamp(1483275600000, 'D MMMM YYYY, HH:mm z', 'GMT', 'fr', false);`
new usage: 
```
formatUnixTimestamp({
  format: 'D MMMM YYYY, HH:mm z',
  isRelative: false,
  locale: 'fr',
  timestamp: 1483275600000,
  timezone: 'GMT',
});
```

Destructuring is useful for reasons of easy readability, also it makes it easier to extend a function in future, by allowing some arguments to have default values.

**Code changes:**
*formatDuration*
- Add usage of `formatDuration` to Readme.
- Pass in locale so we can use this for services that need localisation.
- Update arguments to take an object instead of a list of params (this allows for some arguments to be optionally used)
- Update default format so that it handles hours.

*formatUnixTimestamp*
- Updated to destructure arguments - it makes the function much easier to use/read in code when used elsewhere. 

*localisedMoment*
- renamed from `unixTimestampToMoment` since it needed to be localised too.
- using this for the alt calendar, so that when we use alternative calendars, they're localised.

---


tochwill (approved):

EinsteinNjoroge (approved):

OlgaLyubin (approved):
Looks great 👍

PR #3408, opened 2020-04-22
Update durationLabel usage with detokeniser

### LINKED ISSUE #3375: Pass radio schedule 'duration' translation in as top level prop to component, instead of adding it to the programme data.

hotinglok:
Resolves #3375

**Overall change:** 
Made `durationLabel` a top level prop as well as implemented the detokeniser to be used when we get the translations in.

**Code changes:**
- Passed in radio schedule 'durationLabel' as top level prop to component, instead of adding it to the `programCard` data
- Created a `durationDictionary` function that generates a dictionary object to be passed into a `detokeniser`
- Added tests for `durationDictionary` and `detokeniser`
- Used `detokeniser` in order to make up a duration label within the `VisuallyHiddenText`
- Moved `getDurationFormat` into the utilities folder
- Added `@bbc/psammead-detokeniser` to `devDependencies`

**Additional Context**
- Changes were made to `./ProgramCard/index.jsx` and `./src/index.jsx` to have the same output as we currently do but with the detokeniser added. This doesn't actually have to be done now so it's up to what people think about whether or not this should be done in this PR too since it's essentially just a placeholder at the moment.
- `formatDuration` has some error handling because of moment where as long as a string is passed in, it will fallback to '00:00'. However,  this does not catch any non valid inputs such as null/undefined, would just like some opinions on whether or not it would be worth adding anything to deal with this. Would we ever be in a situation where we either don't get the `duration` prop or the prop is somehow severely in the wrong format?
---


packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -275,11 +269,10 @@ const ProgramCard = ({
       </IconWrapper>
       <DurationWrapper dir={dir} dateTime={duration}>
         <VisuallyHiddenText>
-          {` ${durationLabel} ${formatDuration({
-            duration,
-            format: getDurationFormat(duration, ','),
-            locale,
-          })} `}
+          {`${detokenise(
+            ` ${durationLabel} %duration%`,
```
tochwill: Could you pass in the duration label so it is the whole string, including the token? Different languages will have the token in different places :)

packages/components/psammead-radio-schedule/src/utilities/index.js
```diff
@@ -0,0 +1,20 @@
...
+    timeSections.unshift('h');
+  }
+  return timeSections.join(separator);
+};
+
+const durationDictionary = ({ duration, separator, locale }) => ({
+  '%duration%': formatDuration({
+    duration,
+    format:
+      getDurationFormat(duration, separator) || getDurationFormat(duration),
```
tochwill: Howcome is this function repeated again after an `||` operator?

packages/components/psammead-radio-schedule/CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
-| 2.0.0 | [PR#3377](https://github.com/bbc/psammead/pull/3377) Remove episodeTitle prop. In ProgramCard, generate episodeTitle with moment using startTime. |
+| 2.0.1 | [PR#3408](https://github.com/bbc/psammead/pull/3408) Use detokenizer to construct duration label |
```
tochwill: As we have a new required prop, I think we need to make this a major version bump - if it gets pulled in automatically it's not going to work as intended :)

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
OlgaLyubin: Yes, sorry, this was in place just so that duration would work with what we have in Simorgh now (since we don't have proper translations with %duration% yet). But instead we can change this line + make this pr a major change + only pull it into Simorgh when translations are ready :)

tochwill: Thanks! I think what we might have to do is update the translations in Simorgh to include tokens based on what we already have, and then update them with the token in the correct place once we get the translations back.

packages/components/psammead-radio-schedule/README.md
```diff
@@ -28,6 +28,7 @@ npm install @bbc/psammead-radio-schedule --save
 | script | object | yes | N/A | `{ canon: { groupA: { fontSize: '28', lineHeight: '32',}, groupB: { fontSize: '32', lineHeight: '36', }, groupD: { fontSize: '44', lineHeight: '48', }, }, trafalgar: { groupA: { fontSize: '20', lineHeight: '24', }, groupB: { fontSize: '24', lineHeight: '28', }, groupD: { fontSize: '32', lineHeight: '36', }, }, }` |
 | locale | string | no | N/A | `'en-gb'` |
 | timezone | string | no | N/A | `'Europe/London'` |
+| durationLabel | string | yes | N/A | `'duration'` |
```
tochwill: Could you update this example to include a token?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
```diff
@@ -275,11 +269,10 @@ const ProgramCard = ({
       </IconWrapper>
       <DurationWrapper dir={dir} dateTime={duration}>
         <VisuallyHiddenText>
-          {` ${durationLabel} ${formatDuration({
-            duration,
-            format: getDurationFormat(duration, ','),
-            locale,
-          })} `}
+          {`${detokenise(
+            ` ${durationLabel}`,
```
tochwill: Do you need both of these nested interpolated strings? Can you simplify this so there is just one?

packages/components/psammead-radio-schedule/src/utilities/index.test.js
```diff
@@ -0,0 +1,32 @@
...
+  });
+  it('creates a valid duration dictionary and maps token to key correctly', () => {
+    expect(
+      detokenise(
+        'Duration %duration%',
+        durationDictionary({ duration: 'PT1H30M', locale: 'en-gb' }),
+      ),
+    ).toEqual('Duration 1,30,00');
+  });
+  it('creates a valid duration dictionary and returns with specified duration format', () => {
```
tochwill: What happens if duration and locale are invalid or missing?

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
hotinglok: This is to handle the fact that we need a space before `Duration: %duration%`. This is needed when CSS and Javascript are disabled:
image
If we didn't do this, I'm not sure how we would add the space in. If we don't then I think we might have a UX problem:
image

packages/components/psammead-radio-schedule/src/utilities/index.test.js
hotinglok: If duration is missing then I'd assume the fallback that radio schedules just wouldn't show would kick in before this fails. Are there any situations where we would get duration missing entirely? 
If locale was invalid or missing, again I'd assume other fallbacks at a higher level would come in first. Perhaps this needs further discussion elsewhere.

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
tochwill: Could you change the `  ${durationLabel}` to just be durationLabel, and add the extra space before the detokenise function? Then you only need one interpolation

packages/components/psammead-radio-schedule/src/utilities/index.test.js
tochwill: We're discussing this at 2:30 today, we may want to render the card anyway even if we don't get back the duration. So we'll just have to make sure we're not trying to format it if it doesn't exist - let's make this decision based on the outcome of that meeting :)

packages/components/psammead-radio-schedule/src/ProgramCard/index.jsx
hotinglok: Ah, nice catch. Completely missed that it was all already in a template literal. Fixed now, thanks.

packages/components/psammead-radio-schedule/src/utilities/index.test.js
tochwill: There should be fallbacks at a higher level, yes! My only concern is I think you mentioned that this threw an error if the duration isn't passed, or is passed in an incorrect format. Could we get a test case to cover what happens in this situation, and then we can decide if that's the behaviour we want, and take steps to guard against it if needed :)

packages/components/psammead-radio-schedule/src/utilities/index.js
```diff
@@ -0,0 +1,19 @@
+import { formatDuration } from '@bbc/psammead-timestamp-container/utilities';
+
+const getDurationFormat = (duration = '', separator = ':') => {
```
tochwill: Thanks for all this! I just think we can tweak all of these default parameters a little - `getDurationFormat` is only used in the `durationDictionary` below, correct? I think we can remove these defaults from `getDurationFormat`, as they will never be used, as `durationDictionary` has its own, that will override these every time the function is called (so the `:` separator default will never be used).

tochwill (approved):

packages/components/psammead-radio-schedule/src/utilities/index.test.js
```diff
@@ -0,0 +1,48 @@
...
+        'Duration %duration%',
+        durationDictionary({
+          duration: 'PT1H30M',
+          separator: ':',
+          locale: 'en-gb',
+        }),
+      ),
+    ).toEqual('Duration 1:30:00');
+  });
+  it('creates a valid duration dictionary but has an invalid string passed in for duration', () => {
```
DenisHdz: Thanks for adding these tests. However, all of them start with `creates a valid duration dictionary`. I would suggest adding this as a new `describe` block and then leave/update the reaming text in each of them.

DenisHdz (approved):
Great work both 👍

PriyaKR:
Looks good to me.

PR #3304, opened 2020-03-25
STY: Fix UX for Most Read

12:
Related to bbc/simorgh#5948

**Overall change:** _A very high-level summary of easily-reproducible changes that can be understood by non-devs._

**Code changes:**

- _A bullet point list of key code changes that have been made._
- _When describing code changes, try to communicate **how** and **why** you implemented something a specific way, not just **what** has changed._

---

- [ ] I have assigned myself to this PR and the corresponding issues
- [ ] Automated jest tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

PR #3409, opened 2020-04-23
Adds option to change Typography for MostRead

### LINKED ISSUE #3320: Reduce MostRead font size on story pages
The font size in `psammead-most-read` is too big on STY pages. Reduce its size without impacting MostRead on the front page.

Reduce the font size by adding a new prop similar to https://github.com/bbc/psammead/pull/3304

- Numerals should always use GEL Trafalgar copy sizing; 
- Promo text should always use GEL Pica copy sizing
[GEL typography](https://www.bbc.co.uk/gel/guidelines/typography)

We discussed making a new component but decided not to as the changes are quite minimal.
The decision was made after the investigation https://github.com/bbc/simorgh/issues/5948

**Testing notes**
- Snapshot updates
- Existing unit tests for most read need updating
- Check with UX

**Additional context**
Zeplin design exists - see JIRA NEWSART-936.

HarveyPeachey:
Closes https://github.com/bbc/psammead/issues/3320

**Overall change:** Adds the ability to change the size used for most read `Link` and `Rank`. This is done through a `size` prop which currently only supports `default` and `small`. For more context on the choice of solution see this [comment](https://github.com/bbc/psammead/pull/3409#issuecomment-620620435).

**Code changes:**

- Added an `size` prop for MostReadLink and MostReadRank
- Added extra `min-width` values for rank based on the size being used. This spacing is only for the `oneColumn` formatting at the moment.
- Updated stories in storybook.

---


packages/components/psammead-most-read/src/Item/index.jsx
```diff
@@ -58,14 +58,16 @@ const StyledLink = styled.a`
...
 
   @media (min-width: ${GEL_GROUP_2_SCREEN_WIDTH_MIN}) {
-    ${({ script }) => script && getGreatPrimer(script)};
+    ${({ script, typography }) =>
+      script && typography !== 'pica' && getGreatPrimer(script)}
   }
 `;
 
-const ALIGN_RANK_TO_LINK_SPACING = '0.375rem';
+const ALIGN_RANK_TO_LINK_SPACING = rankTypography =>
```
simonsinclair: This casing is typically used to denote a constant value. As it's now a function, it might be more appropriate to revert to camelCase and also rename it to something like `getItemSpacing`.

packages/components/psammead-most-read/src/Item/index.jsx
```diff
@@ -58,14 +58,16 @@ const StyledLink = styled.a`
...
+  rankTypography === 'trafalgar' ? '0.1rem' : '0.375rem';
 
 const StyledItem = styled.div`
-  padding-top: ${ALIGN_RANK_TO_LINK_SPACING};
+  padding-top: ${props => ALIGN_RANK_TO_LINK_SPACING(props.rankTypography)};
```
simonsinclair: We could improve the readability of this function slightly with a sprinkle of de-structuring, like you've used in your other functions:
```
padding-top: ${({ rankTypography }) => ALIGN_RANK_TO_LINK_SPACING(rankTypography)};
```

packages/components/psammead-most-read/src/utilities/index.jsx
```diff
@@ -40,6 +40,8 @@ export const getItemWrapperArray = ({
   dir,
   withTimestamp = false,
   columnLayout,
+  rankTypography,
+  linkTypography,
```
simonsinclair: When there can only be one of two different values for these props, is there any reason we didn't decide to use a flag to set this component up instead? E.G. `isCompact` or `size="compact"`. This would:
- Limit the complexity of this component both inside and out.
- Make it clear that this component doesn't support the full spectrum of typography classes.

packages/components/psammead-most-read/src/index.stories.jsx
```diff
@@ -18,6 +18,8 @@ const listIndexRange = {
 };
 
 const pageTypes = ['oneColumn', 'twoColumn', 'multiColumn'];
+const rankTypeSize = { default: 'foolscap', trafalgar: 'trafalgar' };
+const linkTypeSize = { default: 'greatprimer', pica: 'pica' };
```
simonsinclair: In the drop-down, this reads `default` and `trafalgar`. As `default` isn't a typography class, would it be more appropriate for it to read `foolscap` and `trafalgar`?

The use of `default` as opposed to typography classes here might support the idea we could merge `rankTypeSize` and `linkTypeSize` into a `size` prop that sets the component up internally – queried in https://github.com/bbc/psammead/pull/3409#discussion_r413784618.

simonsinclair (commented):
Hey @HarveyPeachey, great work, as usual. I've left a couple of comments below. Let me know if you'd like to chat about https://github.com/bbc/psammead/pull/3409#discussion_r413784618 in a call instead.

packages/components/psammead-most-read/src/utilities/index.jsx
HarveyPeachey: @simonsinclair 

It was considered on the original POC https://github.com/bbc/psammead/pull/3304. The consensus I got from the comments on the PR was that passing typography was favoured. Another reason the typography route was chosen, enables flexibility if you want to use larger typography for rank and a smaller one for the link and vice versa.

But I get what you mean as we've only got a few typographies supported atm, either approach is fine with me.

packages/components/psammead-most-read/src/Item/index.jsx
```diff
@@ -58,14 +58,16 @@ const StyledLink = styled.a`
   }
 
   @media (min-width: ${GEL_GROUP_2_SCREEN_WIDTH_MIN}) {
-    ${({ script }) => script && getGreatPrimer(script)};
-  }
+    ${({ script, typography }) =>
+      script && typography !== 'pica' && getGreatPrimer(script)}
+  }ß
```
thekp: 🙈 

```suggestion
  }
```

packages/components/psammead-most-read/src/utilities/index.jsx
simonsinclair: Thanks for explaining, Harvey. Having looked at the comments on that PR, I can see why you would've gone down this route, however I do wonder whether it was the name of the prop that prompted those comments as opposed to the design of the API. @sareh, would you be able to shed some light?

andrewscfc: I think I'd prefer an approach where there is a 'small' and 'large' configuration for most read as opposed to typography that can be configured in multiple ways. I also can't remember off the top of my head which `typography` value renders smaller or larger text, this didn't help me understanding the API. Apologies for not reviewing the original POC where I could have given that opinion earlier.

Just seems quite complex implementing all the combinations for the different typography configurations.

packages/components/psammead-most-read/README.md
```diff
@@ -118,6 +118,7 @@ import {
 | numberOfItems | number | yes | N/A | `10` |
 | dir | string | no | `"ltr"` | `"ltr"` |
 | columnLayout | string | no | 'multiColumn' | `twoColumn`
+| typography | string | no | 'foolscap' | `'trafalgar'` |
```
sareh:
```suggestion
| rankTypography | string | no | 'foolscap' | `'trafalgar'` |
```

packages/components/psammead-most-read/README.md
```diff
@@ -149,6 +150,8 @@ import { MostReadRank } from '@bbc/psammead-most-read';
 | title | string | yes | N/A | `"This is a article headline"` |
 | href | string | yes | N/A | `"/bbc.co.uk/news/00000027051997"` |
 | children | node | no | null | `<Timestamp datetime="2019-03-01T14:00+00:00" script={script} padding={false} service={service}>Last updated: 5th November 2016</Timestamp>` |
+| typography | string | no | 'greatprimer' | `'pica'` |
```
sareh: It would be clearer to call this linkTypography to be more specific with what this is overriding. 

```suggestion
| linkTypography | string | no | 'greatprimer' | `'pica'` |
```

packages/components/psammead-most-read/src/Item/index.jsx
```diff
@@ -58,14 +58,16 @@ const StyledLink = styled.a`
   }
 
   @media (min-width: ${GEL_GROUP_2_SCREEN_WIDTH_MIN}) {
-    ${({ script }) => script && getGreatPrimer(script)};
+    ${({ script, typography }) =>
+      script && typography !== 'pica' && getGreatPrimer(script)}
```
sareh: In the case where you have typography set to 'pica', you keep the media min-width block with no content. It'd be better to not have this block.

packages/components/psammead-most-read/src/Item/index.jsx
```diff
@@ -102,11 +111,15 @@ MostReadLink.propTypes = {
   title: string.isRequired,
   href: string.isRequired,
   children: node, // this node will be a timestamp container
+  typography: oneOf(['greatprimer', 'pica']),
+  rankTypography: oneOf(['foolscap', 'trafalgar']),
```
sareh: If you only want to allow these values, you should also update the documentation props table with a description to ensure that's understood.

packages/components/psammead-most-read/src/Rank/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { shape, string, oneOf, number } from 'prop-types';
 import styled from 'styled-components';
-import { getFoolscap } from '@bbc/gel-foundations/typography';
+import { getFoolscap, getTrafalgar } from '@bbc/gel-foundations/typography';
```
sareh, line 4: I have a concern about both typography sets being pulled into the component for every single version of MostRead.  However, we're already currently pulling in many different typographic values into our components in general, with MostRead being a specific example of this.

packages/components/psammead-most-read/src/utilities/index.jsx
sareh: My concern with something like 'small'/'large' or 'compact' is that in my experience working in our cross-disciplinary teams with re-usable components, there will always be further updates to typography, and as soon as you make something like this a boolean, there will be a third value and a 'smaller'/ 'larger'/ 'smallest' variant needed. For that reason, using GEL typography values would be more future-proof.

HarveyPeachey:
After a call with @simonsinclair discussing the different points made by a few different people on the PR, we decided to try and find a middle ground with the solution.

The solution consists of having a `size` prop which causes typographies to be set internally to reduce complexity with the number of typographies that exist and the small amount we're using to satisfy the requirements of the original issue. 

The `size` prop has been chosen to be a string which has been predefined through prop validation. Currently, the value of `size` can only be either `'default'` or `'small'`. Going down this route means new sizes can be added easily just by adding the new size to the prop validation and adding a check for the new prop to get the corresponding typographies, these are the locations where a check will need to be added:

https://github.com/bbc/psammead/blob/0db1f6a7e5572f2bc98b079bb7c250c0825eab4e/packages/components/psammead-most-read/src/Rank/index.jsx#L150-L153
https://github.com/bbc/psammead/blob/0db1f6a7e5572f2bc98b079bb7c250c0825eab4e/packages/components/psammead-most-read/src/Item/index.jsx#L60-L65

And adding rank spacing and link spacing for the new size:
https://github.com/bbc/psammead/blob/0db1f6a7e5572f2bc98b079bb7c250c0825eab4e/packages/components/psammead-most-read/src/utilities/rankMinWidth.js#L22-L95
https://github.com/bbc/psammead/blob/0db1f6a7e5572f2bc98b079bb7c250c0825eab4e/packages/components/psammead-most-read/src/Item/index.jsx#L68

packages/components/psammead-most-read/README.md
```diff
@@ -180,7 +186,7 @@ import Timestamp from '@bbc/psammead-timestamp';
 
 ### When to use these components
 
-These components are intended to be used on `article` and `front` pages.
+These components are intended to be used on `article`, `front` pages.
```
thekp: We also have most read on its own page now: https://test.bbc.co.uk/pidgin/popular/read
I've specified it as `mostRead` pagetype in simorgh: https://github.com/bbc/simorgh/blob/latest/src/app/routes/mostRead/index.js#L10

```suggestion
These components are intended to be used on `article`, `front` and `mostRead` pages.
```

thekp: and do we wanna mention STY's here as well 🤔

packages/components/psammead-most-read/README.md
```diff
@@ -4,6 +4,8 @@
 
 The `MostRead` component is designed to display the most read articles given a designated period of time which is dependent on service. The component comprises of a `MostReadList` containing an ordered list of `MostReadItemWrapper`. A `MostReadItemWrapper` is a list item comprised of and a `MostReadLink`, a link to the article, and a `MostReadRank` which is a numerical counter representing the article's ranking.
 
+The `MostReadLink` and `MostReadRank` components also support different a size prop. This prop currently only accepts a `default` and `small` size. Internally these values change the typography used for the components.
```
simonsinclair: In "Internally these values change the typography used for the components.", can we focus less on the typography and more on the CSS? Since we're also changing other things, like padding.

I think this might help users of this component understand that `size` relates to more than just the typography and more to the overall 'theme'.

simonsinclair (approved):

packages/components/psammead-most-read/README.md
```diff
@@ -180,7 +186,7 @@ import Timestamp from '@bbc/psammead-timestamp';
 
 ### When to use these components
 
-These components are intended to be used on `article` and `front` pages.
```
12, line 183: Might be a bit misleading, we don't have `mostRead` pages.

HarveyPeachey: Front pages have been working on a mostRead page recently https://github.com/bbc/psammead/pull/3409#discussion_r416635537

thekp: https://test.bbc.co.uk/pidgin/popular/read it's just not live yet 😭

12: Love being part time

12 (approved):
Good work

thekp (approved):
Thanks for the update Harvey, this is 💯

HarveyPeachey:
This passed the UX review 🎉

PR #3412, opened 2020-04-24
Update Swahili relative timestamps

### LINKED ISSUE #3411: Update Swahili locale relative time translations
The current relative translations provided in moment are not grammatically correct. These should be updated. 

Update the swahili locale in psammead-locales to have the correct translations for relative time.

https://github.com/bbc/psammead/blob/a5f33a798e3bd2df7e53e40f43d491fdf6ade64d/packages/utilities/psammead-locales/moment/sw.js
Add translations
```
  relativeTime: {
    past: '%s seconds ago',
    m: '1 minute ago'
    mm: '%d minutes ago,
    h: '1 hour ago'
    hh: '%d hours ago'
  },
```

N/A

**Testing notes**
Checking in Storybook https://bbc.github.io/psammead/?path=/story/utilities-psammead-locales--moment-swahili-sw

Dev insight: Unit tests should be sufficient

**Additional context**
Moment's Swahili locale: https://github.com/moment/moment/blob/develop/src/locale/sw.js#L30-L44

jimjohnsonrollings:
Resolves #3411 

**Overall change:**
Updates the Swahili relative timestamps to match editorial expectations

**Code changes:**

- Updates Swahili relative timestamps
- Updates tests

Screen Shot 2020-04-24 at 11 44 52
Screen Shot 2020-04-24 at 11 44 57
Screen Shot 2020-04-24 at 11 45 05

---

### DIFF
--- a/packages/utilities/psammead-locales/moment/sw.js
+++ b/packages/utilities/psammead-locales/moment/sw.js
@@ -9,12 +9,12 @@ moment.updateLocale('sw', {
     past: '%s',
     s: function (number, withoutSuffix, key, isFuture) {
       return withoutSuffix === false && isFuture === false
-        ? 'hivi punde zilizopita'
+        ? 'sekunde chache zilizopita'
         : 'hivi punde';
     },
     ss: function (number, withoutSuffix, key, isFuture) {
       return withoutSuffix === false && isFuture === false
-        ? 'hivi punde zilizopita'
+        ? 'sekunde chache zilizopita'
         : 'hivi punde';
     },
     m: function (number, withoutSuffix, key, isFuture) {
--- a/packages/utilities/psammead-locales/moment/sw.js
+++ b/packages/utilities/psammead-locales/moment/sw.js
@@ -5,4 +5,37 @@ moment.updateLocale('sw', {
   months: 'Januari_Februari_Machi_Aprili_Mei_Juni_Julai_Agosti_Septemba_Oktoba_Novemba_Disemba'.split(
     '_'
   ),
+  relativeTime: {
+    past: '%s',
+    s: function (number, withoutSuffix, key, isFuture) {
+      return withoutSuffix === false && isFuture === false
+        ? 'hivi punde zilizopita'
+        : 'hivi punde';
+    },
+    ss: function (number, withoutSuffix, key, isFuture) {
+      return withoutSuffix === false && isFuture === false
+        ? 'hivi punde zilizopita'
+        : 'hivi punde';
+    },
+    m: function (number, withoutSuffix, key, isFuture) {
+      return withoutSuffix === false && isFuture === false
+        ? 'Dakika 1 iliyopita'
+        : 'Dakika 1';
+    },
+    mm: function (number, withoutSuffix, key, isFuture) {
+      return withoutSuffix === false && isFuture === false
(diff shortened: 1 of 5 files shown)


packages/utilities/psammead-locales/moment/sw.test.js
```diff
@@ -141,14 +141,18 @@ test('from', function () {
 
 test('suffix', function () {
   assert.equal(moment(30000).from(0), 'hivi punde baadaye',  'prefix');
-  assert.equal(moment(0).from(30000), 'tokea hivi punde', 'suffix');
+  assert.equal(moment(0).from(30000), 'hivi punde zilizopita', 'suffix');
```
EinsteinNjoroge: `hivi punde zilizopita` doesn't sound grammaticality correct. I think just "`hivi punde` should suffice.

Cc @Bopchy

Bopchy: Yes, it doesn't sound quite right. `hivi punde` should be enough. 

EDIT: As it's being used to refer to the recent past in this context, I think `hivi punde zilizopita` is ok, as `hivi punde` means soon

packages/utilities/psammead-locales/moment/sw.js
```diff
@@ -5,4 +5,37 @@ moment.updateLocale('sw', {
   months: 'Januari_Februari_Machi_Aprili_Mei_Juni_Julai_Agosti_Septemba_Oktoba_Novemba_Disemba'.split(
     '_'
   ),
+  relativeTime: {
```
Bopchy, line 8: Curious about this approach. It is my understanding that we avoid making major changes to moment, so that these changes can be adopted by moment. Is there merit to having it this way as opposed to:

```
relativeTime: {
    past: 'hivi punde zilizopita',
    m: 'Dakika 1'
    mm: 'Dakika %d zilizopita,
    h: 'Saa 1 iliyopita',
    hh: 'Saa %d zilizopita',
  },
```
?

packages/utilities/psammead-locales/moment/sw.test.js
EinsteinNjoroge: I think `sekunde chache zilizopita` would be a better translation instead.  It means `a few seconds ago`.

packages/utilities/psammead-locales/moment/sw.js
jimjohnsonrollings: So that suggestion didn't work, because that essentially sets the value to `hivi punde zilizopita` for any relative time in the past, rather than specifying a value for how long ago it was.

There didn't seem to be a way to put logic inside the `past` translation based on the value of the variable - so to get different words for `iliyopita` and `zilizopita` based on the number, the logic had to live where I've put it.

packages/utilities/psammead-locales/moment/sw.test.js
jimjohnsonrollings: Updated :)

packages/utilities/psammead-locales/moment/sw.js
Bopchy: Ok. Thanks for the explanation

Bopchy (approved):

EinsteinNjoroge (approved):

PriyaKR:
LGTM.

PR #3426, opened 2020-04-27
Allow StoryPromo headline to optionally render without h3 tag

### LINKED ISSUE #3406: Allow StoryPromo headline to optionally render without a h3 tag
Following on from investigation https://github.com/bbc/psammead/issues/3325 to implement the Recommendations component into Story pages in Simorgh, we decided to re-use story promos to create the container, however we have acceptance criteria for the headlines in recommendations to not use any HTML heading tags. Currently, `headline` is hardcoded to use [h3's](https://github.com/bbc/psammead/blob/3db1faa1f4836d2a7a87a3b1437659d11032fe75/packages/components/psammead-story-promo/src/index.jsx#L125).

Allow the option of rendering the headline in a `<div>`, whilst keeping h3 as the default HTML element to stop this being a breaking change.

Use the [as prop](https://styled-components.com/docs/api#as-polymorphic-prop).

Limit elements rendered to a div, but allow different level of headings. Prevent images being rendered.

**Testing notes**
Add snapshot tests demonstrating the function.

**Additional context**

Simon2828:
Resolves #3406

**Overall change:** 
Allow the option of rendering the headline in a `<div>`, whilst keeping `<h3>` as the default HTML element.

**Code changes:**

- Pass a tag prop to `<Headline>`
- Use the `as` styled component prop

---


packages/components/psammead-story-promo/src/index.jsx
```diff
@@ -122,29 +122,53 @@ const headlineTypography = script => ({
...
+export const Headline = ({
+  children,
+  promoHasImage,
+  promoType,
+  script,
+  service,
+  tag,
+}) => {
+  return (
+    <StyledHeadline
```
sareh, line 134: Instead of making this change here, can we update the usage in the Story Promo container:
https://github.com/bbc/simorgh/blob/latest/src/app/containers/StoryPromo/index.jsx#L138-L143
I think that would be a better place to have the logic.

There we are able to have the standard Headline from story promo and override it with an 'as' prop only when necessary. 
```
import { Headline } from '@bbc/psammead-story-promo';
....
const headlineProps = {
  script,
  service,
  promoType,
  promoHasImage: displayImage,
  ...(headlingTagOverride && { as: headingTagOverride })
};
...
<Headline {...headlineProps}>{...elements in headline here...}</Headline>
```
These Psammead components should have as little conditional rendering as possible.

Simon2828:
Close in favour of https://github.com/bbc/simorgh/pull/6342

PR #3430, opened 2020-04-28
Change default lang prop in bulletin

### LINKED ISSUE #3428: Radio bulletin - Incorrect lang attribute on visually hidden text on arabic
**Describe the bug**
The radio bulletin on https://www.bbc.com/arabic has some visually hidden text with a lang attribute stating english as the language of this content, this attribute should not be there as the content is not in english.

**Expected behaviour**
The lang attribute should not be there as the content is not in english.

**Screenshots**
Screen Shot 2020-04-27 at 14 45 56

**Testing notes**

EinsteinNjoroge:
Partially resolves #3428 

**Overall change:** _Change default `lang` prop to `null` in psammead-bulletin and apply `lang` attribute to the hidden text if `lang` is defined_

**Code changes:**

- _Change default `lang` prop to null_
- _Apply `lang` attribute if it's defined._
- _Add test for `Arabic` service and update snapshots._

---

### DIFF
--- a/packages/components/psammead-bulletin/src/index.jsx
+++ b/packages/components/psammead-bulletin/src/index.jsx
@@ -180,12 +180,12 @@ const Bulletin = ({
   isLive,
   liveText,
   offScreenText,
+  lang,
 }) => {
   const isAudio = mediaType === 'audio';
   const bulletinType = isAudio ? 'radio' : 'tv';
   const BulletinWrapper = isAudio ? RadioBulletinWrapper : TVBulletinWrapper;
-  const ctaTextIsEnglish = ctaText === 'Watch' || ctaText === 'Listen';
-  const langProp = ctaTextIsEnglish ? { lang: 'en-GB' } : {};
+  const hiddenTextProps = lang ? { lang } : {};
 
   return (
     <BulletinWrapper>
@@ -214,7 +214,7 @@ const Bulletin = ({
               // eslint-disable-next-line jsx-a11y/aria-role
               <span role="text">
                 {offScreenText && (
-                  <VisuallyHiddenText {...langProp}>
+                  <VisuallyHiddenText {...hiddenTextProps}>
                     {`${offScreenText}, `}
                   </VisuallyHiddenText>
                 )}
@@ -259,6 +259,7 @@ Bulletin.propTypes = {
   isLive: bool,
   liveText: string,
   offScreenText: string.isRequired,
+  lang: string,
 };
 
 Bulletin.defaultProps = {
@@ -266,6 +267,7 @@ Bulletin.defaultProps = {
   image: null,
   isLive: false,
   liveText: 'LIVE',
+  lang: null,
 };
 
 export default Bulletin;
--- a/packages/components/psammead-bulletin/src/index.jsx
+++ b/packages/components/psammead-bulletin/src/index.jsx
@@ -180,11 +180,12 @@ const Bulletin = ({
   isLive,
   liveText,
   offScreenText,
-  lang,
 }) => {
   const isAudio = mediaType === 'audio';
   const bulletinType = isAudio ? 'radio' : 'tv';
   const BulletinWrapper = isAudio ? RadioBulletinWrapper : TVBulletinWrapper;
+  const ctaTextIsEnglish = ctaText === 'Watch' || ctaText === 'Listen';
+  const langProp = ctaTextIsEnglish ? { lang: 'en-GB' } : {};
 
   return (
     <BulletinWrapper>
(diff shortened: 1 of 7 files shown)


packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -258,15 +259,13 @@ Bulletin.propTypes = {
...
   offScreenText: string.isRequired,
-  lang: string,
 };
 
 Bulletin.defaultProps = {
   dir: 'ltr',
   image: null,
   isLive: false,
   liveText: 'LIVE',
-  lang: 'en-GB',
```
DenisHdz, line 269: I think we can leave this prop but set it as `null` by default.

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -180,11 +180,12 @@ const Bulletin = ({
   isLive,
   liveText,
   offScreenText,
-  lang,
 }) => {
   const isAudio = mediaType === 'audio';
   const bulletinType = isAudio ? 'radio' : 'tv';
   const BulletinWrapper = isAudio ? RadioBulletinWrapper : TVBulletinWrapper;
+  const ctaTextIsEnglish = ctaText === 'Watch' || ctaText === 'Listen';
```
DenisHdz: Looks good, but I was thinking to do this in Simorgh instead, as we do in the `StoryPromo` container with the Live label https://github.com/bbc/simorgh/blob/latest/src/app/containers/StoryPromo/index.jsx#L101. Happy to hear other people thoughts

EinsteinNjoroge: Good idea.
I think that we'll be a better approach to separate the psammead components from our simorgh implementation.

packages/components/psammead-bulletin/src/index.test.jsx
```diff
@@ -1,11 +1,19 @@
...
 /* eslint-disable react/prop-types */
-const BulletinComponent = ({ script, service, isLive, mediaType, ctaText }) => {
+const BulletinComponent = ({
+  script,
+  service,
+  isLive,
+  mediaType,
+  ctaText,
+  dir = 'ltr',
+  lang = null,
```
DenisHdz, line 15: We can remove this now
```suggestion

```

packages/components/psammead-bulletin/src/index.test.jsx
```diff
@@ -31,6 +39,8 @@ const BulletinComponent = ({ script, service, isLive, mediaType, ctaText }) => {
       ctaText={playCtaText}
       isLive={isLive}
       offScreenText={offScreenText}
+      dir={dir}
+      lang={lang}
```
DenisHdz, line 43:
```suggestion

```

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -185,6 +185,7 @@ const Bulletin = ({
   const isAudio = mediaType === 'audio';
   const bulletinType = isAudio ? 'radio' : 'tv';
   const BulletinWrapper = isAudio ? RadioBulletinWrapper : TVBulletinWrapper;
+  const hiddenTextProps = lang ? { lang } : {};
```
DenisHdz, line 188: I think this will just add the `lang` prop as a boolean and we want `lang={lang}`.

packages/components/psammead-bulletin/src/index.jsx
```diff
@@ -213,7 +214,7 @@ const Bulletin = ({
               // eslint-disable-next-line jsx-a11y/aria-role
               <span role="text">
                 {offScreenText && (
-                  <VisuallyHiddenText lang={lang}>
```
DenisHdz: We can keep this as it was, because if the language is not passed it will be `lang=null` which is the same that `const hiddenTextProps = lang ? { lang } : {};` will return.

packages/components/psammead-bulletin/src/index.test.jsx
EinsteinNjoroge: @DenisHdz The intention here is to test for the right output when the `lang` prop is passed in.

packages/components/psammead-bulletin/src/index.test.jsx
EinsteinNjoroge: https://github.com/bbc/psammead/pull/3430#discussion_r416735056

packages/components/psammead-bulletin/src/index.test.jsx
DenisHdz: Of course, we need it for the other snapshots https://github.com/bbc/psammead/pull/3430/files#diff-5ae3e98c9edd526e83ba2518475fdcafR67 👍

DenisHdz (approved):
Thanks for the changes 👍

OlgaLyubin (approved):
Looks good 👍

EinsteinNjoroge:
This PR doesn't need testing at the component level, it will be tested once it's pulled in to `simorgh`

PR #3438, opened 2020-04-30
Export visually hidden text styles

### LINKED ISSUE #3361: Update Visually Hidden Text to support more use cases

This component returns a fixed element with the visually hidden styles applied. We have found a need to apply visually hidden styles to other wrapping elements, and also control when and where these styles are applied.

E.G. An anchor that becomes visible when it's focussed.

```js
const SkipLink = styled.a`
  &:not(:focus):not(:active) {
    // Styles that keep the link hidden when it's not in focus or active.
  }
`;
```

Flexibility could be achieved, while maintaining backwards compatibility, by adding an export for the style string. E.G.

```js
import { visuallyHiddenStyle } from '@bbc/psammead-visually-hidden-text';

const SkipLink = styled.a`
  &:not(:focus):not(:active) {
    ${visuallyHiddenStyle}
  }
`;
```

n/a

**Testing notes**
[Tester to complete]

**Additional context**
None.

EinsteinNjoroge:
Resolves #3361 

**Overall change:** _Export visually hidden text styles separate from the `visuallyHiddenText`._

**Code changes:**

- _Export visually hidden text styles._
- _Bump package._

---

### DIFF
--- a/packages/components/psammead-visually-hidden-text/src/index.jsx
+++ b/packages/components/psammead-visually-hidden-text/src/index.jsx
@@ -1,6 +1,6 @@
-import styled, { css } from 'styled-components';
+import styled from 'styled-components';
 
-export const VisuallyHiddenTextStyle = css`
+export const visuallyHiddenTextStyle = `
   clip-path: inset(100%);
   clip: rect(1px, 1px, 1px, 1px);
   height: 1px;
@@ -11,7 +11,7 @@ export const VisuallyHiddenTextStyle = css`
 `;
 
 const VisuallyHiddenText = styled.span`
-  ${VisuallyHiddenTextStyle};
+  ${visuallyHiddenTextStyle};
 `;
 
 export default VisuallyHiddenText;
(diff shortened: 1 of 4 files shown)


packages/components/psammead-visually-hidden-text/src/index.jsx
```diff
@@ -1,6 +1,6 @@
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 
-const VisuallyHiddenText = styled.span`
+export const VisuallyHiddenTextStyle = css`
```
simonsinclair: It might help to know that we don't need to use a `css` tagged template literal here, because we're dealing with a simple string.

https://styled-components.com/docs/api#css

This could be re-written as:

```js
export const visuallyHiddenTextStyle = `
  clip-path: inset(100%);
  clip: rect(1px, 1px, 1px, 1px);
  height: 1px;
  overflow: hidden;
  position: absolute;
  width: 1px;
  margin: 0;
`;
```

_Note: You may also notice this example uses camelCase, because we're not exporting a React component (typically written in TitleCase), but a string._

simonsinclair (commented):
Thanks so much for picking this up, @EinsteinNjoroge! We can simplify this even further, if you like. Please see my comment below.

simonsinclair (approved):

tochwill (approved):

EinsteinNjoroge:
This PR doesn't need testing since it's just a code refactor that doesn't update any snapshots.

PR #3472, opened 2020-05-15
do not return current date when timestamp is undefined

### LINKED ISSUE #3474: formatUnixTimestamp returns current date when timestamp is undefined
**Describe the bug**
`formatUnixTimestamp` returns the current date when the timestamp passed in is `undefined`.

While this is down to the default behaviour of the `moment` call it can result in cases where we show the incorrect timestamp if we don't plumb data to timestamp related components and functions correctly because we might not easily spot the difference between the current date and other dates in different languages.

**To Reproduce**
`formatUnixTimestamp(); // current date`

**Expected behaviour**
`formatUnixTimestamp(); // undefined`
`formatUnixTimestamp({ timestamp: Date.now() }); // current date`

**Screenshots**

**Testing notes**

**Additional context**

jroebu14:
Resolves https://github.com/bbc/psammead/issues/3474

**Overall change:**
Fixes bug where `formatUnixTimestamp` returns the current date when you do not pass in a timestamp.

This will help us more easily catch cases when a timestamp is undefined because instead of outputting today's date, which can be difficult to spot the difference in another language, it won't output anything.

**Code changes:**

- Early return if `timestamp` value is undefined
- Add tests to make this a requirement of the function

---


ryanmccombe (approved):
I think the bug might be in our use of this module, rather than the module itself

Using the current time might be a sensible default.  It provides a quick way to output a formatted datestamp, just like the moment implementation

Either way is good though 😄 just needs a version bump

jroebu14:
@ryanmccombe in that case you could do `formatUnixTimestamp({ timestamp: Date.now() })`

packages/containers/psammead-timestamp-container/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 4.0.0 | [PR#3468](https://github.com/bbc/psammead/pull/3472) Do not return current date in formatUnixTimestamp when timestamp is undefined |
```
ryanmccombe:
```suggestion
| 4.0.0 | [PR#3472](https://github.com/bbc/psammead/pull/3472) Do not return current date in formatUnixTimestamp when timestamp is undefined |
```

ryanmccombe:
Just wanted to clarify because I think my post was confused given the standup update

I'm not saying we don't have a bug here, I'm just saying that having a date formatter use the current date by default is not necessarily where the bug is.

This is an established behavior in Moment, as well as JS itself (calling the date constructor with no arguments assumes you want the current date - `new Date()`)

Like you said, I'm fairly ambivalent _how_ we fix the bug - but I'm not ambivalent about _whether_ we fix the bug 😛

RichardPK (approved):
Good solution and important to help debugging imo. I think you mentioned you had - but worth getting @sareh 's 👀 across this?

jroebu14:
> Nothing is better than incorrect information

 _Lily Higham_

PR #3489, opened 2020-05-21
Create `PageHeading` component

DenisHdz:
Part of https://github.com/bbc/simorgh/issues/6414

**Overall change:** 
Create a new `PageHeading` component so it can be reused on different page types as `IDX` and `Most Read`.

**Code changes:**
- Add the new `PageHeading` component within the existing `psammead-headings` package
- Add a new story
- Add snapshots

---

### DIFF
--- a/packages/components/psammead-headings/README.md
+++ b/packages/components/psammead-headings/README.md
@@ -57,9 +57,7 @@ These components can be used at any point on the page, however the `Headline` an
 
 ### Accessibility notes
 
-The `SubHeading` component has a tabindex of `-1` so that it works correctly with screen readers when navigated to via a skip link.
-
-The `Heading` has an id of `content` so it can be announced by screen readers when activating the `Skip to content` link.
+The `SubHeading` and `PageHeading` component have a tabindex of `-1` so that they work correctly with screen readers when navigated to via a skip link.
 
 <!-- ## Roadmap -->
 
--- a/packages/components/psammead-headings/README.md
+++ b/packages/components/psammead-headings/README.md
@@ -39,7 +39,7 @@ const Wrapper = () => (
 );
 ```
 
-`SubHeading` components can be used as page anchors when passed an `id` prop. To take the above usage as an example:
+`SubHeading` and `PageHeading` components can be used as page anchors when passed an `id` attribute. To take the above usage as an example:
 
 ```jsx
 <SubHeading id="some-subheadline" script={latin} service="news">
@@ -51,7 +51,9 @@ This usage will allow for the page anchor: `www.bbc.com/news/articles/articleID#
 
 ### When to use this component
 
-These components can be used at any point on the page, however the `Headline` and `PageHeading` are designed to be used once at the top of the page. The `SubHeading` takes an optional `id` value and passes it to the `h2` which can be used as an anchor when referencing content.
+These components can be used at any point on the page, however the `Headline` and `PageHeading` are designed to be used once at the top of the page.
+
+The `SubHeading` and `PageHeading` can take an optional `id` attribute to be passed to the `h2` which can be used as an anchor when referencing content.
 
 <!-- ### When not to use this component -->
 
--- a/packages/components/psammead-headings/README.md
+++ b/packages/components/psammead-headings/README.md
@@ -59,7 +59,7 @@ The `SubHeading` and `PageHeading` can take an optional `id` attribute which can
 
 ### Accessibility notes
 
-The `SubHeading` and `PageHeading` component have a tabindex of `-1` so that they work correctly with screen readers when navigated to via a skip link.
+The `SubHeading` and `PageHeading` components have a tabindex of `-1`, this ensures that these elements are focusable by assitive technology.
 
 <!-- ## Roadmap -->
 
--- a/packages/components/psammead-headings/README.md
+++ b/packages/components/psammead-headings/README.md
@@ -3,11 +3,11 @@
 
 ## Description
 
-The Headings are a set of three components, `Headline`, `SubHeading` and `PageHeading`.
+The Headings are a set of three components, `Headline`, `SubHeading` and `IndexH1`.
 
-They `Headline` and `SubHeading` use a `h1` and `h2` HTML element respectively, with the `Headline` being designed for a single use on the page, with `SubHeading` being aimed at repeated use.
+The `Headline` and `SubHeading` use a `h1` and `h2` HTML elements respectively, these are used on Article and STY (story) pages.
 
-On the other hand, the `PageHeading` use a `h1` and is designed to be the title of a page, like for example IDX, FIX and Most Read pages.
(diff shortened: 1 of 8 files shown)


packages/components/psammead-headings/src/index.jsx
```diff
@@ -45,3 +59,19 @@ SubHeading.propTypes = {
   script: shape(scriptPropType).isRequired,
   service: string.isRequired,
 };
+
+export const PageHeading = styled.h1.attrs(() => ({
+  id: 'content',
```
DenisHdz: I need to confirm with @greenc05 if we need to add the `tabIndex:-1` attribute.

packages/components/psammead-headings/src/index.jsx
```diff
@@ -45,3 +59,19 @@ SubHeading.propTypes = {
...
+}))`
+  color: ${C_METAL};
+  ${({ script }) => script && getDoublePica(script)};
+  ${({ service }) => getSansRegular(service)};
+  margin: 0;
+  padding: ${GEL_SPACING_DBL} 0 ${GEL_SPACING_TRPL};
```
DenisHdz: What are people's thoughts on adding the spacing directly in the component?

When we generalized the Story and Bulletin promos and the Media Indicator, we tried to remove any outer spacing and apply this in Simorgh. 

The designs seem to have the same padding for IDX and Most Read page, but this component could be used in a different page type which could require a different spacing cc @sareh

OlgaLyubin (commented):
Since `psammead-headings` is one component, should we have it as one component in storybook too? And have `Headline`, `SubHeading` and `PageHeading` as substories instead of having them separately. What do you think?

Screenshot 2020-05-22 at 10 40 31

DenisHdz:
> Since `psammead-headings` is one component, should we have it as one component in stories too? And have `Headline`, `SubHeading` and `PageHeading` as substories instead of having them separately. What do you think?
> 
> Screenshot 2020-05-22 at 10 40 31

I have also thought about that 👍

packages/components/psammead-headings/src/index.jsx
thekp: I’m wondering if we should add the padding logic to our HeadingsContainer in simorgh, and export it from there, this way our psammead comp is generalised and if we do need to update the spacing in the future it would quicker to do if it’s in simorgh.

thekp (approved):
Aside from the open questions...

everything else looks good to me 👍

packages/components/psammead-headings/src/index.jsx
DenisHdz: Confirmed. We need to add it to the `h1` so it is focusable by AT.

EinsteinNjoroge: Since psammead is just a component library, I think we should not set the `id` attribute here. This should be done in our simorgh implementation.

OlgaLyubin (approved):
Looks good 👍

packages/components/psammead-headings/src/index.jsx
DenisHdz: That makes sense, we were doing that already with the `SubHeading` so I've updated the README and stories to reflect that.

EinsteinNjoroge (approved):

packages/components/psammead-headings/README.md
```diff
@@ -39,7 +39,7 @@ const Wrapper = () => (
 );
 ```
 
-`SubHeading` components can be used as page anchors when passed an `id` prop. To take the above usage as an example:
+`SubHeading` and `PageHeading` components can be used as page anchors when passed an `id` attribute. To take the above usage as an example:
```
greenc05: Should we also add the following? "Note, to make them focusable by assitive technology they will also need a tab index attribute with a value of -1."

packages/components/psammead-headings/README.md
```diff
@@ -51,7 +51,9 @@ This usage will allow for the page anchor: `www.bbc.com/news/articles/articleID#
 
 ### When to use this component
 
-These components can be used at any point on the page, however the `Headline` and `PageHeading` are designed to be used once at the top of the page. The `SubHeading` takes an optional `id` value and passes it to the `h2` which can be used as an anchor when referencing content.
+These components can be used at any point on the page, however the `Headline` and `PageHeading` are designed to be used once at the top of the page.
```
greenc05: What is 'Headline'?

packages/components/psammead-headings/README.md
DenisHdz: It is already mentioned further down https://github.com/bbc/psammead/pull/3489/files#diff-d3f2566006f1e21e5760acc50bc7db4dR62

packages/components/psammead-headings/README.md
```diff
@@ -51,7 +51,9 @@ This usage will allow for the page anchor: `www.bbc.com/news/articles/articleID#
...
+
+The `SubHeading` and `PageHeading` can take an optional `id` attribute to be passed to the `h2` which can be used as an anchor when referencing content.
```
greenc05: h2?

packages/components/psammead-headings/README.md
DenisHdz: It is explained at the top of the README:

```
They `Headline` and `SubHeading` use a `h1` and `h2` HTML element respectively, with the `Headline` being designed for a single use on the page, with `SubHeading` being aimed at repeated use.
```

packages/components/psammead-headings/README.md
DenisHdz: I'll correct this since the `SubHeading` component uses a `h2` but not the `PageHeading`

packages/components/psammead-headings/README.md
```diff
@@ -47,13 +51,15 @@ This usage will allow for the page anchor: `www.bbc.com/news/articles/articleID#
...
+These components can be used at any point on the page, however the `Headline` and `PageHeading` are designed to be used once at the top of the page.
+
+The `SubHeading` and `PageHeading` can take an optional `id` attribute which can be used as an anchor when referencing content.
 
 <!-- ### When not to use this component -->
 
 ### Accessibility notes
 
-The `SubHeading` component has a tabindex of `-1` so that it works correctly with screen readers when navigated to via a skip link.
+The `SubHeading` and `PageHeading` component have a tabindex of `-1` so that they work correctly with screen readers when navigated to via a skip link.
```
greenc05: Please can we remove 'skip link' - This functionality for subheads wasn't designed for skip links specifially.

greenc05: I know you haven't added that....

greenc05: How about? "The `SubHeading` and `PageHeading` components have a tabindex of `-1`, this ensures that these elements are focusable by assitive technology."

DenisHdz:
Close in favour of https://github.com/bbc/psammead/pull/3501

PR #3530, opened 2020-06-12
Remove Index Alsos from Grid stories

DenisHdz:
**Overall change:** 
In the Psammead Grid stories, we were importing the `IndexAlsosContainer` and `relatedItems` from a relative path outside the package which is wrong. 

The `relatedItems` can not be accessed directly from the `@bbc/psammead-story-promo`, so I've removed them since they are also irrelevant for the Grid stories we have. 

**Code changes:**
- Remove IndexAlsos from `Example with Top story and regular promos` story

---

### DIFF
--- a/packages/components/psammead-grid/CHANGELOG.md
+++ b/packages/components/psammead-grid/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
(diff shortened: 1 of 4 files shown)


thekp (approved):
LGTM,

> In the Psammead Grid stories, we were importing the IndexAlsosContainer and relatedItems from a relative path outside the package which is wrong.

I believe this works in storybook, because our storybook application treats the entire psammead directory as a single project. However, I also agree we shouldn't be importing directly from different packages within psammead.

OlgaLyubin (approved):
LGTM

PR #3585, opened 2020-07-07
adds prop to override h2 element

### LINKED ISSUE #3574: Allow section-label heading optionally use a `strong` element

`psammead-section-label` currently only renders an H2. We need to be able to choose which element is rendered _inside_ to meet a11y requirements. At the moment, this isn't possible using the [Styled Components `as=` prop](https://styled-components.com/docs/api#as-polymorphic-prop), because styles are tightly coupled to the nested elements.

[Psammead Section Label](https://github.com/bbc/psammead/blob/latest/packages/components/psammead-section-label/src/index.jsx)

We could:
- Pass an `as` prop value down onto the children components.
- Make `SectionLabel` accept `children` and assign the `as` prop directly.

- consider moving all typography, font and style settings from plain title to the top level `h2` that way we don't use lose those if it's overridden with the `"as"` prop.

- consider if we should override the entire section-label in simorgh for our use case.

**Testing notes**
Add snapshot tests demonstrating the function.

- confirm with @greenc05 if we have some other place where section-label uses a strong element and why we need to do that here.
- see where the section label is being used and consider if we could dry up the logic in the section-label component to distinguish if we render a link or plain title.

rhenshaw56:
Resolves #3574 

**Overall change:** _Optionally render any element instead of an h2._

**Code changes:**

- _Adds an overrideHeadingAs prop to override the default h2._

---


packages/components/psammead-section-label/src/__snapshots__/index.test.jsx.snap
```diff
@@ -1821,44 +1841,47 @@ exports[`SectionLabel Without bar With linking title should render correctly 1`]
...
           >
             This is text in a SectionLabel.
           </span>
           <span
             aria-hidden="true"
-            class="c6"
+            class="c7"
             dir="ltr"
           >
             See All
```
greenc05: If the journalist inputs a link - we won't want this in the recommendations component... will it be possible for a journalist to input a link?

rhenshaw56: @greenc05 in the recommendations container in simorgh, it will be hard coded to not have a link text or an href, so we never have this outputted as a link

greenc05: Thanks @rhenshaw56 maybe we don't need tests for this scenario then?

packages/components/psammead-section-label/src/index.test.jsx
```diff
@@ -236,4 +236,19 @@ describe('SectionLabel', () => {
       ).toEqual('section_name');
     });
   });
+
+  describe('With heading overriden to be a strong element', () => {
```
HarveyPeachey: Maybe a test case for an invalid override element?

packages/components/psammead-section-label/src/index.jsx
```diff
@@ -133,6 +135,7 @@ SectionLabel.propTypes = {
   service: string.isRequired,
   visuallyHidden: bool,
   backgroundColor: string,
+  overrideHeadingAs: string,
```
andrewscfc: should we restrict the allowed value for this to be only null or `strong` at this time, just to discourage misuse and encourage people to choose the correct element type in the future?

packages/components/psammead-section-label/src/index.jsx
```diff
@@ -81,11 +81,12 @@ const SectionLabel = ({
   service,
   visuallyHidden,
   backgroundColor,
+  overrideHeadingAs,
```
andrewscfc, line 84: When this is used in Simorgh could we summarise why we are overriding to `strong` in a comment I don't think this will be very obvious when people come to this in the future https://github.com/bbc/psammead/issues/3574#issuecomment-654280799 

I was considering suggesting adding a `supplementary` prop, this trying to express the reason for using `strong`, the section label being inside supplementary content rather than being used in the main content of the page; this could then be used inside the component to switch to use a `strong`. I'm still not sure this is clear enough though, so I'd personally just suggest adding a comment as above.

packages/components/psammead-section-label/src/index.jsx
greenc05: If that is possible, definately

rhenshaw56: @andrewscfc would a prop validation warning suffice? or should it be strictly enforced?

packages/components/psammead-section-label/src/index.test.jsx
rhenshaw56: @HarveyPeachey [done! ](https://github.com/bbc/psammead/pull/3585/files#diff-00f9b2243883179eeed7e9ef2e9fe3b3R255-R272)

packages/components/psammead-section-label/src/index.jsx
andrewscfc: I'd just say prop validation personally, just to guide people in the right direction. Enforcing is a bit pointless because anyone can edit the code, it's just to encourage people to pause and think really.

andrewscfc (approved):
Thanks for addressing my comment apologies for the slow reply.

HarveyPeachey (approved):
Nice job Rowland, LGTM 👍

packages/components/psammead-section-label/src/__snapshots__/index.test.jsx.snap
```diff
@@ -1657,6 +1657,272 @@ exports[`SectionLabel With bar With plain title should render correctly with exp
...
+        >
+          This is text in a SectionLabel.
+        </span>
+      </span>
+    </span>
+  </strong>
+</div>
+`;
+
+exports[`SectionLabel With heading overriden to be a strong element should render correctly as a plain title with no bar 1`] = `
```
greenc05: whats the difference in this to the example above?

rhenshaw56: @greenc05 none actually, just a change in the test description title.

paruchurisilpa:
Looks good to me. Looked at Storybook and we have an option under sectionlabel for 'with heading overriden to be a strong element'

PR #3588, opened 2020-07-09
Adds support to configure story-promo-list to different padding rules

### LINKED ISSUE #3572: Make StoryPromoLi configurable to different padding requirements
We would like to reuse the `StoryPromoLi` component with a different set of ux requirement opposed to how it was originally built, hence the need to make it more configurable for different padding rules.

Instead of entirely replacing the original `StoryPromoLi` component we could create a new `StoryPromoLiThin` in the same file with just bare bones styles and no padding. ie.
```
export const StoryPromoLiThin = styled.li.attrs({
  role: 'listitem',
})`

  @media (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
    border-bottom: 0.0625rem solid ${C_LUNAR};
  }

  ${({ border }) =>
    !border &&
    `
      @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) and (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
          border-bottom: none;
      }
    `}
`;
```

**Testing notes**
- snapshots, chromatic checks, storybook addition.

**Additional context**

rhenshaw56:
Resolves #3572 

**Overall change:** _Adds a new StoryPromoLiThin component that conforms to any given set of padding without restrictions ie. overriding it set padding._

**Code changes:**

- _Adds a new StoryPromoLiThin component._

---


packages/components/psammead-story-promo-list/src/index.jsx
```diff
@@ -12,32 +12,20 @@ import {
   GEL_GROUP_4_SCREEN_WIDTH_MIN,
 } from '@bbc/gel-foundations/breakpoints';
 
-export const StoryPromoLi = styled.li.attrs({
+export const StoryPromoLiThin = styled.li.attrs({
```
simonsinclair: This component appears to have a number of redundant styles/some that I would expect to be applied by the parent. Can we remove them?

packages/components/psammead-story-promo-list/src/index.jsx
```diff
@@ -12,32 +12,20 @@ import {
...
-    border-bottom: 0.0625rem solid ${C_LUNAR};
-  }
-
   ${({ border }) =>
     !border &&
     `
-      @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) and (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
-          border-bottom: none;
-      }
-    `}
```
simonsinclair: It would make sense if this _added_ border styles instead of removed/overwrote them. If we want border to be the default, we can make use of `defaultProps`.

N.B. Currently `border` styles exist on this component by default, which means we have to add the inverse to remove them. This is made even more complex by the media query.

packages/components/psammead-story-promo-list/src/index.jsx
```diff
@@ -52,6 +40,18 @@ export const StoryPromoLi = styled.li.attrs({
   }
 `;
 
+export const StoryPromoLi = styled(StoryPromoLiThin)`
```
simonsinclair: Perhaps if `StoryPromoLiThin` is being used as a 'base class', like it is here, it should be called `StoryPromoLiBase`? Initially I thought `StoryPromoLiThin` would have unshared styles, but if that's not the case I don't think it needs a 'variation name'.

This would also be more in-keeping with what's been written in the readme:
> ...`StoryPromoLiThin` ships with no padding presets thereby making it much more configurable to any set of padding/margin rules set on child elements

simonsinclair (commented):
Great work, Rowland.
I've added a few comments below. Please let me know if they need further clarification and we can arrange a call.

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: @simonsinclair which redundant styles do you mean ? the aim wasn't to get rid of any styles as we want the same behavior the story-promo-li gives us but with much more flexibility on how we apply spacing

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: I thought of that actually and was initially confused as to why it did that maybe that was a requirement when it was first built. The component is already used in a bunch of places so reversing that would mean we would need to go update all instances where it's been used.

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: Great idea!

packages/components/psammead-story-promo-list/src/index.stories.jsx
```diff
@@ -112,3 +112,32 @@ storiesOf('Components|StoryPromo/StoryPromoList', module)
       </Grid>
     );
   });
+
+storiesOf('Components|StoryPromo/StoryPromoListThin', module).add(
```
Simon2828:
```suggestion
storiesOf('Components|StoryPromo/StoryPromoListBase', module).add(
```

packages/components/psammead-story-promo-list/src/index.test.jsx
```diff
@@ -37,3 +37,23 @@ describe('StoryPromo list', () => {
     </StoryPromoUl>,
   );
 });
+
+describe('StoryPromo list thin', () => {
```
Simon2828:
```suggestion
describe('StoryPromo list base', () => {
```

packages/components/psammead-story-promo-list/src/index.jsx
simonsinclair:
```
&:first-child {
    padding-top: 0;
```
What `padding-top: 0;`'s purpose on an element that already has no padding?

packages/components/psammead-story-promo-list/src/index.jsx
simonsinclair: Yeah, it is confusing. I think there's an opportunity to fix this now. Here are the places it's being used in Simorgh: https://github.com/bbc/simorgh/search?q=psammead-story-promo-list&unscoped_q=psammead-story-promo-list.

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: 😄 I agree, that needs to go

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: @simonsinclair let's do it 💪

rhenshaw56: @simonsinclair done!

packages/components/psammead-story-promo-list/src/index.jsx
```diff
@@ -12,22 +12,23 @@ import {
...
-      }
-    `}
+    @media (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
+      border-bottom: 0.0625rem solid ${C_LUNAR};
+    }
+  `}
 
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) and (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
+    border-bottom: none;
+  }
```
simonsinclair: Is this style setting `border-bottom: none;` on an element that doesn't have a `border-bottom` value set?

packages/components/psammead-story-promo-list/src/index.jsx
```diff
@@ -72,3 +72,12 @@ StoryPromoLi.propTypes = {
...
 };
+
+StoryPromoLiBase.propTypes = {
+  children: node.isRequired,
+  border: bool,
+};
+
+StoryPromoLiBase.defaultProps = {
+  border: true,
+};
```
simonsinclair: These are identical to `StoryPromoLi`. Can we create a shared `propTypes` and `defaultProps` object?

packages/components/psammead-story-promo-list/src/index.jsx
```diff
@@ -12,22 +12,23 @@ import {
...
   role: 'listitem',
 })`
-
-  @media (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
-    border-bottom: 0.0625rem solid ${C_LUNAR};
-  }
-
   ${({ border }) =>
-    !border &&
+    border &&
```
simonsinclair, line 28: It looks like this block might be missing some styles:

If a border is set, do we need to ensure it isn't added to the last child?

packages/components/psammead-story-promo-list/src/index.stories.jsx
```diff
@@ -112,3 +112,32 @@ storiesOf('Components|StoryPromo/StoryPromoList', module)
       </Grid>
     );
   });
+
+storiesOf('Components|StoryPromo/StoryPromoListBase', module).add(
```
simonsinclair: It'd be great if we could add a `border` checkbox knob to these stories.

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: @simonsinclair the `border-bottom:none;` is a default for when `border` is `false` and `border-bottom: 0.0625rem solid ${C_LUNAR};` is set for when `border` is `true` at the same break point.

simonsinclair: `border-bottom: none;` should already be the default without this style - unless something is adding a border to it elsewhere?

Simon2828: Could this media query be removed if l.21 was changed to ${GEL_GROUP_2_SCREEN_WIDTH_MAX} ?

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: done.

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: Good spot @simonsinclair, fixed.

packages/components/psammead-story-promo-list/src/index.jsx
rhenshaw56: @Simon2828 I just confirmed [from here](https://bbc.github.io/psammead/?path=/story/components-storypromo-storypromolist--default) that the `border-bottom:none;` isn't needed anymore, it was previously when we were negating the logic [see comment](https://github.com/bbc/psammead/pull/3588#discussion_r452306854).

rhenshaw56: @simonsinclair fixed.

packages/components/psammead-story-promo-list/src/index.stories.jsx
rhenshaw56: @simonsinclair done!

Simon2828 (approved):
LGTM 👍

packages/components/psammead-story-promo-list/src/index.stories.jsx
andrewscfc: ~This doesn't seem to change anything when I check it in storybook: https://5d28eb5ee163f6002046d6fb-erzsbwuwtq.chromatic.com/?path=/story/components-storypromo-storypromolistbase--default~
Didn't resize to the correct breakpoint, can see its effect now 👍

andrewscfc (approved):

packages/components/psammead-story-promo-list/src/index.jsx
```diff
@@ -12,22 +12,36 @@ import {
...
+    @media (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
+      border-bottom: 0.0625rem solid ${C_LUNAR};
+    }
+  `}
 
-  @media (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
-    border-bottom: 0.0625rem solid ${C_LUNAR};
+  &:last-child {
+    border: none;
   }
```
simonsinclair, line 37: This should ideally be part of the block that gets added when border styles are set.

```js
${({ border }) =>
  border &&
  `
  @media (max-width: ${GEL_GROUP_3_SCREEN_WIDTH_MAX}) {
    border-bottom: 0.0625rem solid ${C_LUNAR};
  }
  &:last-child {
    border: none;
  }
`}
```

packages/components/psammead-story-promo-list/src/index.stories.jsx
```diff
@@ -54,7 +56,10 @@ storiesOf('Components|StoryPromo/StoryPromoList', module)
           );
 
           return (
-            <StoryPromoLi key={item.info.headline}>
+            <StoryPromoLi
+              key={item.info.headline}
+              border={boolean('show border?', true)}
+            >
```
simonsinclair, line 62: This is great. 😊

simonsinclair (approved):
Thanks for addressing all my comments, Rowland. This is a really thorough job. 👍 
Please see one final comment below.

rhenshaw56:
Moving from ready to test as the update to simorgh would be on a standalone pr, so we can can properly observe regression on the changes..

PR #3655, opened 2020-08-05
Update elliptic to security patched 6.5.3

andrewscfc:
No issue

**Overall change:** 
Update elliptic to security patched 6.5.3, associated packages that use it updated

**Code changes:**
Before:
image

- `npm update --depth 10 browserify-sign`
- `npm update --depth 10 create-ecdh`

After:
image

Ideally `@storybook/react` would have released an update and all dependencies would be upgraded down to the patched `elliptic`, as it is the highest packages in the dependency tree that could be updated were `browserify-sign` and `create-ecdh`. The change has little risk as it only related to Storybook.

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead",
-  "version": "2.1.0",
+  "version": "2.1.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -10222,8 +10222,7 @@
     "brorand": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/brorand/-/brorand-1.1.0.tgz",
-      "integrity": "sha1-EsJe/kCkXjwyPrhnWgoM5XsiNx8=",
-      "dev": true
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -10304,22 +10303,45 @@
...
         "create-hmac": "^1.1.7",
-        "elliptic": "^6.5.2",
+        "elliptic": "^6.5.3",
         "inherits": "^2.0.4",
         "parse-asn1": "^5.1.5",
         "readable-stream": "^3.6.0",
         "safe-buffer": "^5.2.0"
       },
       "dependencies": {
+        "elliptic": {
```
HarveyPeachey, line 10322: Just questioning why this wasn't listed under `dependencies` before but now it is?

andrewscfc: I don't know tbh, `browserify-sign` explicitly calls out `6.5.3` as being the version it requires, I'm not sure of the significance of the dependencies section here.

HarveyPeachey: Fairs, the build runs so it looks like we're good👍

HarveyPeachey (approved):

jroebu14 (approved):

PR #3728, opened 2020-09-02
Storybook V6 upgrade

andrewscfc:
Migrating Storybook from v 5.3.19 to 6.X.X

## Resolved issues:

Some minor stuff around:
- Updated integration of a11y addon
- Implementation of new `manager.js` `main.js` `preview.js` structure
- Removed use of `addParameters` from `preview.js`, migrated to the new method
- Importing a `storybook.css` file to overwrite new padding being applied to the story `<body>` tag

### 1. Components aren't re-rendering when navigating around Storybook, or modifying knobs. 
**[FIXED ✅ ]** 
_`6.0.2` & `6.0.21` (with or without forcing `@storybook/core-events` resolutions)._ 

**tl;dr: Fixed by migrating from `notes` to `docs` and removing the `storybook-readme` package. Note that docs is far more 'sophisticated' than notes and is trying to do a lot of dynamic stuff to our readmes, that doesn't work very well. One for the backlog**

Knobs and path changes cause an error, and the component does not re-render. 
Console error: 
```
TypeError: Cannot read property 'debug' of undefined
    at PostmsgTransport.handleEvent
```

It turns out this is being caused by our use of `addon-notes`, which has been deprecated in v6 and replaced by [addon-docs](https://github.com/storybookjs/storybook/tree/next/addons/docs)

A bit of digging showed that we still had an old instance of `@storybook/logger`, which was erroring. This was being used by an old instance of `@storybook/components`, which was being used by... `storybook-readme: ^5.0.8` 

Completely nuking `addon-notes` & `storybook-readme` [has fixed the problem,](https://github.com/bbc/psammead/pull/3781) now to migrate to `addon-docs`! [Migration docs](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/recipes.md#migrating-from-notesinfo-addons).

---

### 2. Infinite spinner on-load. 
**[FIXED ✅ ]** `6.0.21` 

**tl;dr: This is no longer triggered. 6.0.21 can safely be loaded without an infinite spinner, and without forcing the `"@storybook/core-events": "6.0.2",` resolution.**

---

### 3. Components` required props are undefined
**[FIXED ✅ ]**.  _`6.0.2` & `6.0.21` (with or without forcing `@storybook/core-events` resolutions)._ 

**tl;dr:  'Args' are now the first argument passed into a story, which means our decorators (i.e. withServiceKnob) were not being interpreted correctly as component props. https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#args-passed-as-first-argument-to-story.**

There are two potential fixes for this:

Migrate from 

```js
({ service }) => (
      <Component
         service={service}/>
```

to

```js
( _args, { service }) => (
      <Component
         service={service}/>
```

Or add `passArgsFirst: false,` to `config.js` (which is what I've done)

before: 
Screenshot 2020-09-17 at 16 54 34

after: 
Screenshot 2020-09-22 at 10 19 11

---

### 4. `|` is no longer a valid story hierarchy separator
**[FIXED ✅ ]**  

**tl;dr: replaced `|` with `/`**

We use this throughout. It needs to be replaced by `/`.

**Impact:**
- Component tree visualisation is gubbed.
Screenshot 2020-09-17 at 15 35 21

---

### 5. Arbitrary decorators, and those added after the first story now cause an error
**[FIXED ✅ ]** `6.0.21` 

**tl;dr: removed any arbitrary `addDecorator` funcs in the middle of components**

 i.e. https://github.com/bbc/psammead/blob/2de4871bc90c5be179594b6962ee36037d0f03ff/packages/components/psammead-media-player/src/index.stories.jsx#L130 Although our decoration of stories has tended to be pretty arbitrary, it seems like this component is the only one triggering these errors, so should be a straightforward fix.

**Impact:**
 - Triggers error `You cannot add a decorator after the first story for a kind.` Storybook won't render

---

### Future work:
- **Currently best practice, but soon to be enforced. As outlined [here](https://medium.com/storybookjs/storybook-6-migration-guide-200346241bb5)**
- Do you use storiesOf to write your stories? Upgrade to Component Story Format (CSF). Or choose MDX if you want more control over your documentation.
- Do you use config.js to configure Storybook? Main.js declarative configuration, introduced in 5.3, is a huge improvement and the strong recommendation. Upgrade to main.js.
- Do you use addon-knobs? Storybook Controls is the no-code successor to Knobs that auto-generates the controls based on your component props. Here’s how to migrate your stories to controls.
- Do you use global parameters or decorators? In 6.0 you should add them declaratively; we’ve deprecatedaddParameters/addDecorator.

---

### Testing notes:

- Checkout the `dependabot-storybook` branch locally within Psammead
- Run storybook `npm run storybook`
- Check that there are no issues as you navigate around the components (specifically those above, but keep an eye for any weirdness). Comparing it with https://bbc.github.io/psammead/ will be useful for this

---


karinathomasbbc (approved):

FK78 (commented):
I believe these major bumps have been done in this PR: https://github.com/bbc/psammead/pull/3729

andrewscfc:
Work completed in https://github.com/bbc/psammead/pull/3729

andrewscfc:
I've re-opened this PR following @rhenshaw56 's revert https://github.com/bbc/psammead/pull/3736 ; I am trying to get the upgrade to Storybook 6.0 to work

On `latest` we experienced an infinite loading wheel with a too much recurision message, I have stopped this by forcing `@storybook/core-events` to `6.0.21` https://github.com/bbc/psammead/pull/3728/commits/b5c4bdf644385f91cc7c14d8302bf9085e59a3b6 

This was necessary because [`@storybook/addon-notes`](https://www.npmjs.com/package/@storybook/addon-notes) does not have a storybook 6 compatible verision released; I tried the alpha and that didn't seem to work.

## Next Steps
- Try and remove `@storybook/addon-notes` ? - Is it really needed, it isn't good practice to force resolutions, if not this can be left forced for time being.
- Resolve the next error:
image
  - This should be a bit simpler problem to work out, there are docs here: https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#can-no-longer-add-decorators-parameters-after-stories 
   - We seem to use this deprecated approach quite a lot though so it may be a fairly sweeping change.

RichardPK:
### Update 22/9/20:

PR has moved on quite a bit from this comment. I'll keep the description up to date with where it's at.

---

I've made some changes to get this running a bit more. Sorry for polluting your gloriously clean commit log @andrewscfc. 

The good:

- The infinite spinner seems fixed, without forcing `@storybook/core-events`, and using v 6.0.21, rather than 6.0.2
- Also stopped the decorator error by removing the offender. The decorator causing the error needs re-implementing in a different way
- `addon-notes`' depreciation doesn't appear to cause any breaking problems. It seems OK to leave it in `5.3.21` for now - though I'm not sure if we use it?
The other bug introduced in v6 is our component tree is visually gubbed. This is a straightforward replace `|` separators with `/` though.

The bad (i.e. what I'm still digging into):

- Lots of prop type fails across a range of props & components: `"Failed prop type: The prop 'service' is marked as required in 'Timestamp', but its value is 'undefined'".` 

- A recurring `TypeError: Cannot read property 'debug' of undefined at PostmsgTransport.handleEvent` message.

- These manifest in two major (and possibly interconnected) errors in this version: **Components render irregularly or not at all.**. & **Components aren't re-rendering when navigating around Storybook, or modifying knobs.**

RichardPK:
Bit of debugging alongside @andrewscfc isolated the last remaining problem - which was an erroring logger, caused ultimately by an old version of the depreciated `storybook-readme` package.

The way to fix the error, is to absolutely remove `addons-notes` and `storybook-readme`, which I've documented in this draft PR, for reference: https://github.com/bbc/psammead/pull/3781. 

The next - and hopefully final - step for this PR is to migrate to `addons-docs`.

As ever, I'm keeping this PR description as up-to-date as possible.

RichardPK:
That should be everything fixed. **tl;drs** of each issue resolved included above.

The only two to note going forward: 

1. 'Args' are now the first argument passed into a story, which created undefined props in a lot of our stories. I added a config line: `passArgsFirst: false,` to `config.js` to resolve this. The other solution is to start passing args before props in each of our story components i.e.

```js
( _args, { service }) => (
      <Component
         service={service}/>
```

2. Notes > Docs addon migration is completed in so far as nothing breaks, but docs uses `MDX` and has some behaviour like auto-scrolling to selected components, which it tries to do with our `md` files (badly). We should probably look at proper integration later https://github.com/storybookjs/storybook/tree/next/addons/docs

.yeoman/templates/components/src/index.stories.jsx
```diff
@@ -5,7 +5,7 @@ import { withServicesKnob } from '@bbc/psammead-storybook-helpers';
 import notes from '../README.md';
 import <%= packageName %> from './index';
 
-storiesOf('Components|<%= packageName %>', module)
+storiesOf('Components/<%= packageName %>', module)
```
RichardPK, line 8: All of these `|` > `/` changes are fixing:
> 4. `|` is no longer a valid story hierarchy separator

And so can mostly be ignored to filter the noise. Best way to check these are OK is to run locally and 👀  the component structure

packages/components/psammead-media-player/src/index.stories.jsx
```diff
@@ -21,7 +21,8 @@ const StyledMessageContainer = styled.div`
   overflow: hidden;
 `;
 
-storiesOf('Components|Media Player', module)
+storiesOf('Components/Media Player', module)
+  .addDecorator(withKnobs)
```
RichardPK, line 25: `storiesOf` / `knobs` changes in this file are fixing

> 5. Arbitrary decorators, and those added after the first story now cause an error

.storybook/config.js
```diff
@@ -48,6 +49,7 @@ import '@bbc/psammead-locales/moment/yo';
 import { Helmet } from 'react-helmet';
 
 addParameters({
+  passArgsFirst: false,
```
RichardPK, line 53: Fixes

> 3. Components` required props are undefined 

See that issue for details

.storybook/config.js
```diff
@@ -62,6 +64,16 @@ addParameters({
       iframes: true,
     },
   },
+  docs: {
```
RichardPK, line 68: Part of the fix for

> 1. Components aren't re-rendering when navigating around Storybook, or modifying knobs.

Which is all just removing `addon-notes` and using `addon-docs` instead

packages/utilities/psammead-assets/src/svgs.stories.jsx
```diff
@@ -61,7 +61,7 @@ getSVG.propTypes = {
   height: number,
 };
 
-const stories = storiesOf('Utilities|SVGs/Brand Svgs', module).addDecorator(
+const stories = storiesOf('UTILITIES/SVGS/Brand Svgs', module).addDecorator(
```
karinathomasbbc: Just a minor thing - any reason why Utilities is now all CAPS?

karinathomasbbc (approved):
Looks good? :shipit:

packages/utilities/psammead-assets/src/svgs.stories.jsx
RichardPK: Ah must have been because I was REALLY REALLY ANGRY when I wrote it

No absolutely not, I'll revert. Storybook automatically capitalises component folders, I might have copied and pasted from there 🤷‍♂️

rebeccamcginn (approved):

LilyL0u:
There are some padding differences introduced in this upgrade. Link to issue describing the bug https://github.com/bbc/simorgh/issues/7917

PR #3784, opened 2020-09-24
Web vitals

### LINKED ISSUE #3778: [web-vitals] Create new package for collecting web-vitals
With the introduction of core [web-vitals](https://web.dev/vitals/?gclid=CjwKCAjw-5v7BRAmEiwAJ3DpuGq4yydZBEHoLscEvvrOE5Ef01huTLJebM6jdIWsGKS5L3QeMT5GfhoCNpkQAvD_BwE) form Google we are looking too collect and use this data to help us make product/technical decisions for our web based frontends.

Web-Vitals are key to understanding how readers perceive our pages across the wide array of networks and devices our content is accessed from. These vitals aim to help ensure we deliver the best possible web offering to everyone.

### Core Web-Vitals

- [CLS](https://web.dev/cls/) (Cumulative Layout Shift)
- [LCP](https://web.dev/lcp/) (Largest Contentful Paint)
- [FID](https://web.dev/fid/) (First Input Delay)
- [FCP](https://web.dev/fcp/) (First Contentful Paint)

### Other data (useful for categorising the web-vitals data received)

- [TTFB](https://web.dev/time-to-first-byte/) (Time To First Byte)
- Device Memory
- Device CPU
- Effective Connection Type

Google have a set of utility functions/hooks to collect some if not all of the above metrics: https://www.npmjs.com/package/react-adaptive-hooks

Utilise the [web-vitals](https://github.com/GoogleChrome/web-vitals) library from google and build a utility package/hook to collect the above data and send to an endpoint (tbc) via a post request. This data should be sent to the endpoint before the page unloads of when the reader leaves the page and should only be sent once per page visit.

### This package must:

- Accept a reporting endpoint as a parameter
- Use the [web-vitals](https://github.com/GoogleChrome/web-vitals) library to collect core web vitals
- Collect device hardware info listed above ([React Adaptive Hooks](https://www.npmjs.com/package/react-adaptive-hooks)) can help with this.
- Only send data once per page view
- Implement some form of sampling (we cannot send data for every single page view as this will overwhelm the reporting endpoint and rack up infrastructure costs). A simple way to do this is generate a random number and if the random matches matches say 2 out of the 5 possible numbers, then send the request if does not match then do not send the request.
- If possible calculate the time from. page load to page unload, sending this in milliseconds as the `age` param.
- Accept a list of query params to append to the reporting endpoint, one use case for this so far is to send the pageType e.g. STY, MAP, ARTICLE, INDEX, FIX etc
- It might be worth seeing if we can use the `navigator.sendBeacon()` function to send the beacon when the page unloads (see https://github.com/GoogleChrome/web-vitals#send-the-results-to-an-analytics-endpoint)

[Here](https://github.com/bbc/simorgh/pull/6553) is a POC that shows a very similar package working in Simorgh. We would like this package to be easily reusable by anyone with a react frontend.

DO NOT namespace the package with `psammead-` as this is not a visual frontend component

## Data structure to follow the below format
```
{
    "age": 0, // age here is the time the web page was opened to the time this data was sent, not the readers age.
    "type": "web-vitals",
    "url": "https://www.bbc.com/mundo",
    "body": {
      "cls": 0.002242123597597478,
      "fid": 726.5349999943282,
      "lcp": 726.5349999943282,
      "ttfb": 304.16499997954816,
      "device_mem": 8,
      "device_cpu": 4,
      "device_effective_connection": "3g"
    }
  }
```

N/A

**Testing notes**
[Tester to complete]

**Additional context**

mtwilliams5:
Resolves #3778

**Overall change:** 
Creates a utility hook that can be added to a React app that will collect web vitals and device capability metrics for reporting to an endpoint over https.

**Code changes:**

- Creation of useWebVitals hook
  - Transmit of metrics to reporting endpoint can be toggled off, for legal compliance requirements (this is necessary because hooks should not be called conditionally by the app you use them in)
  - Allows the specifying of a reporting endpoint and a logger callback function as props, to ensure that it can be used in multiple different apps
  - Allows a sample rate percentage to be set, to ensure that backend reporting servers aren't overloaded with requests
  - Allows a reportParams object to be sent, to append query parameters to the reporting url, to enable potential grouping of metrics based on things like page type
- Creation of useEvent hook (placed within web-vitals utility, due to CI packaging limitations. It could be split into a separate utility at a later date)
  - Manages the addition and removal of window-level event handlers in a React Hook-friendly/best-practices way
---


packages/utilities/web-vitals/src/index.js
```diff
@@ -0,0 +1,70 @@
...
+    try {
+      await sendBeacon(beacon, reportingEndpoint);
+    } catch (error) {
+      loggerCallback('web_vitals_send_error', {
+        ...error,
+      });
+    }
+  };
+
+  if (enabled) {
```
karinathomasbbc: I would be inclined to remove this and let the calling app decide whether web-vitals is enabled or not.

packages/utilities/web-vitals/src/index.js
```diff
@@ -0,0 +1,70 @@
...
+    event.preventDefault();
+    const pageExitTime = new Date();
+    const pageAge = pageExitTime - pageLoadTime;
+
+    const beacon = [{ ...webVitalsBase, age: pageAge, body: { ...vitals } }];
+
+    try {
+      await sendBeacon(beacon, reportingEndpoint);
+    } catch (error) {
+      loggerCallback('web_vitals_send_error', {
```
karinathomasbbc: Should we just return the error (either as a string or an object?) to the logger, and omit `web_vitals_send_error`

packages/utilities/web-vitals/src/index.js
mtwilliams5: Yeah, this bit of code is definitely going to change. Main reason I thought I'd pass in enabled as a prop rather than leaving it to the app to decide to call the hook at all was down to the fact that we shouldn't be triggering hooks conditionally (but hooks can have conditional logic inside of them). I figured I'd probably move the logic for it further down since it's currently making it call useEvent and useEffect conditionally within the hook

packages/utilities/web-vitals/src/index.js
mtwilliams5: Yeah, that makes sense. This was largely just be copying over code from the POC and reworking it slightly to not rely on app config

packages/utilities/web-vitals/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Web Vitals Changelog
+
+| Version | Description |
+|---------|-------------|
+| 1.0.0   | [PR#XX](https://github.com/BBC-News/psammead/pull/XX) Initial creation of package. |
```
karinathomasbbc:
```suggestion
| 1.0.0   | [PR#3784](https://github.com/bbc/psammead/pull/3784) Initial creation of package. |
```

packages/utilities/web-vitals/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Description
+
+The `WebVitals` utility provides a configurable hook which will collect device capability metrics (effective network type, device CPU count, and device memory) and Web Vitals metrics (Cumulative Layout Shift, Time To First Byte, etc) for each page view, and then send those metrics to a reporting endpoint when the user leaves the page.
```
karinathomasbbc: Is it worth listing out all of the web-vitals metrics here? And perhaps a link to the web-vitals docs for information (you may be able to copy & paste directly from the issue! https://github.com/bbc/psammead/issues/3778

packages/utilities/web-vitals/README.md
```diff
@@ -0,0 +1,75 @@
...
+| reportParams      | Object   | No       | `undefined` | `{ reportParams: { pageType: 'STY' }` |
+
+## Usage
+
+The hook can be configured in a number of ways:
+
+* `enabled` - setting this to `true` will enable sending these metrics to the reporting endpoint. This is so that the reporting functionality can be turned on only when users have agreed to the appropriate privacy and tracking settings where appropriate.
+* `reportingEndpoint` - the endpoint you wish to send your metrics to
+* `loggerCallback` - a function that will be called whenever an attempt to send the metrics fails.
+* `sampleRate` - providing a number between 1-100 will set the percentage of sampling, to prevent overloading the backend reporting server. e.g. a `sampleRate` of `5` will lead to a 5% change that a given page view's metrics will be sent to the reporting server.
```
karinathomasbbc: Tiny typo - `chance` instead of `change`.

Or perhaps a wee reword - 
```
e.g. a `sampleRate` of `5` will mean that metrics for 5% of requests will be sent to the reporting server.
```

packages/utilities/web-vitals/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@bbc/web-vitals",
+  "version": "1.0.0",
+  "main": "dist/index.js",
+  "module": "esm/index.js",
+  "sideEffects": false,
+  "description": "Generic hook for reporting web vitals metrics to an endpoint",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/BBC-News/psammead/tree/latest/packages/utilities/web-vitals"
```
karinathomasbbc:
```suggestion
    "url": "https://github.com/bbc/psammead/tree/latest/packages/utilities/web-vitals"
```

packages/utilities/web-vitals/package.json
```diff
@@ -0,0 +1,29 @@
...
+  },
+  "author": {
+    "name": "Psammead Maintainers",
+    "email": "PsammeadMaintainers@bbc.co.uk"
+  },
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/bbc/psammead/issues"
+  },
+  "homepage": "https://github.com/BBC-News/psammead/blob/latest/packages/utilities/web-vitals/README.md",
```
karinathomasbbc:
```suggestion
  "homepage": "https://github.com/bbc/psammead/blob/latest/packages/utilities/web-vitals/README.md",
```

karinathomasbbc (approved):

packages/utilities/web-vitals/README.md
mtwilliams5: Yeah, that makes sense. I was mostly trying to dance around the fact that it's not strictly 5% of requests that will get sent to the reporting server; but rather that your particular page view will have a 5% chance of being reported, which we hope to ultimately be as close to the first thing as possible (just down to it being a random number generated client-side). The difference between the two probably doesn't matter enough to complicate the readme with though.

mtwilliams5:
I haven't created a storybook story for this (largely because I figure the UI tests wouldn't be any use to us on this), but creating one and opening that story iframe in a new tab was an effective way of testing the functionality of this in an actual browser.

Steps if anyone wants to reproduce themselves:

Create a story following the below basic outline. You can set the visual element to be whatever you like;

```jsx
import React from 'react';
import { storiesOf } from '@storybook/react';
import useWebVitals from './index';

storiesOf('Utilities|Web Vitals', module).add(
  'basic',
  () => {
    const vitalsConfig = {
      enabled: true,
      reportingEndpoint: 'http://localhost:8080',
      loggerCallback: console.error,
      sampleRate: 100,
    };
    useWebVitals(vitalsConfig);

    return (
      <div>
        <p>Here is some text</p>
      </div>
    );
  },
  { notes },
);
```

Open it in a browser (if you use this `storiesOf` config, then `http://localhost:8180/iframe.html?id=utilities-web-vitals--basic` should be a direct link to the specific story).

To validate that the event fires when expected, go to a different URL within that browser tab. It's best to set your console and network tabs to preserve logs, but if you're running a reporting endpoint locally, you should be able to see the incoming requests there too.

To trigger the same behaviour without actually navigating away, you can use `window.dispatchEvent(new Event('pagehide'));`

packages/utilities/web-vitals/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+The `WebVitals` utility provides a configurable hook which will collect device capability metrics and [Web Vitals](https://web.dev/vitals/?gclid=CjwKCAjw-5v7BRAmEiwAJ3DpuGq4yydZBEHoLscEvvrOE5Ef01huTLJebM6jdIWsGKS5L3QeMT5GfhoCNpkQAvD_BwE) metrics for each page view, and then send those metrics to a reporting endpoint when the user leaves the page.
+
+### Metrics we report on
+
+#### Core Web-Vitals
+
+- [CLS](https://web.dev/cls/) (Cumulative Layout Shift)
+- [LCP](https://web.dev/lcp/) (Largest Contentful Paint)
+- [FID](https://web.dev/fid/) (First Input Delay)
```
JoshTumath, line 13: Pedantic alert ⚠️ : FID is a 'web vital' but not one of the three 'core web vitals'.

packages/utilities/web-vitals/src/index.js
```diff
@@ -0,0 +1,130 @@
...
+  const sendVitals = async () => {
+    const pageExitTime = Date.now();
+    const pageAge = pageExitTime - pageLoadTime;
+
+    const beacon = [
+      { ...webVitalsBase, age: pageAge, body: { ...vitals, ...deviceMetrics } },
+    ];
+
+    try {
+      if (shouldSample(sampleRate)) {
```
JoshTumath: To save on unnecessary computation, can the `shouldSample` function be run at the very beginning of the hook? There's no point in collecting the data and setting up an event listener if we're not going to send it.

JoshTumath (approved):
Awesome! I just have a comment about the sampling...

packages/utilities/web-vitals/README.md
mtwilliams5: I copied that lot from the [issue](https://github.com/bbc/psammead/issues/3778) - easy change though

packages/utilities/web-vitals/src/index.js
mtwilliams5: Could do. I mostly put it here because I figured it wasn't worth doing the shouldSample if `enabled` is `false`. I could refactor it to do it at the start of the hook but only if `enabled` is `true`, I suppose.

packages/utilities/web-vitals/README.md
mtwilliams5: Just double checked, FID is a core vital; it's FCP that isn't.

JoshTumath: D'oh! Sorry.

packages/utilities/web-vitals/src/index.js
```diff
@@ -0,0 +1,130 @@
...
+    updateDeviceMetrics({
+      effectiveConnectionType,
+      numberOfLogicalProcessors,
+      deviceMemory,
+    });
+    getCLS(updateWebVitals);
+    getFID(updateWebVitals);
+    getLCP(updateWebVitals);
+    getFCP(updateWebVitals);
+    getTTFB(updateWebVitals);
```
JoshTumath, line 124: Is it possible to also check for `shouldSendVitals` before even collecting the data?

mtwilliams5: We could do, but I'm worried that may mean we miss the CLS generated by the cookie banner (since `enabled` will be `false` until the cookie banner has been interacted with). Capturing the vitals from the start but not sending them anywhere unless the user accepts the relevant permissions should hopefully mean we can get it to capture that CLS.
This'll likely be something that we'll only be able to play around with and find out/tweak once it's able to be integrated into either webcore or simorgh

JoshTumath: Oh good point. Depends on how the internal Web APIs work to get the data. They could be collecting the data anyway.

chris-hinds (approved):
This is some great work. Looking forward to testing this in both Simorgh and WebCore 🥳

PR #3832, opened 2020-10-06
Storybook updates to support Reith Qalam

### LINKED ISSUE #3800: Add Reith Qalam font typography
Add Qalam font typography

- Add new constants including font sizes & line heights for Reith Qalam to https://github.com/bbc/psammead/blob/latest/packages/utilities/gel-foundations/src/typography.js (described in the Designs -> Line Heights spec within JIRA NEWSWORLDSERVICE-1104)
- This should be a major version bump, to prevent automated dependency updates 

**Testing notes**
[Tester to complete]

**Additional context**

karinathomasbbc:
Resolves #3800 

**Overall change:** 
Upgrade gel-foundations library to use Reith Qalam font

**Code changes:**
- Add Reith Qalam fonts to Storybook static assets
- package, package-lock & changelog updates

---


simonsinclair (commented):
Are we unable to use the hosted version of these fonts? My concern is that they get updated over time, especially during the release stages, and we might be duplicating effort hosting them in multiple places.

**Edit:**
I can see there is already [logic in place to use all web fonts locally](https://github.com/bbc/psammead/blob/f4aeb93f6210deb8ee4cc93108332ccf8b4c11ac/.storybook/preview.js#L93-L100), so this is probably not one for this PR.

simonsinclair (approved):

karinathomasbbc:
> Are we unable to use the hosted version of these fonts? My concern is that they get updated over time, especially during the release stages, and we might be duplicating effort hosting them in multiple places.
> 
> **Edit:**
> I can see there is already [logic in place to use all web fonts locally](https://github.com/bbc/psammead/blob/f4aeb93f6210deb8ee4cc93108332ccf8b4c11ac/.storybook/preview.js#L93-L100), so this is probably not one for this PR.

Yeah I agree with you - this is not ideal. I will raise an issue to address this - the same could be said for all of those fonts - it't not ideal having them in multiple places.

FK78 (approved):

PR #3853, opened 2020-10-12
DPUB-1466: Add numbered list component

leostaples:
Resolves https://jira.dev.bbc.co.uk/browse/DPUB-1466

**Overall change:** Add numbered list component

**Code changes:**

---


packages/components/psammead-numbered-list/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,101 @@
...
+}
+
+@media (min-width:37.5rem) {
+  .c0 {
+    font-size: 1.375rem;
+    line-height: 2rem;
+  }
+}
+
+<ol
```
greenc05, line 81: Not sure if you are adding your own list numbers as per https://bbc.github.io/psammead/?path=/story/components-mostread-list--default or if you are using the default CSS list ordering numbers, if adding your own numbers you may need an attribute of role="list" on the ul so VoiceOver Mac and iOS users get list semantics.

leostaples: This was one of the issues I wanted to discuss. I saw the Most Read implementation but I noticed the translations only go up to 10, so not sure how we'd handle longer lists.  Can we not rely on browsers to provide correct numerals or do they not translate them? Happy to discuss on Slack if that's easier.

greenc05: Think we add our numbers as we wanted different styling for the numbers to the text, that's not possible if you use standard CSS list numbers...

greenc05: If you don't need different styling I would just use the standard CSS list numbers provided they work ok with AT...

packages/components/psammead-numbered-list/src/index.jsx
```diff
@@ -0,0 +1,28 @@
+import styled from 'styled-components';
```
ryanmccombe: We've switched styled-components for Emotion

The rest of your code does seem compatible with this change - just need to update the import 🙂 
```suggestion
import styled from '@emotion/styled';
```

packages/components/psammead-numbered-list/src/index.jsx
```diff
@@ -0,0 +1,28 @@
...
+import { string, oneOf, shape } from 'prop-types';
+import { getBodyCopy } from '@bbc/gel-foundations/typography';
```
ryanmccombe, line 3: Can you list the dependencies in the `dependencies` or `peerDependencies` sections of `package.json`

aeroplanejane:
Closing for now

PR #3866, opened 2020-10-13
Switch Web Vitals utility to use Promise syntax instead of Async/Await

mtwilliams5:
**Overall change:**
Switch Web Vitals utility to use Promise syntax instead of Async/Await. This should hopefully resolve an issue when attempting to integrate it into apps where a runtime error occurs referring to `regeneratorRuntime`, e.g.:

```
Uncaught ReferenceError: regeneratorRuntime is not defined
    sendVitals index.js:114
    useWebVitals index.js:150
```

We believe this happens because async functions defined within other functions are hoisted in such a way by Babel that they are then missing the `regeneratorRuntime` that they need to function (based annecdotally on that issue going away in previous projects when we altered our Babel config to not target browsers that don't support native async, and by that particular error going away when npm linking this utility after the change)

**Code changes:**

* Altered the `sendVitals` function to no longer be an async function, and instead of `try { await func() } catch (error) { log(error) }`, use promise-based chaining, e.g.: `func().catch(error => log(error))`
* Altered the `sendBeacon` function to ensure that a Promise is always returned (`navigator.sendBeacon` just returns a boolean, whereas `fetch` returns a promise already), to ensure that the change in `sendVitals` works with as little additional logic as possible
* Updated tests to reflect the above changes, including updating the mocked return values for `navigator.sendBeacon` to be more representative of the function's actual API.

---


mtwilliams5:
Marked PR as requiring manual testing because I still can't get it working in another app while npm linking; I get the same invalid use of hooks error that we got last time, that we think comes from it using a different local React than the app itself does. So it may still be a case of getting this version into NPM to then pull down into our apps to be able to test it there.

packages/utilities/web-vitals/CHANGELOG.md
```diff
@@ -3,3 +3,4 @@
 | Version | Description |
 |---------|-------------|
 | 1.0.0   | [PR#3784](https://github.com/bbc/psammead/pull/3784) Initial creation of package. |
+| 1.0.1   | [PR#3866](https://github.com/bbc/psammead/pull/3866) Rewrite async function to prevent regeneratorRuntime errors |
```
ryanmccombe:
```suggestion
| 1.0.1   | [PR#3866](https://github.com/bbc/psammead/pull/3866) Rewrite async function to prevent regeneratorRuntime errors |
| 1.0.0   | [PR#3784](https://github.com/bbc/psammead/pull/3784) Initial creation of package. |
```

ryanmccombe (approved):

chris-hinds (approved):

PR #3960, opened 2020-10-26
Ensure we monitor all events for getCLS and getLCP

chris-hinds:
Resolves #NUMBER

**Overall change:**

- Sends the extra `true` param for the `getCLS()` and `getLCP()` functions to ensure we update the values on all events fired.

- Calls the `getCLS()` function one more time just before sending the beacon to ensure we have the most up to date value.

---

### DIFF
--- a/packages/utilities/web-vitals/src/index.js
+++ b/packages/utilities/web-vitals/src/index.js
@@ -100,6 +100,9 @@ const useWebVitals = ({
     const pageExitTime = Date.now();
     const pageAge = pageExitTime - pageLoadTime;
 
+    // Last chance to get the CLS before sending the beacon.
+    getCLS(updateWebVitals, true);
+
     const beacon = [
       { ...webVitalsBase, age: pageAge, body: { ...vitals, ...deviceMetrics } },
     ];
--- a/packages/utilities/web-vitals/src/index.js
+++ b/packages/utilities/web-vitals/src/index.js
@@ -117,9 +117,9 @@ const useWebVitals = ({
       numberOfLogicalProcessors,
       deviceMemory,
     });
-    getCLS(updateWebVitals);
+    getCLS(updateWebVitals, true); // Setting 'true' will report all CLS changes
     getFID(updateWebVitals);
-    getLCP(updateWebVitals);
+    getLCP(updateWebVitals, true); // Setting 'true' will report all LCP changes
     getFCP(updateWebVitals);
     getTTFB(updateWebVitals);
   }, []);
--- a/packages/utilities/web-vitals/CHANGELOG.md
+++ b/packages/utilities/web-vitals/CHANGELOG.md
@@ -2,5 +2,6 @@
 
 | Version | Description |
 |---------|-------------|
+| 1.0.2   | [PR#3960](https://github.com/bbc/psammead/pull/3960) Add extra prop to always report on CLS and LCP |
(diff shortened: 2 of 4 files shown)


packages/utilities/web-vitals/src/index.js
```diff
@@ -117,9 +120,9 @@ const useWebVitals = ({
       numberOfLogicalProcessors,
       deviceMemory,
     });
-    getCLS(updateWebVitals);
+    getCLS(updateWebVitals, true); // Setting 'true' will report all CLS changes
```
karinathomasbbc, line 123: Do we have / need tests for this?

karinathomasbbc (approved):
Looks good :shipit:

packages/utilities/web-vitals/src/index.js
chris-hinds: there's tests already in place to ensure this function is called. Not 100% sure we need to update the tests for this though

karinathomasbbc: Could we update the test make sure it has been called with `true` as a param?

packages/utilities/web-vitals/src/index.js
```diff
@@ -100,6 +100,9 @@ const useWebVitals = ({
     const pageExitTime = Date.now();
     const pageAge = pageExitTime - pageLoadTime;
 
+    // Last chance to get the CLS before sending the beacon.
+    getCLS(updateWebVitals, true);
```
JoshTumath, line 104: I always get really confused by how these web-vitals lib functions work, because my understanding is the function will automatically call the `updateWebVitals` callback more than once. So _in theory_, calling the `getCLS` function twice won't make a difference. 🤔 But I'm guessing you've found that it actually does?

> This callback will fire any time either:
>
> * The final value of the metric has been determined.
> * The current metric value needs to be reported right away (due to the page being unloaded or backgrounded).

JoshTumath (approved):

packages/utilities/web-vitals/src/index.js
chris-hinds: Was just going off of @mtwilliams5  idea to call it twice. Perhaps its not actually needed though

mtwilliams5: So basically what will happen is as soon as the hook gets called (i.e. the component that uses it gets rendered), we'll call the `getCLS` function, and tell it to give us what the value for CLS is right there and then. But the CLS figure is constantly updating, so we will want to know what it is when the user leaves the page. We can get that by making the CLS call again within just before we send the beacon off, or by moving all of those `getX` calls to that point.

Doing that second one may not be as bad of a thing now that I understand a bit better how these are working - they're asking the browser for the figures, rather than asking the browser to start recording the figures - but I'm not sure if there's any risk of them not returning properly before the page gets unloaded if we're running all of them immediately before the sendBeacon call. The actual behaviour of the javascript while the page is actively unloading is still a bit of an unknown to me, in terms of if it takes a while to finish what it's doing, will it just kill it off, or let the process hang around for a bit? It was that concern that made me think that maybe asking for it at the start then updating it when the user leaves could be the safer option.

mtwilliams5 (approved):

packages/utilities/web-vitals/src/index.js
chris-hinds: I'll do that in a separate PR as it involves mocking the updateWebVitals function which currently isnt done yet

PR #3979, opened 2020-11-05
[Alpha] Update Episode List to support Videos and Dark Mode

ryanmccombe:
Part of https://github.com/bbc/simorgh/issues/8304

**Overall change:**
Update the Psammead Episode List Component to support images and dark mode

**Code changes:**

- Remove external padding/margins/line heights
- Add a new `EpisodeList.Image` component
- Update the `EpisodeList.Episode` component to render differently if passed an `EpisodeList.Image`
- Add a `darkMode` prop
- Update the audio example fixture to better implement a11y criteria

---

### DIFF
--- a/packages/components/psammead-episode-list/CHANGELOG.md
+++ b/packages/components/psammead-episode-list/CHANGELOG.md
@@ -3,6 +3,6 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
-| 0.1.0-alpha.3 | [PR#3979](https://github.com/bbc/psammead/pull/3979) Updating media icon styling |
+| 0.1.0-alpha.3 | [PR#3979](https://github.com/bbc/psammead/pull/3979) Implement episode image component and dark mode support |
 | 0.1.0-alpha.2 | [PR#3962](https://github.com/bbc/psammead/pull/3962) Updating media icon styling |
 | 0.1.0-alpha.1 | [PR#3837](https://github.com/bbc/psammead/pull/3837) Initial creation of package. |
--- a/packages/components/psammead-episode-list/CHANGELOG.md
+++ b/packages/components/psammead-episode-list/CHANGELOG.md
@@ -3,5 +3,6 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 0.1.0-alpha.3 | [PR#3979](https://github.com/bbc/psammead/pull/3979) Updating media icon styling |
 | 0.1.0-alpha.2 | [PR#3962](https://github.com/bbc/psammead/pull/3962) Updating media icon styling |
 | 0.1.0-alpha.1 | [PR#3837](https://github.com/bbc/psammead/pull/3837) Initial creation of package. |
--- a/packages/components/psammead-episode-list/src/Episode.jsx
+++ b/packages/components/psammead-episode-list/src/Episode.jsx
@@ -15,7 +15,7 @@ import Image from './Image';
 const Wrapper = styled.div`
   position: relative;
   &:hover {
-    .underlined_hover {
+    [class*='--hover'] {
       text-decoration: underline;
     }
     .rounded-play-button__outer-circle,
--- a/packages/components/psammead-episode-list/src/Episode.jsx
+++ b/packages/components/psammead-episode-list/src/Episode.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import styled from '@emotion/styled';
-import { arrayOf, element, string } from 'prop-types';
+import { node, string } from 'prop-types';
 import {
   GEL_SPACING_DBL,
   GEL_SPACING_QUIN,
@@ -11,11 +11,24 @@ import {
 } from '@bbc/gel-foundations/breakpoints';
 import tail from 'ramda/src/tail';
 import pathOr from 'ramda/src/pathOr';
+import { C_POSTBOX, C_WHITE } from '@bbc/psammead-styles/colours';
 import MediaIndicator from './MediaIndicator';
 import Image from './Image';
 
 const Wrapper = styled.div`
   position: relative;
+  &:hover {
+    .underlined_hover {
+      text-decoration: underline;
+    }
+    .rounded-play-button__outer-circle,
+    .rounded-play-button__inner-circle {
+      fill: ${C_POSTBOX};
+    }
+    .rounded-play-button__triangle {
+      fill: ${C_WHITE};
+    }
+  }
 `;
 
 const TextWrapper = styled.div`
@@ -53,12 +66,8 @@ const Episode = ({ children, dir }) => {
 };
 
 Episode.propTypes = {
-  children: arrayOf(element),
+  children: node.isRequired,
   dir: string.isRequired,
-};
-
-Episode.defaultProps = {
-  children: [],
 };
 
 export default Episode;
--- a/packages/components/psammead-episode-list/src/Episode.jsx
+++ b/packages/components/psammead-episode-list/src/Episode.jsx
@@ -13,7 +13,7 @@ import Image from './Image';
 const Wrapper = styled.div`
   display: inline-block;
   max-width: calc(
-    100% - ${({ sideBarWidth }) => sideBarWidth}px - ${GEL_SPACING_DBL}
+    100% - ${({ narrow }) => (narrow ? 230 : 50)}px - ${GEL_SPACING_DBL}
   );
   vertical-align: top;
 `;
@@ -27,7 +27,7 @@ const Episode = ({ children, dir }) => {
       ) : (
         <MediaIndicator size={GEL_SPACING_QUIN} dir={dir} />
       )}
-      <Wrapper sideBarWidth={firstChildIsImage ? 230 : 50}>
+      <Wrapper narrow={firstChildIsImage}>
         {firstChildIsImage ? tail(children) : children}
       </Wrapper>
     </>
--- a/packages/components/psammead-episode-list/src/Episode.jsx
+++ b/packages/components/psammead-episode-list/src/Episode.jsx
@@ -2,19 +2,32 @@ import React from 'react';
 import styled from '@emotion/styled';
 import { arrayOf, element, string } from 'prop-types';
 import { GEL_SPACING_QUIN } from '@bbc/gel-foundations/spacings';
+import tail from 'ramda/src/tail';
+import is from 'ramda/src/is';
+import isEmpty from 'ramda/src/isEmpty';
+import pathOr from 'ramda/src/pathOr';
 import MediaIndicator from './MediaIndicator';
+import Image from './Image';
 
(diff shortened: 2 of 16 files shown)


packages/components/psammead-episode-list/CHANGELOG.md
```diff
@@ -3,5 +3,6 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 0.1.0-alpha.3 | [PR#3979](https://github.com/bbc/psammead/pull/3979) Updating media icon styling |
```
ryanmccombe:
```suggestion
| 0.1.0-alpha.3 | [PR#3979](https://github.com/bbc/psammead/pull/3979) Implement episode image component and dark mode support |
```

packages/components/psammead-episode-list/src/Image.jsx
```diff
@@ -0,0 +1,105 @@
...
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    width: 230px;
+  }
+`;
+
+const EpisodeImage = props => {
+  const { duration, alt } = props;
+
+  // This component only uses a subset of its props
+  // the remaining props are passed down to the underyling <img> element
```
ryanmccombe:
```suggestion
  // the remaining props are passed down to the underlying <img> element
```

packages/components/psammead-episode-list/src/Episode.jsx
```diff
@@ -1,20 +1,52 @@
...
+  @media (min-width: ${GEL_GROUP_2_SCREEN_WIDTH_MIN}) {
+    max-width: calc(
+      100% - ${({ narrow }) => (narrow ? 120 : 50)}px - ${GEL_SPACING_DBL}
+    );
+  }
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    max-width: calc(
+      100% - ${({ narrow }) => (narrow ? 230 : 50)}px - ${GEL_SPACING_DBL}
+    );
+  }
```
ryanmccombe, line 32: There must be a better way of doing this

packages/components/psammead-episode-list/src/fixtures.jsx
```diff
@@ -43,75 +55,201 @@ export const exampleEpisodes = [
...
+      {...(withSurroundingComponents ? { script, service, dir, darkMode } : {})}
+    >
+      <EpisodeList script={script} service={service} dir={dir}>
+        {episodes.map(episode => (
+          <EpisodeList.Link key={episode.id} href={episode.url}>
+            <EpisodeList.Episode>
+              {/* eslint-disable-next-line jsx-a11y/aria-role */}
+              <span role="text">
+                <VisuallyHiddenText>Audio, </VisuallyHiddenText>
+                <EpisodeList.Title className="underlined_hover fade_visited">
```
ryanmccombe: These magic class names could be documented or replaced with props

packages/components/psammead-episode-list/src/helpers.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+// Used to make service, script and dir passed to <EpisodeList> available to children
```
ryanmccombe:
```suggestion
// Used to make props passed to <EpisodeList> available to children
```
Current limitation here where the `Episode` component cannot be used in isolation - can only be used as a child of `EpisodeList` due to the way context is implemented - could be resolved later

packages/components/psammead-episode-list/src/index.test.jsx
```diff
@@ -3,36 +3,61 @@ import { render } from '@testing-library/react';
...
 
 describe('Episode List ', () => {
   it('should render the list', () => {
     const { getByRole } = render(
-      renderEpisodes(exampleEpisodes, scripts.latin, 'news', 'ltr'),
+      renderEpisodes({
+        episodes: exampleEpisodes,
+        script: scripts.latin,
+        service: 'news',
+        dir: 'ltr',
```
ryanmccombe, line 20: Update `renderEpisodes` to use these settings by default, so they only need provided if they're being overridden

packages/components/psammead-episode-list/src/fixtures.jsx
```diff
@@ -43,75 +55,201 @@ export const exampleEpisodes = [
...
+              </span>
+            </EpisodeList.Episode>
+          </EpisodeList.Link>
+        ))}
+      </EpisodeList>
+    </Wrapper>
+  );
+};
+
+export const renderVideoEpisodes = ({
```
ryanmccombe, line 207: Need to provide an example implementation of the a11y requirements here

May also negate this function when done - could end up being very similar to the function for rendering audio episodes

packages/components/psammead-episode-list/src/Image.jsx
```diff
@@ -0,0 +1,105 @@
...
+  background-color: ${C_EBON};
+  padding: ${GEL_SPACING_HLF};
+  @media (min-width: ${GEL_GROUP_2_SCREEN_WIDTH_MIN}) {
+    padding: ${GEL_SPACING};
+  }
+
+  svg {
+    margin: 0 0 2px 0;
+    height: 10px;
+    width: 12px;
```
rebeccamcginn: Need to remove px and maybe change white to GEL_WHITE_HORSES

packages/components/psammead-episode-list/src/Link.jsx
```diff
@@ -1,8 +1,27 @@
...
+  C_POSTBOX,
+  C_WHITE,
+  C_STONE,
+} from '@bbc/psammead-styles/colours';
 
 const Link = styled.a`
-  display: block;
+  :before {
+    position: absolute;
+    inset: 0;
```
jroebu14: what does this do?

packages/components/psammead-episode-list/src/Link.jsx
```diff
@@ -1,8 +1,27 @@
...
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    content: '';
+    white-space: nowrap;
```
jroebu14: is this needed?

packages/components/psammead-episode-list/src/index.jsx
```diff
@@ -69,18 +61,22 @@ EpisodeList.propTypes = {
...
+  darkMode: bool,
 };
 
 EpisodeList.defaultProps = {
   children: [],
   dir: 'ltr',
+  darkMode: false,
 };
 
 // This module also has a range of supplemental components to provide consumers with some compositational control
```
jroebu14: I'm not sure what this sentence means. Also has a spelling mistake `compositational`. Is it basically saying this component is a compound component?

packages/components/psammead-episode-list/src/helpers.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+// Used to make props passed to <EpisodeList> available to children
```
jroebu14, line 3: which props need passed to all children? Is it `darkMode`? Wondering if there should be a Psammead Theme Context that all Psammead components can use and Simorgh can import the context provider.

packages/components/psammead-episode-list/src/Image.jsx
```diff
@@ -0,0 +1,105 @@
...
+    width: 230px;
+  }
+`;
+
+const EpisodeImage = props => {
+  const { duration, alt } = props;
+
+  // This component only uses a subset of its props
+  // the remaining props are passed down to the underlying <img> element
+  const selectImgProps = omit(['alt', 'duration']);
```
jroebu14, line 82: would being explicit here help us know which props StyledImage needs?

packages/components/psammead-episode-list/src/fixtures.jsx
jroebu14: if we use class names should we maybe use BEM naming convention? e.g. `episode-list__title--hover` and `episode-list__title--visited`

packages/components/psammead-episode-list/src/Image.jsx
```diff
@@ -0,0 +1,105 @@
...
+  GEL_GROUP_2_SCREEN_WIDTH_MIN,
+  GEL_GROUP_3_SCREEN_WIDTH_MIN,
+} from '@bbc/gel-foundations/breakpoints';
+
+import { withEpisodeContext } from './helpers';
+
+const Wrapper = styled.div`
+  display: inline-block;
+  position: relative;
+  line-height: 0;
```
jroebu14: what does this do?

packages/components/psammead-episode-list/src/Link.jsx
ryanmccombe: Was on Catherine's HTML - seems to be a shorthand way of setting `top`, `bottom`, `left` and `right` - https://developer.mozilla.org/en-US/docs/Web/CSS/inset

Didn't work on all browsers though, so I had to set `top`, `bottom`, `left` and `right` anyway - will remove the `inset`

packages/components/psammead-episode-list/src/Link.jsx
ryanmccombe: Doesn't seem to have any effect - removed 👍

packages/components/psammead-episode-list/src/index.jsx
ryanmccombe: Removed comment - seems unnecessary given the examples in the readme

packages/components/psammead-episode-list/src/helpers.jsx
ryanmccombe: `darkMode`, `service`, `script` and `dir`

I think there should be a psammead context that passes these down - especially `service`, `script` and `dir` rather than passing it into every psammead component - not sure if it's in scope for this though - already a pretty big PR

packages/components/psammead-episode-list/src/Image.jsx
ryanmccombe: Not sure what you mean here - `StyledImage` has no required props - the expected props are in the `propTypes`

Or do you mean converting this to a `pick` instead of an `omit`?  This would be picking attributes that are valid for a base HTML `` - there would be dozens of options there, including wildcards (`data-*`)

packages/components/psammead-episode-list/src/Image.jsx
ryanmccombe: Deleted this 👍  Was originally to remove additional spacing below the image, but added the `line-height: 0`  to the list item instead

packages/components/psammead-episode-list/src/helpers.jsx
jroebu14: yeh sorry was just an idea and definitely not in scope for this PR

jroebu14 (approved):

rebeccamcginn (approved):

PR #4002, opened 2020-11-18
Replace instances of BBC Nassim with BBC Reith Qalam

simonsinclair:
No issue.

**Overall change:** 

This PR updates `psammead-styles` and replaces instances of BBC Nassim with BBC Reith Qalam.

**Code changes:**

- Add BBC Reith Qalam font-face rules
- Remove BBC Nassim font-face rules
- Apply BBC Reith Qalam to services previously using BBC Nassim
- Update relevant snapshots
- Bump major package version
- Add changelog entry

---


packages/utilities/psammead-styles/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 7.0.0 | [PR#????](https://github.com/bbc/psammead/pull/????) Replace BBC Nassim with BBC Reith Qalam. |
```
simonsinclair:
```suggestion
| 7.0.0 | [PR#4002](https://github.com/bbc/psammead/pull/4002) Replace BBC Nassim with BBC Reith Qalam. |
```

packages/utilities/psammead-styles/src/font-families.js
```diff
@@ -369,7 +364,7 @@ const optimoBaseFontStyles = {
 export const afaanoromoo = helmetFontStyles;
 export const afrique = helmetFontStyles;
 export const amharic = amharicStyles;
-export const arabic = nassimFontStyles(nassimArabicFontFamily);
+export const arabic = reithQalamStyles;
```
karinathomasbbc, line 367: Are we going to refactor / remove this, or will that be a separate PR?

packages/utilities/psammead-styles/CHANGELOG.md
karinathomasbbc: This is where I got into trouble before - should this be an alpha release?

karinathomasbbc (commented):
Do we need to update gel-foundations before we do this - the font heights & weights need to be updated?

packages/utilities/psammead-styles/CHANGELOG.md
simonsinclair: Replacing BBC Nassim with BBC Reith Qalam is a breaking change, so the major version has been incremented. Unless there is a specific reason to add an alpha tag(?), this should be enough.

packages/utilities/psammead-styles/src/font-families.js
simonsinclair: I explored this idea, but the use of this list is systemic and would involve a lot of manual work in both Psammead and Simorgh. 😞

simonsinclair:
> Do we need to update gel-foundations before we do this - the font heights & weights need to be updated?

`gel-foundations` can be updated following these changes (or after) - neither packages depend on one another. They will be released as major versions, which will require someone to willingly install them.

karinathomasbbc (approved):

andrewscfc (approved):

PR #4005, opened 2020-11-19
Podcast Promos

### LINKED ISSUE #3987: Create a podcast promo component
We need to promote our podcast content.

A podcast promo component to display on our pages.

View the designs here - https://zpl.io/awvWplJ

### Acceptance criteria

**Wrapper**
- should render the wrapping section element with role=region attribute
- should render the section header/label

**Title**
- should render the title text in a h3 element
- screenreaders should read out "[Brand Title], Episodes" (using visually hidden text)
- should render the link inside the h3 element and should wrap the title text
- link element should use psuedo element to make full promo clickable

**Description**
- should render the description in a paragraph element

**Image**
- should render the brand image in a figure element
- should render the brand image alt text
- should be a responsive image using correct srcset and sizes attributes
- image should lazy load and render without JS
- image should have some placeholder to prevent content shifting

**Call to action**
- should render the "Episodes" call to action in a paragraph element
- call to action should be hidden to screenreaders

**SVGs**
- Icons should be SVG and be added to psammead-assets
- SVGs should change colour when user specifies colour preferences
- SVGs should use tabindex attribute to ensure the icon is not focusable in the tabbing order (IE 11)

**Questions**
- some Qs around whether this is truncated on smaller devices or uses different data values (medium/short) that are hidden based on screen size
- Ask Ed for SVG assets then attach here

**Jira ticket** - https://jira.dev.bbc.co.uk/browse/WSMEDIA-295

**Testing notes**
[Tester to complete]

**Additional context**

ryanmccombe:
Requires: https://github.com/bbc/psammead/pull/4009
Resolves https://github.com/bbc/psammead/issues/3987

**Overall change:** Implement a new `psammead-podcast-promo` package.

**Code changes:**

- All code is new.  See README.md for overview - https://github.com/bbc/psammead/blob/eb1eb6b8c540c1fa693229f2f6a3df16339aad49/packages/components/psammead-podcast-promo/README.md

---


packages/components/psammead-podcast-promo/src/components/card-call-to-action.jsx
```diff
@@ -0,0 +1,37 @@
...
+  color: ${C_METAL};
+`;
+
+const StyledSeriesStackIcon = styled(SeriesStackIcon)`
+  width: 12px;
+  color: ${C_METAL};
+  fill: currentColor;
+  position: relative;
+  top: 2px;
+  margin-right: 8px;
```
rebeccamcginn: px will need changing

packages/components/psammead-podcast-promo/src/components/card-image.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+import ImagePlaceholder from '@bbc/psammead-image-placeholder';
+
+const Wrapper = styled.figure`
+  width: 88px;
+  display: inline-block;
+  vertical-align: top;
+  margin: 8px 0 0 8px;
+  @media (min-width: ${GEL_GROUP_2_SCREEN_WIDTH_MIN}) {
+    width: 109px;
```
rebeccamcginn, line 15: Are these from hovering over Zeplin designs?

ryanmccombe: aye

packages/components/psammead-podcast-promo/src/components/card-image.jsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    width: 178px;
+    margin: 0;
+  }
+  @media (min-width: ${GEL_GROUP_4_SCREEN_WIDTH_MIN}) {
+    width: 100%;
+    margin: 0;
+  }
+  img {
```
jroebu14: should we avoid nested selectors where we can?

packages/components/psammead-podcast-promo/src/components/card-link.jsx
```diff
@@ -0,0 +1,29 @@
...
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    content: '';
+    overflow: hidden;
+    z-index: 1;
+  }
+  &:visited {
+    .podcast-promo__visited {
```
jroebu14: should it be `.podcast-promo--visited` since it's a modifier class?

packages/components/psammead-podcast-promo/src/components/title.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+import {
+  GEL_SPACING_HLF,
+  GEL_SPACING_DBL,
+} from '@bbc/gel-foundations/spacings';
+import { getGreatPrimer } from '@bbc/gel-foundations/typography';
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+import { C_EBON } from '@bbc/psammead-styles/colours';
+
+import PodcastIcon from '../icons/podcast';
```
jroebu14, line 10: should we have icons in psammead assets?

packages/components/psammead-podcast-promo/src/index.test.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { shouldMatchSnapshot } from '@bbc/psammead-test-helpers';
+import * as scripts from '@bbc/gel-foundations/scripts';
+
+import BasicExample from './examples/basic';
+
+describe('Podcast Promo', () => {
```
jroebu14: could we have unit tests for some of the acceptance criteria especially the a11y stuff?

packages/components/psammead-podcast-promo/src/components/card-description.jsx
```diff
@@ -0,0 +1,15 @@
+import styled from '@emotion/styled';
+import { getLongPrimer } from '@bbc/gel-foundations/typography';
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+import { C_METAL } from '@bbc/psammead-styles/colours';
+import { GEL_SPACING } from '@bbc/gel-foundations/spacings';
+
+const CardDescription = styled.p`
+  ${({ script }) => getLongPrimer(script)};
+  ${({ service }) => getSansRegular(service)};
+  max-width: 30rem;
```
jroebu14, line 10: should all other px measurements use rem?

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,355 @@
...
+<div>
+  <section
+    class="emotion-26 emotion-27"
+    role="region"
+  >
+    <div
+      class="emotion-4 emotion-5"
+    >
+      <svg
+        class=" emotion-0 emotion-1"
```
jroebu14: I don't see the tabindex requirement for ie11

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,355 @@
...
+    >
+      <figure
+        class="emotion-8 emotion-9"
+      >
+        <div
+          class="emotion-6 emotion-7"
+        >
+          <img
+            alt="Что это было?"
+            src="http://ichef.bbci.co.uk/images/ic/3000x3000/p0776f5z.jpg"
```
jroebu14, line 516: should we use a responsive and lazy loaded image?

packages/components/psammead-podcast-promo/src/components/card-call-to-action.jsx
ryanmccombe: Updated

packages/components/psammead-podcast-promo/src/components/card-image.jsx
ryanmccombe: 👍 Updated `img` to use its own styled component.  Can you explain (or link something that explains) why this is better?

packages/components/psammead-podcast-promo/src/components/card-link.jsx
ryanmccombe: 👍 Updated

packages/components/psammead-podcast-promo/src/components/card-description.jsx
ryanmccombe: I don't think so?  If people wanted to make everything bigger, they would use the page zoom feature.  REM should only be used in situations where we want things to change size based on the user's selected text size

We do use REM for other things too though - I don't really know what the BBC / Simorgh guidance is for when to use REM and when to use PX

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
ryanmccombe: 👍  is that requirement asking for a tabIndex of `-1`?  I added that - no other SVGs in psammead seem to have it though

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
ryanmccombe: API lets consumer pass their attributes to this `img` - `srcset` will be on them if they want to use it

re lazy loading, do we have examples of that being done already, or is this a new feature?

packages/components/psammead-podcast-promo/src/index.test.jsx
ryanmccombe: Only a11y thing implemented on the psammead side is aria-hiding the call-to-action - added test for that, but I'm not even sure we should be aria-hiding content in psammead

I know one snapshot test looks really barren here, but there's just no business logic in this component (not even a conditional) and I think snapshots are the best way of testing CSS and HTML attributes

packages/components/psammead-podcast-promo/src/components/card-image.jsx
jroebu14: Nesting increases specificity and means we can't restructure HTML without worrying about messing up styles. Also it's better for performance if not nested since browsers evaluate selectors from right to left. In this example, all img elements would be selected first and then the figure element. Element selectors are not great either especially if we decided to use a picture element instead of an img element

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
jroebu14: yeh -1. It was a requirement from Catharine. It must be just in this context then 🤔

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
jroebu14: ah I think lazy loading is handled in simorgh. We use this which I think is just a component to wrap image elements https://www.npmjs.com/package/react-lazyload

ryanmccombe: aah okay, the way this is set up at the minute won't let people wrap the images in custom elements - will need to update that

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
ryanmccombe: Had a deeper look - seems the `mediaIcons` in psammead are using this `focusable` attribute: https://github.com/bbc/psammead/blob/latest/packages/utilities/psammead-assets/src/svgs/mediaIcons.jsx#L10

Seems to be a non-standard IE attribute that achieves the same thing as tab index of -1 so will inherit that when I moved the icons in there

packages/components/psammead-podcast-promo/src/index.test.jsx
jroebu14: should we test that the component meets the acceptance criteria? e.g. renders the title, renders the description, renders the image, image has alt text. these things can break and would be hard to spot with one snapshot. I know there's no logic and we have 100% coverage with just 1 snapshot but it's good to have the confidence a component meets it's requirements and will be protected against any regressions with specific unit tests

packages/components/psammead-podcast-promo/src/components/card-image.jsx
ryanmccombe: Oh wow - didn't know about the right to left thing - so browsers would select all images on the page, and then filter them for ones that are children of the left selector

That's crazy - I assumed they'd just scan the subtree defined by the left selector

thanks!

packages/components/psammead-podcast-promo/src/components/title.jsx
ryanmccombe: 👍 https://github.com/bbc/psammead/pull/4009

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
ryanmccombe: Updated this image component to just be a basic styled `figure` where you have to pass your own `img` - updated the example and readme with how it can be used with the psammead-placeholder and lazy loader in simorgh

packages/components/psammead-podcast-promo/src/index.test.jsx
ryanmccombe: There's not one component here though - there are 9, with the idea being that they can be composed together (and with other components and HTML) to implement a range of criteria.

The acceptance criteria on the ticket are specific to how we want it used in Simorgh.  The example in `/examples/src` show how to use the component in simorgh to meet the simorgh AC, but we shouldn't assume it is always going to be used that way.

If we're creating a Psammead component, it is supposed to be a generic, reusable library.

eg, we shouldn't be forcing other psammead users to have stuff like `a section with a role of region` or `screenreaders to read out "[Brand Title], Episodes"` if they don't want to.  Those are our goals - not necessarily the goals of other psammead users

I know those "other users" barely exist, but if we want to start to solve that, we shouldn't be filling psammead with inflexible components hardcoded to Simorgh's use case

Alternatively, if we are creating a component we only want to be relevant to Simorgh, I think we should just create it in Simorgh - it makes our lives a lot easier.

ryanmccombe: That said, I do agree the snapshot is too noisy though - I can test each of the 9 components individually

But those tests would just be things  like `a card-title is an h3 with these css attributes` and I think tests like those are best represented as snapshots anyway - curious what others think

packages/components/psammead-podcast-promo/src/components/card-link.jsx
```diff
@@ -0,0 +1,29 @@
+import styled from '@emotion/styled';
+
+import { C_PEBBLE } from '@bbc/psammead-styles/colours';
+
+const CardLink = styled.a`
+  color: unset;
```
RichardPK, line 6: I think this `a` needs to have its default underline removed.

```suggestion
  color: unset;
  text-decoration: none;
```

packages/components/psammead-podcast-promo/src/components/card-image.jsx
```diff
@@ -0,0 +1,31 @@
...
+  margin: 8px 0 0 8px;
+  @media (min-width: ${GEL_GROUP_2_SCREEN_WIDTH_MIN}) {
+    width: 109px;
+  }
+  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
+    width: 178px;
+    margin: 0;
+  }
+  @media (min-width: ${GEL_GROUP_4_SCREEN_WIDTH_MIN}) {
+    width: 100%;
```
RichardPK, line 22: From the designs it looks like the row arrangement below 1008px could also be needed at 1008px and above? Maybe the screen width isn't appropriate as the only determinant for which arrangement of the card to use? 

Screenshot 2020-11-20 at 16 38 32

OSS contributors may also be like 'ooh a podcast card coool wait WHAT-'
Screenshot 2020-11-20 at 16 34 25

packages/components/psammead-podcast-promo/src/components/card-image.jsx
```diff
@@ -0,0 +1,31 @@
...
+import styled from '@emotion/styled';
+import { number, string } from 'prop-types';
+
+import {
+  GEL_GROUP_2_SCREEN_WIDTH_MIN,
+  GEL_GROUP_3_SCREEN_WIDTH_MIN,
+  GEL_GROUP_4_SCREEN_WIDTH_MIN,
+} from '@bbc/gel-foundations/breakpoints';
+
+import ImagePlaceholder from '@bbc/psammead-image-placeholder';
```
RichardPK: My naughty man has been `--no-verify`ing up in 'ere

RichardPK (approved):
Couple of comments but this is cool man, and the approach is quite a flip of how some other components are built - where we would usually mock up data (i.e. in a storybook fixture), pass that as one object into the parent component, then have those object values passed as specific props to the various child components. 

It'll obviously require quite a lot of composition on the Simorgh-side, but that gives us a lot more flexibility. It also makes this component a lot more useful for others.

packages/components/psammead-podcast-promo/src/components/card-image.jsx
ryanmccombe: aye it's definitely a bit janky in that regard - I'm not too sure how to solve this without using something like container queries

did consider adding some `compact` prop to make this configurable, but considering we're only going with this design for now, that's probably something that should wait until there's a v2 of this

packages/components/psammead-podcast-promo/src/index.test.jsx
jroebu14: I'm curious to hear others think too. I'm not totally convinced that snapshots alone are good quality tests. Any change made will cause the tests to fail. They don't point out regressions, just that something has changed. It's too easy to accept the change and update the snapshots and then potentially lose original acceptance criteria.

packages/components/psammead-podcast-promo/src/components/card-image.jsx
pharingee: Should we use GEL spacing here?

packages/components/psammead-podcast-promo/src/components/card-link.jsx
```diff
@@ -0,0 +1,30 @@
...
+  text-decoration: unset;
```
ryanmccombe:
```suggestion
  color: unset;
  text-decoration: none;
```

packages/components/psammead-podcast-promo/src/index.test.jsx
pharingee: I think we should test that the elements specified by  the a11y requirements are met. e.g. title should be an h3 element. But the css attributes shouldn't matter since they are not part of the requirements. I don't think specifying elements should be part of a11y requirements but I don't know enough about a11y to know if that is good or bad practice.

I do think that snapshots can miss this requirements quite easily and we need better tests to ensure that when the requirements change, someone needs to go in the tests to change them, without simply updating snapshots

pharingee (approved):
Added a comment on testing. Otherwise, LGTM

packages/components/psammead-podcast-promo/src/components/card-image.jsx
ryanmccombe: Gel spacing is all REM values - I don't think we want the image sizes to change based on font size.  Plus we don't have GEL_SPACING variables that match these sizes anyway :)

PR #4040, opened 2020-11-24
Social Embeds - Add support for lazy loading

andrewscfc:
Part of https://github.com/bbc/simorgh/issues/8103

Add support for embedding `CanonicalSocialEmbed` components in [react-lazyload](https://www.npmjs.com/package/react-lazyload) wrapper.

**Code changes:**
- Add code to render twitter and instagram embeds in a useEffect
  - This includes logic for twitter that is [currently in Simorgh](https://github.com/bbc/simorgh/blob/latest/src/app/containers/SocialEmbed/enrichTweet.jsx) and adds new logic for instagram; a subsequent PR will remove the existing logic from Simorgh and uptdate to this package.
- Add tests to ensure twitter and instagram apis are called

---


packages/components/psammead-social-embed/src/Canonical/index.jsx
```diff
@@ -59,6 +59,20 @@ export const providers = {
 const CanonicalEmbed = ({ provider, oEmbed }) => {
   useScript(providers[provider].script);
 
+  if (provider === 'twitter') {
```
karinathomasbbc: Do we need to worry about Facebook / YouTube / TikTok 🤣 ?

packages/components/psammead-social-embed/src/index.test.jsx
```diff
@@ -1,47 +1,64 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
```
karinathomasbbc, line 2: I have a feeling that `waitFor` is deprecated - can we use `act` instead? We did this in Simorgh recently https://github.com/bbc/simorgh/pull/8405

packages/components/psammead-social-embed/src/Canonical/index.jsx
andrewscfc: We don't support Facebook and TikTok yet (notsureif joking 😉 ), youtube doesn't need any extra logic to work with lazy loading afaik.

karinathomasbbc (approved):
LGTM

packages/components/psammead-social-embed/src/index.test.jsx
andrewscfc: I think `waitFor` is okay but `wait` is deprecated: https://testing-library.com/docs/dom-testing-library/api-async/#wait-deprecated-use-waitfor-instead

packages/components/psammead-social-embed/src/Canonical/index.jsx
simonsinclair: - We should try and avoid using hooks inside conditionals - this article is helpful in understanding why: https://reactjs.org/docs/hooks-rules.html.
- Is it possible to call `load` or `process` on the relevant embed instead of globally every time one is rendered?
- Instead of hard-coding provider specific logic inside this component, we could make use of the `providers` object on L21 and store a callback function as a property that we call inside `useEffect`. This is similar to the pattern used by `useScript`.

andrewscfc: I'll look into some of these improvements now, thanks @simonsinclair

andrewscfc: Hopefully addressed the above here: https://github.com/bbc/psammead/pull/4040/commits/60e835fe25faf247674637a18271f7d6490ea270

> Is it possible to call load or process on the relevant embed instead of globally every time one is rendered?

If I understand your question, from testing I think the api call on the window only enriches the embed in view and not any others that have yet to come into view. I don't know if it 're-enriches' those you have already seen? There is nothing in the docs here about enriching in a more granular way: https://developers.facebook.com/docs/instagram/oembed/

simonsinclair (approved):

PR #4075, opened 2020-12-01
Update RadioSchedules to be able to use client side routing

pharingee:
Resolves #NUMBER

**Overall change:** 
Add a new prop `linkComponent` to the Radio Schedules to allow the use of `ReactRouter`

**Code changes:**

- Add `linkComponent` and `linkComponentAttr` to `ProgramCard` and `RadioSchedule` components

---


SetJackal (approved):

jroebu14:
I think the pattern we wanted to go for was to export a Link component that the consumer (Simorgh) imports and composes in the application and then we use the `as` prop in Simorgh with the react-router Link component. Does that make sense? It's the same as what the recent episodes component does.

pharingee:
It makes sense but is essentially a rewrite of a component that hard hardly be reused for anything other than a radio schedule. For us to expose the link we need to essentially expose the program card and the link as well making it difficult to simply use as a single component.

Essentially, if we expose the inner workings all the way to the link, we will need to do more work in Simorgh to get a radio schedule component (similar to recent episodes, but that can be reused for video and audio). I  don't think the requirement of a csr link justifies that. We can revisit it when we need more finer control of a radio schedule component

packages/components/psammead-radio-schedule/src/index.test.jsx
```diff
@@ -17,4 +19,16 @@ describe('RadioSchedule', () => {
       selectedService: 'arabic',
     }),
   );
+
+  it('should render with passed component', () => {
+    const { container } = render(
+      renderRadioSchedule({
+        linkComponent: 'aside',
```
RichardPK, line 26: What made you pick an `aside` element with a `to` attr? Just to prove it'll have a go at rendering whatever you give it?

pharingee: We wanted to pick an element not likely to be used in immediate future updates to the radio schedules component. That's why we picked `aside`. And we picked `to` attr to ensure that other allowed JSX attributes other than the `href` default work

SetJackal (approved):

packages/components/psammead-radio-schedule/src/index.test.jsx
RichardPK: Yea I didn't think 'to' would be a valid attribute for an aside but #yolo 

Oh - it looks like since React 16, they allow any attribute, and it'll find its way to the DOM (from https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html):

```
// Your code:
<div mycustomattribute="something" />
```
would render an empty div to the DOM with React 15:

```
// React 15 output:
<div />
```

In React 16, we are making a change. Now, any unknown attributes will end up in the DOM:
```
// React 16 output:
<div mycustomattribute="something" />
```

RichardPK (approved):

PR #4107, opened 2020-12-08
Fix column overflow

andrewscfc:
Part of https://github.com/bbc/simorgh/issues/8558

**Overall change:**
_Edited:_
This PR has been changed to allow optional configuration of `width` for the `psammead-grid`, previously a width of 100% was hardcoded in the component to fix the issue. This an alternative to this Simorgh fix where the width 100% is applied in a media query https://github.com/bbc/simorgh/pull/8563, overriding the media queries applied in psammead.

I believe this solution may be more maintainable as it explicitly allows the psammead grid to have its width configured by the integrator. Overriding styles creates an implicit coupling between our Simorgh integration and the `psammead-grid`, I believe having explicit configuration allows future maintainers of `psammead-grid` to take into account requirements around the amount of space it occupies in its parent container rather than that being patched in the integration; `psammead-grid` could change in the future causing the overriden styles not to work, causing a regression.
_End of edit_

----

The overflow issue appears to be caused by the width applied to the [`GridPrimaryColumn` in Simorgh](https://github.com/bbc/simorgh/blob/latest/src/app/pages/StoryPage/StoryPage.jsx#L218):
image

With current value of `initial` firefox calculates the width incorrectly:
image

However Chrome calculates correctly:
image

~Moving to~ Configuring `width: 100%` ensures firefox now calculates the width correctly:
image

The width of columns calculated by css grid is now being respected:
image

PrimaryGrid spans 8 columns:
66.6667 * 8 = 533.3336

Plus 7 gaps at 16px:
16 * 7 = 112

Equals correct width:
533.3336 + 112 = 645.3336

---


RichardPK (approved):
Seems reasonable! `width: initial` was a root cause for a lot of layout issues actually - we had to override it in a lot of places in Simorgh.

This will touch a lot of components - Chromatic will be useful to see if this changes anything for the worse.

andrewscfc:
> Seems reasonable! `width: initial` was a root cause for a lot of layout issues actually - we had to override it in a lot of places in Simorgh.

Happy to jump on a call tomorrow and try and override in Simorgh to decrease the blast radius. We may want to do a change in Simorgh first before rolling this bigger fix out but I couldn't get my override to work.

packages/components/psammead-grid/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead-grid",
-  "version": "3.0.7",
+  "version": "3.0.8",
```
simonsinclair: Is this meant to be a patch bump? At the very least I would have thought minor (if you want to class it as a backwards compatible bug-fix), but it also alters the output, which has the potential to be a breaking change.

packages/components/psammead-grid/src/index.jsx
```diff
@@ -97,7 +97,7 @@ const gridMediaQueries = ({
       min: groups[group].min,
       max: groups[group].max,
       styles: `
-      width: initial;
+      width: 100%;
```
simonsinclair: Block elements default to a width of 100%, which, theoretically, is equivalent to `initial` in this case.
1. I'm curious why this rule is required and how changing it fixes this.
2. What happens if it is removed altogether?

packages/components/psammead-grid/package.json
andrewscfc: hmm, I saw it as a 'bug fix' so went for a patch; I see minor as 'adding a new feature but not a breaking change'. Happy do minor instead or even major, not precious about it.

packages/components/psammead-grid/src/index.jsx
andrewscfc: > I'm curious why this rule is required and how changing it fixes this.

I don't really know tbh 😞 , please feel free to try and work this out. Fwiw, it is doing what MDN says it should do rather than relying on less clear, default behaviour.

> What happens if it is removed altogether?

I can give this a try, I'll see what happens

andrewscfc: Dropping `width: initial` seems to adjust to width to be narrower, not really sure why:
image

RichardPK: Hey so this is because after dropping the `width: initial`, it then inherits from another grid style, which is assigning the width to 66.66666%

Screenshot 2020-12-09 at 13 21 00

RichardPK:
Simorgh-side fix for the specific issue on STYs in firefox: https://github.com/bbc/simorgh/pull/8563

Still all up for discussion how best to do this/why it's happening. Just wanted to PR that potential fix.

packages/components/psammead-grid/package.json
andrewscfc: This is now a minor bump, I hope this will be more agreeable now @simonsinclair ? The I have added new configuration but the change is completly backwards compatible.

simonsinclair:
Referenced in https://github.com/bbc/psammead/issues/4122.
Closing in favour of https://github.com/bbc/simorgh/pull/8563.

PR #4109, opened 2020-12-09
enable style overrides in timestamp

jroebu14:
**Overall change:**
Enables style overrides on the time element.

**Code changes:**

- adds `className` prop to timestamp container and timestamp component so that consumers can override styles.

---

### DIFF
--- a/packages/components/psammead-timestamp/src/index.jsx
+++ b/packages/components/psammead-timestamp/src/index.jsx
@@ -34,6 +34,7 @@ const Timestamp = ({
   padding,
   service,
   darkMode,
+  className,
 }) => (
   <StyledTimestamp
     dateTime={datetime}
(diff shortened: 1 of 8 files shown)


chris-hinds (approved):

packages/components/psammead-timestamp/src/index.jsx
```diff
@@ -34,6 +34,7 @@ const Timestamp = ({
   padding,
   service,
   darkMode,
+  className,
 }) => (
   <StyledTimestamp
```
ryanmccombe, line 39: This component seems to be unnecessary - it's just drilling props untouched

I think deleting line 29-49 and just exporting the `StyledTimestamp` directly might be better

ryanmccombe (approved):

packages/components/psammead-timestamp/src/index.jsx
ryanmccombe: > untouched

Not entirely - it's renaming `datetime` to `dateTime`

PR #4116, opened 2020-12-10
filter out undefined children

jroebu14:
Resolves #NUMBER

**Overall change:** Fixes the error erroring when trying to clone a child that is undefined when the child is conditionally not rendered.

**Code changes:**

- filters out falsey React children before mapping cloning
---

### DIFF
--- a/packages/components/psammead-episode-list/src/Episode.jsx
+++ b/packages/components/psammead-episode-list/src/Episode.jsx
@@ -34,9 +34,9 @@ const Episode = ({ children, dir }) => {
   return (
     <Wrapper dir={dir} showMediaIndicator={showMediaIndicator}>
       {showMediaIndicator ? (
-        Children.filter(Boolean).map(children, child =>
-          cloneElement(child, { showMediaIndicator }),
-        )
+        Children.toArray(children)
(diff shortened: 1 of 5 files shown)


ryanmccombe (approved):

packages/components/psammead-episode-list/src/Episode.jsx
```diff
@@ -34,7 +34,7 @@ const Episode = ({ children, dir }) => {
   return (
     <Wrapper dir={dir} showMediaIndicator={showMediaIndicator}>
       {showMediaIndicator ? (
-        Children.map(children, child =>
+        Children.filter(Boolean).map(children, child =>
```
RichardPK: Neat. Didn't know you could do this 👌

RichardPK (approved):

SetJackal (approved):

packages/components/psammead-episode-list/src/Episode.jsx
RichardPK: CI fail on this... looks like this can't be called if Children isn't an array? i.e. undefined

`TypeError: _react.Children.filter is not a function`

Issue #4123, opened 2020-12-10
Add English language version of podcast promo to help with a11y swarm


RichardPK:
**Is your feature request related to a problem? Please describe.**
- The current Podcast promo Storybook example is in Russian. It will be difficult to test if a screen reader is reading the correct thing, without understanding Russian
- Pass Russian as default service for current podcast promo story
- Update Podcast Promo tests with correct fixture data / service / script combo

**Describe the solution you'd like**
- [ ] Add support of multiple languages to podcast promo stories
- [ ] Add English language fixture data for 'news' service to help with swarm

**Additional context**

PR #4124, opened 2020-12-10
Range of podcast promo a11y fixes

### LINKED ISSUE #4108: Podcast Promo a11y fixes
**Describe the bug**
- A11y acceptance criteria diverged between the Podcast Promos issue, and the Jira ticket / Zeplin designs
- Changes required are detailed below

Jira: https://jira.dev.bbc.co.uk/browse/WSMEDIA-295
Screen reader UX: https://app.zeplin.io/project/5d4d77b1fcbfc79bf47e0294/screen/5fa4500498f1b6435cb487b5

**Expected behaviour**
- [x] Section - should be aria-labelledby the same id that the H2 value has. (see screen reader UX)
- [x] H2 icon - element wrapping the svg, change it from i to a span. There is an i tag - this is for italics text etc, this isn’t semantic HTML, lets please swap this for a non semantic element if it’s needed https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i 
- [x] Text resize - Icons don’t scale. They should use REM values. 
- [x] Image has a figure element - this can be announced by some screen readers, while you can use a figure for just an image on it’s own its only usually used if there is a caption, it’s also not in the screen reader ux. Remove the figure el
- [x] In the screen reader UX the link just has the brand title as the announced content, we also have ‘episodes’ - this isn’t needed 
- [x] Is the link the correct font weight? And correct visited colour? 
- [x] Change bg colour to #fdfdfd
- [x] 'More episodes' is aria hidden it shouldn’t be, and the screen reader UX specifies a paragraph element for the content here, also lets remove the unsemantic HTML e.g. the i tag wrapping the SVG

**Testing notes**

**Additional context**

RichardPK:
Resolves #4108 

**Overall change:** 

A range of changes to improve podcast promos' a11y compliance, and better match screenreader UI. More details of acceptance criteria are in the issue.

**Code changes:**

CallToAction
- Renamed `Card.CallToAction` to `Card.EpisodesText` as it is no longer intended as a call to action 

Content
- Remove `max-width` styling within card-content. This is now handled by `display: flex` in the parent `Card` component.

In the Storybook example:
- Give `region` an `aria-labelledby` property that matches the `PodcastPromo.Title` id
- Remove `VisuallyHiddenText`

SVGs
- Move out of their own components and directly place imported components into a parent `h2` or `p` element
- Make use of `rem` for sizing throughout so that svg size changes as base html font size is changed
- Unset colour of `SeriesStackIcon` so that it inherits from parent `span` rather than the svg's default

CardImage
- Change to styled `div`, not `figure`
- Change sizing to `rem` not pixels for better scaling with fonts
- Renamed component to `CardImageWrapper` to better reflect what the component is

Other
- Change card bg color to `GHOST`
- Set unvisited `a` color in `CardLink` to `EBON`

CallToAction
- Remove wrapping `span`
- Change text from `span` to `p` 

---


packages/components/psammead-podcast-promo/src/components/card-call-to-action.jsx
```diff
@@ -14,10 +14,10 @@ const CallToActionText = styled.span`
   color: ${C_METAL};
 `;
 const CardCallToAction = ({ children, ...props }) => (
-  <span aria-hidden="true">
-    <SeriesStackIcon />
+  <p>
+    <SeriesStackIcon focusable="false" />
```
ryanmccombe: Curious what effect this has - it seems to be setting this value on a span?

I didn't think the span would be focusable anyway - the SVGs are focusable on IE, but the SVG elements already have `focusable="false"`

packages/components/psammead-podcast-promo/src/index.test.jsx
```diff
@@ -85,28 +85,13 @@ describe('Podcast Promo', () => {
...
   describe('card-call-to-action', () => {
-    assertTypeOfElement(PodcastPromo.Card.CallToAction, 'span');
+    assertTypeOfElement(PodcastPromo.Card.CallToAction, 'p');
     shouldMatchSnapshot(
       'should match snapshot',
       <PodcastPromo.Card.CallToAction script={scripts.latin} service="russian">
         Click Me
       </PodcastPromo.Card.CallToAction>,
     );
-    it('should be aria-hidden', () => {
```
ryanmccombe, line 95: Just needs a README update - it has "This component is `aria-hidden` to screen readers." in the docs for this component

packages/components/psammead-podcast-promo/src/components/card-call-to-action.jsx
```diff
@@ -14,10 +14,10 @@ const CallToActionText = styled.span`
...
     <CallToActionText {...props}>{children}</CallToActionText>
-  </span>
+  </p>
```
ryanmccombe: Do we still need the wrapping element now that this is no longer aria-hidden?

ryanmccombe (approved):
🚀

packages/components/psammead-podcast-promo/src/components/card-call-to-action.jsx
RichardPK: Ah so they do - https://github.com/bbc/psammead/blob/d1306b84060e19a30d58e7084b512da9a48bdc3b/packages%2Futilities%2Fpsammead-assets%2Fsrc%2Fsvgs%2FmediaIcons.jsx#L10 

Nice one I'll remove 👍

packages/components/psammead-podcast-promo/src/components/card-call-to-action.jsx
RichardPK: Yea good point - I think I misunderstood which element should be a `p` here.

I've changed the `CallToAction` text to be a `p` and wrapped the elements in a React Fragment rather than a span. Screen reader seems to read the contents of the `CallToActionText` correctly.

packages/components/psammead-podcast-promo/src/index.test.jsx
RichardPK: 👍

packages/components/psammead-podcast-promo/README.md
```diff
@@ -53,7 +53,7 @@ This component is a basic `p` that is styled to be used for the synopsis/descrip
 
 ### Components: `PodcastPromo.Card.CallToAction`
 
-This component intended to be used as the "call to action" within each card. The child text is positioned alongside an icon. Any props passed to this component are forwarded to the `span` element that wraps the text. This component is `aria-hidden` to screen readers.
+This component intended to be used as the "call to action" within each card. The child text is positioned alongside an icon. Any props passed to this component are forwarded to the `p` element that wraps the text.
```
greenc05: Is this referring to the 'Episodes' text? Not sure that's a CTA?

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -208,8 +208,8 @@ exports[`Podcast Promo Examples basic example 1`] = `
 
 .emotion-2 > svg {
   margin-left: 0;
-  width: 22px;
-  height: 22px;
+  width: 1.375rem;
+  height: 1.375rem;
   fill: currentColor;
   position: relative;
   bottom: 5px;
```
greenc05: Should this be a pixel value? What's this for?

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -249,34 +249,34 @@ exports[`Podcast Promo Examples basic example 1`] = `
   }
 }
 
-.emotion-26 {
+.emotion-24 {
   padding: 0.5rem;
   display: inline-block;
   max-width: calc(100% - 96px);
```
greenc05: What's this going output? % or px?

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -370,25 +370,30 @@ exports[`Podcast Promo Examples basic example 1`] = `
...
+  color: #6E6E73;
+}
+
+.emotion-20 > svg {
   fill: currentColor;
+  color: unset;
+  width: 1rem;
+  height: 1rem;
   position: relative;
   bottom: 2px;
```
greenc05: Just checking we had inteded a px value here and for right?

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -517,65 +513,56 @@ exports[`Podcast Promo Examples basic example 1`] = `
...
         <h3
-          class="emotion-16 emotion-17"
+          class="emotion-14 emotion-15"
         >
           <a
-            class="emotion-14 emotion-15"
+            class="emotion-12 emotion-13"
             href="https://www.bbc.com/russian/media-47937790"
           >
             <span
```
greenc05, line 594: do we need this span?

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -517,65 +513,56 @@ exports[`Podcast Promo Examples basic example 1`] = `
...
-            </span>
           </a>
         </h3>
         <p
-          class="emotion-18 emotion-19"
+          class="emotion-16 emotion-17"
         >
           Мы быстро, просто и понятно объясняем, что случилось, почему это важно и что будет дальше. Никаких ненужных подробностей и передергиваний - только факты и взвешенная аналитика.
         </p>
         <span
```
greenc05: I would prefer to see this inside the p tag as it's associated with that content

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -460,27 +466,16 @@ exports[`Podcast Promo Examples basic example 1`] = `
...
-    class="emotion-30 emotion-31"
+    aria-labelledby="some-id"
+    class="emotion-28 emotion-29"
     role="region"
   >
     <div
       class="emotion-6 emotion-7"
     >
-      <i
+      <span
```
greenc05: I would prefer to see this inside the h2 tag as it's associated with that content

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -923,7 +904,7 @@ exports[`Podcast Promo Examples on-page example 1`] = `
       </div>
     </div>
     <div
-      class="emotion-40 emotion-41"
+      class="emotion-38 emotion-39"
     >
       <div
         class="emotion-4 emotion-1"
```
greenc05, line 971: what's the height for?

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -1169,35 +1149,31 @@ exports[`Podcast Promo card-call-to-action should match snapshot 1`] = `
...
+      />
+      <path
+        d="M0,32H24V8H0ZM4,12H20V28H4Z"
+      />
+    </svg>
   </span>
+  <p
+    class="emotion-4 emotion-5"
+  >
+    Click Me
```
greenc05: Please could we use better text, this isn't a link.

packages/components/psammead-podcast-promo/src/index.test.jsx
```diff
@@ -85,28 +85,13 @@ describe('Podcast Promo', () => {
   });
 
   describe('card-call-to-action', () => {
-    assertTypeOfElement(PodcastPromo.Card.CallToAction, 'span');
+    assertTypeOfElement(PodcastPromo.Card.CallToAction, 'p');
     shouldMatchSnapshot(
       'should match snapshot',
       <PodcastPromo.Card.CallToAction script={scripts.latin} service="russian">
```
greenc05: should russian be hard coded?

greenc05 (commented):
I would prefet to see all the example content in english including the alt text, does storybook output a russian lang attribute? If not, the screen readers accent will not be correct when testing in Storybook. This also makes testing in storybook easier for those who do not understand Russian.

packages/components/psammead-podcast-promo/README.md
RichardPK: Yea - I think this is another example of where the spec changed after we wrote the first issue, but the issue wasn't updated. At some point I think this was intended to be more of a CTA. I'll change the naming of this component.

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: ~These are the podcast icon svg's width & height. REM scales based on the base html font size, whereas hard-coded pixel values don't. This change makes the SVG scale as the user changed their browser font size~

Edit: Apologies, didn't realise you were talking about the 'bottom' on ln 215

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: ~This is another instance of using REM to size an SVG rather than pixels, so that it resizes based on the html base font size.~

Edit: Didn't realise you were talking about the bottom value on ln 392

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: With `calc()` the value parsed by the browser is the actual `calc()` expression. i.e.

```css
// Value specified in CSS
.foo {
    width: calc(100% - 50px);
}

// Computed value in browser
.foo {
    width: calc(100% - 50px);
}
```

I actually do want to dig into this `max-width` value... I need to ask @ryanmccombe why we need it.

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: These are just visually offsetting the SVG slightly to get its alignment correct. I've changed them to REM so that they scale with browser font size changes

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: These are just visually offsetting the SVG slightly to get its alignment correct. I've changed them to REM so that they scale with browser font size changes

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: Yes I think. so - it's receiving a className that conditionally styles the text depending on whether the link has been visited, focused or hovered.

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: Sure, that makes sense - done!

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: Nice one - done

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: This is the height of some of the dummy content used in the `on page` Storybook fixture. The height is just assigning how much height the element should take up...

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: Changed to 'Episodes'

packages/components/psammead-podcast-promo/src/index.test.jsx
RichardPK: We need to hard code _some_ service value here. It probably doesn't make sense to have latin script and English text with the Russian service. I'll address this as part of https://github.com/bbc/psammead/issues/4123

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
RichardPK: Separate to your query Catharine but an improvement was to use flex instead of max-width, so these calcs are gone 💥

packages/components/psammead-podcast-promo/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 0.1.0-alpha.7 | [PR#????](https://github.com/bbc/psammead/pull/????) Range of a11y improvements for component |
```
RichardPK:
```suggestion
| 0.1.0-alpha.7 | [PR#4124](https://github.com/bbc/psammead/pull/4124) Range of a11y improvements for component |
```

packages/components/psammead-podcast-promo/README.md
```diff
@@ -53,7 +53,7 @@ This component is a basic `p` that is styled to be used for the synopsis/descrip
 
 ### Components: `PodcastPromo.Card.CallToAction`
```
greenc05: does the name of this needs updating?

RichardPK: Ah yes - I had but hadn't pushed the commit. It's updated now

packages/components/psammead-podcast-promo/src/__snapshots__/index.test.jsx.snap
```diff
@@ -751,34 +791,17 @@ exports[`Podcast Promo Examples on-page example 1`] = `
   z-index: 1;
 }
 
-.emotion-20:visited .podcast-promo--visited {
+.emotion-16:visited .podcast-promo--visited {
```
greenc05, line 794: How come some styles don't have the emotion naming?

RichardPK: We manually assign these classes to the component here https://github.com/bbc/psammead/blob/3abe5ce9093f550d3ecfc44ce89d24f74c330355/packages/components/psammead-podcast-promo/src/examples/basic.jsx#L32

And then apply the style to that class based on the relevant psuedoclass (`&:visited`) here
https://github.com/bbc/psammead/blob/8713ebd264d85249405ce74ab0addc5e76ceb4a1/packages/components/psammead-podcast-promo/src/components/card-link.jsx#L17

greenc05 (approved):

PR #4125, opened 2020-12-10
Use CSS border for episode duration separator

### LINKED ISSUE #4034: a11y bug - episode list - the line before the timestamp should be a CSS border as this is presentation
**Describe the bug**
The line before the timestamp should be a CSS border as this is presentation. Using a character to achieve this is bad practice and also this looks odd/is not very clear/readable when there is no CSS 

e.g.
```html
<span aria-hidden=“true” class=“css-mro3c9 e1c0i89j0”>|</span>
```

**To Reproduce**

**Expected behaviour**

**Screenshots**

**Testing notes**

**Additional context**
CC @greenc05

pharingee:
Resolves #4034

**Overall change:** 
Create a `VerticalSeparator` component that can be used when a vertical separator is needed for audio episodes

**Code changes:**
- Create `VerticalSeparator` component in index  file
- Use component in fixtures and README.ms

---


packages/components/psammead-episode-list/src/fixtures.jsx
```diff
@@ -206,7 +201,7 @@ export const renderEpisodes = ({
               <EpisodeList.Metadata>
                 <>
                   {' '}
-                  <StyledSpan aria-hidden>|</StyledSpan> {episode.date}
+                  <EpisodeList.VerticalSeparator /> {episode.date}
                 </>
               </EpisodeList.Metadata>
```
ryanmccombe: I don't think we need the spaces here any more given the `VerticalSeperator` has its own spacing (from the padding and margin)

The spaces seem to be mis-aligning the divider: top is with spaces, bottom without:
Screen Shot 2020-12-11 at 10 07 42

packages/components/psammead-episode-list/src/index.jsx
```diff
@@ -34,6 +34,16 @@ const StyledEpisodeListItem = styled.li`
   }
 `;
 
+const VerticalSeparator = styled.span`
+  padding-left: 8px;
+  margin-right: 8px;
+  border-right: 1px solid ${C_CLOUD_LIGHT};
```
ryanmccombe: This probably looks better if users resize text:
```suggestion
  padding-left: 0.5rem;
  margin-right: 0.5rem;
  border-right: 0.0625rem solid ${C_CLOUD_LIGHT};
```
(0.5rem could be the `GEL_SPACING` const instead)

ryanmccombe (approved):
👍

rebeccamcginn (approved):

rebeccamcginn:
Ryan got in there first I'm afraid! You'll need to bump to 20

greenc05:
Sorry I wasn't working at the end of last week, please could we reopen this, while it does the job it doesn't take into account good practice. 

It's not good practice to use an empty span just for presentation e.g. a border. It would be better to remove this span element and apply a CSS border on the container span, reducing the page weight and improving practice.

PR #4150, opened 2020-12-15
Remove hanging separator span

pharingee:
**Overall change:** Remove the empty span used as a vertical separator and add border to existing elements

**Code changes:**
- Remove `VerticalSeparator`
- Create `BorderedSpan`
- Update tests
- Update  fixtures and README

---


packages/components/psammead-episode-list/src/index.test.jsx
```diff
@@ -202,9 +202,21 @@ describe('Episode List ', () => {
     expect(getAllByRole('text')[0].closest('a')).toBeInTheDocument();
   });
 
-  it('should render the vertical separator with an aria-hidden attribute', async () => {
-    const { container } = render(<EpisodeList.VerticalSeparator />);
+  it('should render the bordered span with a left border', async () => {
```
greenc05: would this be better? 'should render a left border'

packages/components/psammead-episode-list/src/index.test.jsx
```diff
@@ -202,9 +202,21 @@ describe('Episode List ', () => {
     expect(getAllByRole('text')[0].closest('a')).toBeInTheDocument();
   });
 
-  it('should render the vertical separator with an aria-hidden attribute', async () => {
-    const { container } = render(<EpisodeList.VerticalSeparator />);
+  it('should render a left border', async () => {
+    const { container } = render(<EpisodeList.BorderedSpan />);
```
greenc05: and shall we update the naming here too?

jroebu14 (commented):
It feels wrong to export a component that just has a border style.

I would just add the style in the fixture component and in Simorgh

pharingee:
@jroebu14 I think is a key part of the styling for the episode list so it might make sense to include it as part of the package to reduce external styling.

jroebu14:
@pharingee in that case I'd go down the route of encapsulating the logic this inside of this psmmead component so you don't have to implement it twice - once here and once in simorgh.

You could have logic that styles all children in the metadata component with a border-right (except the last child) or something like that e.g.

```jsx
{Children.toArray(children).map((child, i) => {
  const isLastChild = i === Children.count(children)

  if (isLastChild) {
    return child
  }
  return cloneElement(child, {
    css: { borderRight: 'solid 1px gray' },
  })
})}
```

packages/components/psammead-episode-list/src/Metadata.jsx
```diff
@@ -1,14 +1,50 @@
...
+      {Children.toArray(children).map((child, i) => {
+        const isFirstChild = i === 0;
+
+        if (isFirstChild) {
+          return child;
+        }
+
+        // Using array index here because data is immutable per episode
+        // eslint-disable-next-line react/no-array-index-key
+        return <BorderedSpan key={i}>{child}</BorderedSpan>;
```
jroebu14: I think if the child was a span it would wrap it in a span

packages/components/psammead-episode-list/src/fixtures.jsx
```diff
@@ -195,15 +195,9 @@ export const renderEpisodes = ({
                     locale: episode.locale,
                   })}`}
                 </span>
+                {episode.episodeTitle && episode.date}
```
jroebu14: it's in acceptance criteria for the date to be outside of the link so that screenreaders don't read it out when tabbing links

pharingee: I think it will be easier to make it aria-hidden then

jroebu14: but it's not supposed to be hidden from screenreaders. It still needs to be read out when you press the down arrow keys

jroebu14 (approved):

packages/components/psammead-episode-list/README.md
```diff
@@ -73,19 +73,19 @@ The following example shows a more advanced usage of this package. This is how w
...
         </EpisodeList.Description>
-        <EpisodeList.Metadata>
-          {episode.duration}
-          {episode.episodeTitle && <> <EpisodeList.VerticalSeparator> {episode.date}</>}
-        </EpisodeList.Metadata>
-      </EpisodeList.Episode>
-    </EpisodeList.Link>
+        <EpisodeList.Metadata>{episode.duration}</EpisodeList.Metadata>
+      </EpisodeList.Link>
+      <EpisodeList.Metadata hasBorder>
```
ryanmccombe, line 87: We should probably document what this prop does in the `EpisodeList.Metadata` section above

ryanmccombe (approved):
👍

packages/components/psammead-episode-list/src/__snapshots__/index.test.jsx.snap
```diff
@@ -515,14 +530,9 @@ exports[`Episode List  should render radio episodes correctly 1`] = `
...
+          class="emotion-44 emotion-17"
           dir="ltr"
         >
-          <span
-            aria-hidden="true"
-            class="emotion-44 emotion-45"
-          />
-           
           21 octobre 2020
         </span>
```
greenc05, line 537: How come we aren't seeing the time HTML element here?

pharingee: The html element here wasn't for the time but for the separator. This is what we fixed. This span was only there for the vertical separator and  we've fixed that in this pr

PR #4152, opened 2020-12-15
Recent tv episodes link & dateTime storybook fixture updates

### LINKED ISSUE #4149: Recent TV Episodes - Link & Time el fixture fixes
Psammead PR: https://github.com/bbc/psammead/pull/4152
Simorgh PR: https://github.com/bbc/simorgh/pull/8602

- Following a quick review by Catharine, she spotted some mismatches between the screen reader UX & the existing Recent TV Eps component: 

Screenreader UX https://app.zeplin.io/project/5d4d77b1fcbfc79bf47e0294/screen/5f2c330faba685397d2e12fb

**EpisodeList.Link**
- [x] VisuallyHidden 'Video,' text should be within the `<span role=text>` that contains the list, title etc. 
- [x] VisuallyHidden commas should be added between each separate piece of information in that `span`.
Something like:
Screenshot 2020-12-14 at 16 48 31

**Time el**
- [x]  When there is a bottom timestamp, within a time element (i.e. '20 Octobre 2020' in storybook example) it should not have a role=text attr.
- [x]  This time element should have a `datetime` attribute
- [x]   Currently, with css disabled - this time element runs into the video information. Wrap time element inside a div to make it a block el, meaning it breaks onto the next line when css is disabled.

**Testing notes**
[Tester to complete]

RichardPK:
Resolves https://github.com/bbc/psammead/issues/4149

**Overall change:**
- A range of changes were required to the Storybook Fixture for the TV Episodes List, to bring it up to date with the markup in Simorgh.
- This is caused by the way this compound component is built - we need to maintain changes to the way it is composed both in Simorgh, on the page itself, and in the Psammead Storybook.

**Code changes**
- The changes required are detailed in the issue here https://github.com/bbc/psammead/issues/4149
- One change: the use of a `div` rather than a `span` to wrap the timestamp, and removal of the `role=test` attr, have also been carried forward in Simorgh:  https://github.com/bbc/simorgh/pull/8602

---

### DIFF
--- a/packages/components/psammead-episode-list/src/fixtures.jsx
+++ b/packages/components/psammead-episode-list/src/fixtures.jsx
@@ -240,11 +240,12 @@ export const renderVideoEpisodes = ({
                 locale: episode.locale,
               })}
             />
-            <VisuallyHiddenText>Video, </VisuallyHiddenText>
             <EpisodeList.Link href={episode.url}>
+              <VisuallyHiddenText>Video, </VisuallyHiddenText>
               <EpisodeList.Title className="episode-list__title--hover episode-list__title--visited">
                 {episode.brandTitle}
               </EpisodeList.Title>
+              <VisuallyHiddenText>, </VisuallyHiddenText>
               <EpisodeList.Description className="episode-list__description--hover episode-list__description--visited">
                 {episode.episodeTitle || episode.date}
               </EpisodeList.Description>
@@ -258,11 +259,11 @@ export const renderVideoEpisodes = ({
               </VisuallyHiddenText>
             </EpisodeList.Link>
             {episode.episodeTitle && (
-              <span role="text">
+              <div>
                 <EpisodeList.Metadata as="time">
                   {episode.date}
                 </EpisodeList.Metadata>
-              </span>
+              </div>
             )}
           </EpisodeList.Episode>
         ))}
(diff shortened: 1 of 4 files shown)


packages/components/psammead-episode-list/src/fixtures.jsx
```diff
@@ -258,11 +262,11 @@ export const renderVideoEpisodes = ({
               </VisuallyHiddenText>
             </EpisodeList.Link>
             {episode.episodeTitle && (
-              <span role="text">
-                <EpisodeList.Metadata as="time">
+              <div>
```
RichardPK, line 265: Do you think this change should also be carried forward to Simorgh @jroebu14 ?

- The `div` turns it into a block element, which means when css is disabled, it still correctly breaks onto the next line
- `role=text` is unnecessary because there is only one nested text element. I think this is also true in Simorgh?

RichardPK:
Codeclimate spitting its dummy out at fixture data being similar 🤦  https://codeclimate.com/github/bbc/psammead/pull/4152

packages/components/psammead-episode-list/src/fixtures.jsx
RichardPK: Mocked the change in Simorgh regardless https://github.com/bbc/simorgh/pull/8602

jroebu14: yep good shout as long as it's still OK with CSS enabled too!

RichardPK: Nice one man - aye looks visually unchanged w/CSS. I've taken [the Simorgh PR](https://github.com/bbc/simorgh/pull/8602) out of draft

greenc05 (commented):
The screen reader UX looks much better now, thanks

chris-hinds (approved):

rebeccamcginn (approved):

PR #4217, opened 2021-01-12
Update axios version to mitigate high-severity vulnerability 

pharingee:
**Overall change:** Update the shared axios dependency to >=0.21.1

**Code changes:**

- Add axios as a dev dependency in package.json
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 4.0.62 | [PR#4217](https://github.com/bbc/psammead/pull/4217) Update axios version to mitigate high-severity vulnerability |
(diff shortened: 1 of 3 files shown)


simonsinclair (commented):
Hey @pharingee, based on the diff, it looks like we are _installing_ Axios in Psammead root as opposed to updating it. Please can you help me understand what effect this has?

joshcoventry:
I think axios is a sub dependency of another package we use, so we should not be installing it directly, we should update whatever uses axios.

pharingee:
@simonsinclair There are three dependencies that depend on `axios`: `chromatic` `github-api` and `yeoman-generator` and they all do not have updates that fix this vulnerability.

Screen Shot 2021-01-13 at 10 15 57 AM

I explored using [npm shrinkwrap](https://docs.npmjs.com/cli/v6/commands/npm-shrinkwrap) but that seems to be an overkill for this problem. Which is why I resulted to manually upgrading the dependency as a dev dependency.

pharingee:
Closing because it's fixed in https://github.com/bbc/psammead/pull/4223

Issue #4216, opened 2021-01-12
Update ConsentBanner component to use updated designs


andrewscfc:
**Is your feature request related to a problem? Please describe.**
We have updated the consent banner designs as shown here: https://zpl.io/bzKA46A, we need to apply to our amp and canonical consent banners.

Amp extended functionality will be covered by a seperate issue #4215 

We also need to address long standing accessibility issues, the accessibility criteria is below:
<details>
<summary>Accessibility Criteria</summary>

> Content Order 
> 
> Scenario: Content Order 
> 
> - Given I use a screen reader
> - When I navigate to the privacy or consent banner 
> - Then this should be the first content on the page
> - And the content order of the banner should match the screen reader UX

> Landmarks
> 
> Scenario: Landmarks 
> 
> - Given I use a screen reader
> - When I navigate to the consent or privacy banner on a canonical page  
> - Then a banner landmark will be announced
> - And this should not be nested within another landmark e.g. the main landmark or another banner landmark
> 
> Focus management
> 
> Scenario: Focus 
> 
> - Given I have activated the “OK” button on the privacy banner  
> - When the cookie banner is then displayed
> - Then focus should be on the H2 at the top of banner e.g. (“Let us know you agree to Cookies”) 
> 
> Scenario: Focus from buttons 
> 
> - Given I am on a consent banner 
> - When I select to accept/reject
> - Then the consent banner is closed  
> - And my focus is returned to the top of the page
> 
> Buttons/Links
> 
> - Given I am viewing [component]
> - When I inspect the code
> - Then ‘OK/Yes, I accept’ is a button HTML element
> - And not a link HTML element

> - Given I am viewing [component]
> - When I inspect the code
> - Then the ‘No, take me to settings/Find out what has changed’ is a link HTML element
> - And not a button HTML element

> - Given I am viewing [component]
> - And I tap on all buttons
> - Then touch targets must be large enough to touch accurately 
> - And they must have a width and height of 44px as a minimum
> 
> Scenario: Links/Buttons - Hover 
> 
> - Given I use a mouse
> - When I hover over link/button elements
> - Then they have a visible state change
> - And this is not identified by colour alone
> 
> Scenario: Links/Buttons - Focus
> 
> - Given I use a keyboard only
> - When I navigate through [component] by pressing the TAB key
> - Then all links/buttons elements are focusable
> - And they have visible state change on focus 
> - And this is not identified by colour alone
> - And this is in addition to the browser default focus outline styling
> 
[...]

PR #4225, opened 2021-01-14
Update consent banner

HarryVerhoef:
Contributes to completion of https://github.com/bbc/psammead/issues/4216

**Overall change:** _New consent banner design ._

**Code changes:**

- _1-column instead of 3._
- _Positive option is more obviously a button._

---


packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -41,7 +42,7 @@ const Wrapper = styled.div`
  * another div inside. This implementation mirrors the current orbit banner.
  */
 const CenterWrapper = styled.div`
-  max-width: ${GEL_GROUP_5_SCREEN_WIDTH_MIN};
+  max-width: 36.25rem;
```
andrewscfc: Just was trying to find where you derived this width from, looks like it come from here on the design:
image
582px to rem on a few website seems to come out at a slightly different value: `36.375rem`: https://www.easyunitconverter.com/px-to-rem

Don't see that as a huge issue though.

Just wondered if one of our UXers can point to where they derived the width, if it is defined somewhere in GEL it would be good to use an existing constant rather than a hardcoded width value.

andrewscfc: https://www.bbc.co.uk/gel

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -80,35 +80,35 @@ const Options = styled.ul`
   list-style-type: none;
 
   & li + li {
-    padding-top: ${GEL_SPACING};
+    margin-top: 0.75rem;
```
andrewscfc: Where does this come from on the design? Can that use a GEL constant?

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -80,35 +80,35 @@ const Options = styled.ul`
...
+    float: ${ltrRtl('right', 'left')};
   }
 
-  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
-    width: 18%;
-    float: ${ltrRtl('right', 'left')};
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
+    & li + li {
+      margin-top: 1rem;
```
andrewscfc: Where does this come from on the design? Can that use a GEL constant?

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -80,35 +80,35 @@ const Options = styled.ul`
...
+    }
   }
 `;
 
 const hoverFocusStyles = `
+  &:focus {
+    outline: none;
+    box-shadow: 0px 0px 1px 3px #68A1F8;
```
andrewscfc: Is this colour you are introducing in psammead-styles?

andrewscfc: We should always use `rem` units as these resize based on user's browser configuration for accessibility etc

packages/components/psammead-consent-banner/src/index.jsx
HarryVerhoef: 0.75rem = 12px, and the gap between the text and the top/bottom of the button is 12px, i could define a new GEL constant for 0.75rem if that is correct

packages/components/psammead-consent-banner/src/index.jsx
HarryVerhoef: Good spot, yeah i'll do that now

packages/components/psammead-consent-banner/src/index.jsx
HarryVerhoef: Yeah that's the one, it's for the focus effect for buttons and links

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -117,25 +117,35 @@ export const ConsentBannerText = styled.p`
 // Style `button` and `a` as children due to inability to set `on`
 // prop on styled component as required for the amp useage
 const ListItem = styled.li`
+  text-align: center;
+  display: inline-block;
+  width: 17.3125rem;
```
andrewscfc: again maybe extract a constant linked named after the gel guideline this dervied from maybe this aligns with 'spacing units' https://www.bbc.co.uk/gel/guidelines/spacing-units ?

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -117,25 +117,35 @@ export const ConsentBannerText = styled.p`
...
   & button {
-    ${({ script }) => script && getGreatPrimer(script)};
-    color: ${C_CONSENT_ACTION};
-    font-weight: 700;
-    background: none;
+    ${({ script }) => script && getLongPrimer(script)};
+    width: 100%;
+    min-height: 44px;
```
andrewscfc: use of px again, convert to rem

packages/components/psammead-consent-banner/src/index.jsx
andrewscfc: here this appears to align with 3 'spacing units' defined here, maybe name a constant after that? https://www.bbc.co.uk/gel/guidelines/spacing-units I'd consult with the UX guys to see if this is how they made their spacing choices.

packages/components/psammead-consent-banner/src/index.jsx
HarryVerhoef: Odd, the body text is of `width: 580px;`. The header here is positioned 2px to the left of the body text and the button.

packages/components/psammead-consent-banner/src/index.jsx
HarryVerhoef: This is for the width of the button which in the design is 277px, I checked the GEL spacing units and breakpoints and I couldn't find it, i'm not sure if it's derived from a breakpoint or from spacing units.

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -33,108 +51,95 @@ const Wrapper = styled.div`
...
 `;
 
-/*
- * The '&::after' below is to ensure that the background colour covers the
- * banner as the inner elements are float. The alernative is to have
- * another div inside. This implementation mirrors the current orbit banner.
- */
 const CenterWrapper = styled.div`
-  max-width: ${GEL_GROUP_5_SCREEN_WIDTH_MIN};
+  max-width: 36.25rem;
```
simonsinclair: Echoing what Andrew has said – let's try and reuse a constant closest to the Zeplin width rather than introducing an arbitrary one. 👍🏼

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -33,108 +51,95 @@ const Wrapper = styled.div`
...
-  }
 `;
 
 /*
  * The '& li + li' below allows for styling every `li` element except the first.
  */
 const Options = styled.ul`
   ${({ script }) => script && getLongPrimer(script)};
+  display: flex;
+  flex-direction: row;
```
simonsinclair: As discussed, this is the default and shouldn't need to be set. 👍🏼

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -33,108 +51,95 @@ const Wrapper = styled.div`
...
   padding: 0;
   margin: 0;
   list-style-type: none;
 
   & li + li {
-    padding-top: ${GEL_SPACING};
+    height: 100%;
+    margin-top: auto;
+    margin-bottom: auto;
   }
```
simonsinclair: As Harry and I have discussed, we should avoid these kind of styles - we should be making use of `text-decoration`. We should notify UX that this means the underline position cannot be controlled.

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -33,108 +51,95 @@ const Wrapper = styled.div`
...
 
-  @media (min-width: ${GEL_GROUP_3_SCREEN_WIDTH_MIN}) {
-    width: 18%;
-    float: ${ltrRtl('right', 'left')};
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
+    flex-direction: column;
```
simonsinclair: As Harry and I have discussed, these should be the default styles in a 'mobile-first' approach. Not much is needed here other than to flip the relevant styles around.

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -33,108 +51,95 @@ const Wrapper = styled.div`
...
     ${hoverFocusStyles}
   }
 
-  & a {
-    color: ${C_CONSENT_ACTION};
-    text-decoration: none;
-
-    ${hoverFocusStyles}
+  @media (max-width: ${GEL_GROUP_2_SCREEN_WIDTH_MAX}) {
+    width: 100%;
```
simonsinclair: Similar to the mobile-first styles comment. 👍🏼

simonsinclair (commented):
Super work, @HarryVerhoef! Please see my comments below - although we have already discussed them over Zoom. 😄

packages/components/psammead-consent-banner/src/index.jsx
HarryVerhoef: I spoke to Gail about this, the designs have been updated to `599px`, which is `GEL_GROUP_2_SCREEN_WIDTH_MAX` 👍

andrewscfc:
hey, I appreciate you didn't put together this originally but I can't understand how the consent banner works! 
- What code actually generates a `button` for the accept button and an `a` for the reject?
- Why aren't they both just buttons?
image

andrewscfc (commented):
I think the PR is missing storybook updates:
image

https://5d28eb5ee163f6002046d6fb-jvzreqnxib.chromatic.com/?path=/story/components-consentbanner--default

Otherwise be good if someone could help me with my question so I can review with a bit more confidence 👍

HarryVerhoef:
> hey, I appreciate you didn't put together this originally but I can't understand how the consent banner works!
> 
> * What code actually generates a `button` for the accept button and an `a` for the reject?
> * Why aren't they both just buttons?

I think the options rendering is done in Simorgh under /src/containers/ConsentBanner/Banner/index.canonical.jsx:

image

And I think the reason they aren't both buttons is for a11y, so that users know if pressing on something takes them to another page or not 👍

HarryVerhoef:
> I think the PR is missing storybook updates:
> image
> 
> https://5d28eb5ee163f6002046d6fb-jvzreqnxib.chromatic.com/?path=/story/components-consentbanner--default
> 
> Otherwise be good if someone could help me with my question so I can review with a bit more confidence 👍



> I think the PR is missing storybook updates:
> image
> 
> https://5d28eb5ee163f6002046d6fb-jvzreqnxib.chromatic.com/?path=/story/components-consentbanner--default
> 
> Otherwise be good if someone could help me with my question so I can review with a bit more confidence 👍



> I think the PR is missing storybook updates:
> image
> 
> https://5d28eb5ee163f6002046d6fb-jvzreqnxib.chromatic.com/?path=/story/components-consentbanner--default
> 
> Otherwise be good if someone could help me with my question so I can review with a bit more confidence 👍

I think that's the up to date component, should I update the component to contain more relevant text?

HarryVerhoef:
> I think the PR is missing storybook updates:

I think that is the up to date component, should I update the story text content to be more relevant?

simonsinclair (approved):
This is really great work, @HarryVerhoef. Please ensure UX are happy too. 
AMP uses this component, so when this is merged, it can be integrated in Simorgh as part of https://github.com/bbc/simorgh/pull/8718.

andrewscfc:
> I think the options rendering is done in Simorgh under /src/containers/ConsentBanner/Banner/index.canonical.jsx

Thanks for explaining @HarryVerhoef 

I think that's not great we've got styles for dom elements that are passed in from a completely seperate repo, maybe we should just pass the text in from Simorgh rather than dom elements? That would make this a bit easier to understand imo? The current approach creates coupling between Simorgh and Psammead.


> And I think the reason they aren't both buttons is for a11y, so that users know if pressing on something takes them to another page or not 👍

Makes sense, thanks!

HarryVerhoef:
Thanks for all the help @simonsinclair and @andrewscfc, both been awesome.

> I think that's not great we've got styles for dom elements that are passed in from a completely seperate repo, maybe we should just pass the text in from Simorgh rather than dom elements? That would make this a bit easier to understand imo? The current approach creates coupling between Simorgh and Psammead.

Yeah I agree, perhaps I could create an issue surrounding this? The buttons and links were being styled before this update (and still are currently) and I'm sure it's happening in other components too.

andrewscfc:
> Yeah I agree, perhaps I could create an issue surrounding this? The buttons and links were being styled before this update (and still are currently) and I'm sure it's happening in other components too.

My instinct is now may be a good time to sort this out as you have your head in how this all works, but happy to do an issue if people think this is excessive scope creep?

andrewscfc:
> I think that is the up to date component, should I update the story text content to be more relevant?

Sorry I didn't reply to this, I think it might be good to change the text to make it look more sensible in the story, just matching the english wording in the UX design?

HarryVerhoef:
> My instinct is now may be a good time to sort this out as you have your head in how this all works, but happy to do an issue if people think this is excessive scope creep?

Yeah I could give this a go, I'd have to update this component first then correct the Simorgh accept/reject rendering right?

andrewscfc:
> > My instinct is now may be a good time to sort this out as you have your head in how this all works, but happy to do an issue if people think this is excessive scope creep?
> 
> Yeah I could give this a go, I'd have to update this component first then correct the Simorgh accept/reject rendering right?

Yeah, you'd move the dom elements into here then just pass the text in the props in Simorgh when you integrate later. You'll also need to update the readme to indicate how the props are used now (passing a string rather than a DOM element).

I haven't looked at how this works in Simorgh closely so if this change becomes too difficult or feels the wrong way to go, feel free to backtrack and stick with it as it is

simonsinclair:
> > My instinct is now may be a good time to sort this out as you have your head in how this all works, but happy to do an issue if people think this is excessive scope creep?
> 
> Yeah I could give this a go, I'd have to update this component first then correct the Simorgh accept/reject rendering right?

Wait a minute, this is not a good time to re-engineer this component. Passing components to other components like this is a '[render prop](https://reactjs.org/docs/render-props.html)' pattern and actually reduces coupling. This PR should solve the issue at hand and it does that perfectly. It is currently blocking the AMP consent work, so needs to go out. Any further changes should be written up in another issue. Cc @andrewscfc @joshcoventry

andrewscfc:
> > > My instinct is now may be a good time to sort this out as you have your head in how this all works, but happy to do an issue if people think this is excessive scope creep?
> > 
> > 
> > Yeah I could give this a go, I'd have to update this component first then correct the Simorgh accept/reject rendering right?
> 
> Wait a minute, this is not a good time to re-engineer this component. Passing components to other components like this is a '[render prop](https://reactjs.org/docs/render-props.html)' pattern and actually reduces coupling. This PR should solve the issue at hand and it does that perfectly. It is currently blocking the AMP consent work, so needs to go out. Any further changes should be written up in another issue. Cc @andrewscfc @joshcoventry

I'd be happy to sort it in another issue but I'd dispute it doesn't introduce coupling, the css is in this component for an unknown set of dom elements passed in from another. If you passed in a `<div>` for arguments-sake, the css in here wouldn't work. If you went for the render prop pattern I'd move the css out of this component to Simorgh, both are valid solutions to coupling problem I see.

To be clear feel free to leave this to another issue 👍

andrewscfc (approved):
Happy to approve, would you be able to raise an issue @HarryVerhoef to capture potential coupling problem we've discussed?

HarryVerhoef:
> Happy to approve, would you be able to raise an issue @HarryVerhoef to capture potential coupling problem we've discussed?

Great, cheers, all done 👍
https://github.com/bbc/psammead/issues/4275

simonsinclair:
In Gail's absence, Pete Warren reviewed this over Zoom and was happy there were no glaring issues. It would be good to get this in front of Gail at some point next week. 👍🏼

packages/components/psammead-consent-banner/src/__snapshots__/index.test.jsx.snap
```diff
@@ -15,18 +15,35 @@ exports[`ConsentBanner should correctly render for ltr service 1`] = `
...
+.emotion-2 a {
+  color: #F6A21D;
+  -webkit-text-decoration: underline;
+  text-decoration: underline;
+  -webkit-text-decoration-color: #AEAEB5;
+  text-decoration-color: #AEAEB5;
+}
+
+.emotion-2 a:focus {
+  outline: none;
```
greenc05, line 39: Can you not use outline when you use box shadow? What's the reason for using box shadow instead? And why do something different here to the rest of the website? We use focus outline for everything to date.

packages/components/psammead-consent-banner/src/__snapshots__/index.test.jsx.snap
```diff
@@ -85,56 +101,85 @@ exports[`ConsentBanner should correctly render for ltr service 1`] = `
...
+    font-size: 0.875rem;
+    line-height: 1.125rem;
   }
 }
 
-.emotion-8 button:focus,
-.emotion-8 button:hover {
-  color: #FFFFFF;
+.emotion-8 button:focus {
+  outline: none;
```
greenc05, line 169: Same quesitons as above in comment on the link focus

packages/components/psammead-consent-banner/src/__snapshots__/index.test.jsx.snap
```diff
@@ -85,56 +101,85 @@ exports[`ConsentBanner should correctly render for ltr service 1`] = `
...
-  font-weight: 700;
-  background: none;
+  font-size: 0.9375rem;
+  line-height: 1.125rem;
+  width: 100%;
+  min-height: 2.75rem;
+  color: #222222;
+  font-weight: bold;
+  background-color: #FDFDFD;
   border: none;
```
greenc05, line 149: Missing a transparent border for high contrast mode on windows I believe? See last AAC on the issue

greenc05: Also missing a transparent border for the top of the component so you can see where the component starts in high contrast mode on windows and when the user changes colours in firefox?

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -33,108 +51,97 @@ const Wrapper = styled.div`
...
-    float: ${ltrRtl('left', 'right')};
-    ${ltrRtl('margin-right: 3.5%;', 'margin-left: 3.5%;')}
-    width: 22%;
-  }
 `;
 
 /*
  * The '& li + li' below allows for styling every `li` element except the first.
  */
 const Options = styled.ul`
```
greenc05, line 85: Please can we add an attribute of role list to the ul please? This will resinstate the list semantics on VoiceOver, which it removes due to seeing list stlye none on the CSS

greenc05 (commented):
Good work @HarryVerhoef just a couple of minor improvements needed I think to meet all the AAC and screen reaader UX? Also does anyone know why we use hidden="" on the outer div?

packages/components/psammead-consent-banner/src/__snapshots__/index.test.jsx.snap
HarryVerhoef: I used `box-shadow` because it's used here: https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/6000c4a406344f7bdcb17809

Happy to change to outline, but won't be able to get the same "fade" effect shown in the designs.

packages/components/psammead-consent-banner/src/__snapshots__/index.test.jsx.snap
HarryVerhoef: For a transparent border, would `solid 1px rgba(0,0,0,0)` be sufficient?

greenc05: This is what we use on the footer so you can see the top of it, would be good to be consistent: 0.0625rem solid transparent

greenc05: Something similar should work for the button too, though you'll need to check that.

HarryVerhoef: Perfect, thanks

packages/components/psammead-consent-banner/src/__snapshots__/index.test.jsx.snap
greenc05: We should always use the default outline. I believe the design was an 'illustration' of the default outline - Gail is going to make this clearer

PR #4248, opened 2021-01-18
Topic tags to storybook

### LINKED ISSUE #4230: Create topic tags component
Need to build a new component to storybook to show topics on simorgh pages.

Designs

https://zpl.io/anwnkEv

Screenreader UX 

https://zpl.io/amQj7y6

a11y a/c 

JIRA NEWSWORLDSERVICE-1103

Tags 

The tag should use the following fields:

'topicName:' to get the name of the Topic in the correct language
'topicId': to get the link for the topic (this is the TIPO ID)
The topic link will be of the format:

https:/www.bbc.com/{service}/topics/{topicid}
https:/www.bbc.com/{service}/{var}/topics/{topicid}
Users can click the tag to go the related topic page

E.g. the 'Europa' tag on the Mundo site would link to: https://www.bbc.com/mundo/topics/cyx5krnw43pt

NB: For CPS assets the about Schema.org will remain unchanged and continue to utilise the existing tags data in ares

One psammead package, composed of:

 # individual topic tag component

Create a new react component for the topic tag with props:
- Topic name
- Topic link
 
Use existing service prop and ensure topic tag height matches that required for the service: https://app.zeplin.io/project/5d4d77b1fcbfc79bf47e0294/screen/5fbbe4929669f3634150a667 

 # topic tags component

Encapsulates a collection of topic tag components, handles the logic on breakpoints and spacing.
The "related content" heading is not a part of the topic tags component, and is instead done in Simorgh.

Will require the direction prop.

**Testing notes**
[Tester to complete]

**Additional context**

HarryVerhoef:
Resolves #4230 

**Overall change:** _Publishes the topic tags container and topic tag components to storybook._

**Code changes:**

- _Created the new topic tags component and added stories to storybook_

---


packages/components/psammead-topic-tags/src/index.stories.jsx
```diff
@@ -0,0 +1,46 @@
...
+          <TopicTag topicName={shortText} topicLink="#" />
+          <TopicTag topicName={shortText} topicLink="#" />
+          <TopicTag topicName={shortText} topicLink="#" />
+        </TopicTags>
+      );
+    },
+    { notes, knobs: { escapeHTML: false } },
+  );
+
+buildRTLSubstories(STORY_KIND, { include: ['default', 'multiple topics'] });
```
HarveyPeachey: Is there a reason why we're adding separate RTL stories when the RTL services can just be accessed through the other stories?

packages/components/psammead-topic-tags/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Topic Tags Changelog
+
+| Version | Description |
+|---------|-------------|
+| 1.0.0   | [PR#4248](https://github.com/BBC-News/psammead/pull/4248) Initial creation of package. |
```
greenc05: Should have alpha version number until it's passed the accessiblity swarm?

packages/components/psammead-topic-tags/README.md
```diff
@@ -0,0 +1,56 @@
...
+);
+```
+
+### When to use this component
+
+The `TopicTag` component should only be used inside of a `TopicTags` component, and a `TopicTags` component should only be used to contain `TopicTag` components.
+
+### Accessibility notes
+
+The `TopicTags` component is fundamentally a `<ul>`, and so the `TopicsTags` component is announced as a list when using a screen reader, using the ARIA `role="text"` attribute.
```
greenc05: Think you got slightly mixed up, it's role list to announce as a list, so: "...using the ARIA role="list" attribute, this reinstates the list semantics on VoiceOver which removes them due to the use of list style type none in the CSS."

packages/components/psammead-topic-tags/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,649 @@
...
+  display: -ms-flexbox;
+  display: flex;
+  padding-top: 0.5rem;
+  padding-bottom: 0.5rem;
+  -webkit-text-decoration: none;
+  text-decoration: none;
+  color: #222222;
+}
+
+.emotion-2 a:hover {
```
greenc05: Missing focus styles?

packages/components/psammead-topic-tags/src/__snapshots__/index.test.jsx.snap
```diff
@@ -0,0 +1,649 @@
...
+.emotion-2 a>span {
+  background-color: #F2F2F2;
+  padding: 0.375rem 0.4375rem;
+}
+
+.emotion-2:not(:first-of-type) {
+  margin-top: 0.0625rem;
+}
+
+<div>
```
greenc05, line 73: So no list when only 1 link? Excellent.

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+  ${({ service }) => getSansRegular(service)}
+  ${({ script }) => script && getBrevier(script)}
+  ${ltrRtl('margin-right', 'margin-left')}: ${GEL_SPACING};
+
+  ${topicTagParent}
+`;
+
+export const TopicTag = ({ topicName, topicLink }) => (
+  <a href={topicLink}>
+    <span>{topicName}</span>
```
HarveyPeachey: Do we need the `span` here? Or can we just have the topicName in the `a` tag?

HarveyPeachey (commented):
Looks amazing, great work! Just a couple of small questions

greenc05 (commented):
Great work, just a couple of minor comments

packages/components/psammead-topic-tags/src/index.jsx
greenc05: Think this is to achieve the required height for the touch target area along with the design requirements

packages/components/psammead-topic-tags/src/index.stories.jsx
HarryVerhoef: Hey @HarveyPeachey, thanks for the comments!

I must have instinctively used this method after the consent banner work. Happy to remove it 👍

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: The `span` here is essentially the grey box of the `TopicTag` component, and was used so that the hit area of the link can be expanded beyond the borders of the grey box, to a minimum of `44px` 👍.

packages/components/psammead-topic-tags/CHANGELOG.md
HarryVerhoef: Thanks for the comments @greenc05,

Nice spot, on it!

packages/components/psammead-topic-tags/src/__snapshots__/index.test.jsx.snap
HarryVerhoef: Yes, the component renders a `<div>` instead of a `<ul>` if there is only 1 `TopicTag` component within the `TopicTags` component 👍

packages/components/psammead-topic-tags/package-lock.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@bbc/psammead-topic-tags",
+  "version": "1.0.0",
```
simonsinclair: The package(-lock).json version's should match the changelog entry.

packages/components/psammead-topic-tags/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/bbc/psammead/issues"
+  },
+  "homepage": "https://github.com/BBC-News/psammead/blob/latest/packages/components/psammead-topic-tags/README.md",
+  "dependencies": {
+    "@bbc/psammead-styles": "^7.2.1"
+  },
+  "peerDependencies": {
+    "react": "&gt;=16.9.0"
```
simonsinclair: Is that "`"&gt;=16.9.0"`" GitHub being weird? 🤔

packages/components/psammead-topic-tags/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "publishConfig": {
+    "tag": "alpha"
```
simonsinclair, line 31: This is often forgotten! 👍🏼

packages/components/psammead-topic-tags/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@bbc/psammead-topic-tags",
+  "version": "1.0.0",
+  "main": "dist/index.js",
+  "module": "esm/index.js",
+  "sideEffects": false,
+  "description": "Contains topic links",
```
simonsinclair: If these packages weren't public, it wouldn't matter as much, but can we make this a little more descriptive and relevant to people outside of our circle?

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+import { getBrevier } from '@bbc/gel-foundations/typography';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+
+const ltrRtl = (ltrValue, rtlValue) => ({ dir }) =>
+  dir === 'ltr' ? ltrValue : rtlValue;
+
+const topicTagParent = `
+  word-break: break-word;
+  & a {
```
simonsinclair: This ampersand shouldn't be necessary for nested children. It's useful to form pseudo selectors like `:hover` on the parent. e.g. `parent:focus {}`, but `parent a {}` should happen naturally.

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+import { string, element, oneOf, shape, arrayOf } from 'prop-types';
+import styled from '@emotion/styled';
+import { C_LUNAR, C_EBON, C_METAL } from '@bbc/psammead-styles/colours';
+import { GEL_SPACING } from '@bbc/gel-foundations/spacings';
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+import { getBrevier } from '@bbc/gel-foundations/typography';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+
+const ltrRtl = (ltrValue, rtlValue) => ({ dir }) =>
+  dir === 'ltr' ? ltrValue : rtlValue;
```
simonsinclair: I realise this kind of thing is all over our codebase, but it is an anti-pattern. The more dynamic our styles, the more of a cumulative performance hit we get due to lack of caching.

Please can we explore using symmetrical margins? Being able to draw a line down the middle of a design and flip it without the styles needing to change is a real sweet spot.

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+      text-decoration: underline;
+    }
+    &:visited {
+      color: ${C_METAL};
+    }
+  }
+
+  & a > span {
+    background-color: ${C_LUNAR};
+    padding: 0.375rem 0.4375rem;
```
simonsinclair: Are these arbitrary values necessary? From what I gather this is 6px and 7px respectively. If you're referring to Zeplin and units are not multiples of 4, then they are often inaccurate and you can apply your own interpretation using GEL_SPACING values. Where spacing genuinely needs to be 'off', we would ideally create a descriptive constant that holds this value.

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+  ${topicTagParent}
+`;
+
+const SingleTopicTagItem = styled.div`
+  ${({ service }) => getSansRegular(service)}
+  ${({ script }) => script && getBrevier(script)}
+  ${ltrRtl('margin-right', 'margin-left')}: ${GEL_SPACING};
+
+  ${topicTagParent}
+`;
```
simonsinclair: From what I can see these are identically styled components. Have you thought about using the `as` prop? https://emotion.sh/docs/styled#as-prop

packages/components/psammead-topic-tags/package-lock.json
HarryVerhoef: Thanks for the comments @simonsinclair! I'll get started working through them 👍 Can't believe I forgot to change the package jsons version number 😂

packages/components/psammead-topic-tags/package.json
HarryVerhoef: Huh, very odd, this is on my local branch too. It must have come from either `npm run create:package` or VSCode playing up somehow, good spot.

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: Thanks for informing me of that prop! Very useful.

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: I spoke with Yan about these spacings, and confirmed that the spacings are 6px and 7px respectively, I can fire up a PR that would introduce these as constants in `gel-foundations/spacings.js` 👍

simonsinclair: Sorry, I should have been clearer! I mean a constant local to the file (rather than in `gel-foundations`). For example, see `MIN_TAP_HEIGHT` in the following excerpt:

```js
//...

import {
  GEL_MARGIN_BELOW_400PX,
  GEL_MARGIN_ABOVE_400PX,
  GEL_SPACING_DBL,
  GEL_SPACING,
} from '@bbc/gel-foundations/spacings';

const MIN_TAP_HEIGHT = '2.75rem';

const COMMON_HEADING_STYLES = `
  color: ${C_WHITE};
  margin-top: 0;
  margin-bottom: 0;
`;

const Wrapper = styled.div`
  background-color: ${C_CONSENT_BACKGROUND};
  max-height: 100vh;

// ...
```

Assuming this is topic tag text padding, you might describe that in the following way:

```js
const [TAG_TEXT_PAD_Y, TAG_TEXT_PAD_X] = ['0.375rem', '0.4375rem'];
```

HarryVerhoef: Ah! My mistake, that makes a lot of sense, I'll implement that now 👍 Thanks!

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,130 @@
...
+  a > span {
+    background-color: ${C_LUNAR};
+    padding: ${TAG_TEXT_PAD_Y} ${TAG_TEXT_PAD_X};
+  }
+`;
+
+const topicTagsContainer = `
+  display: flex;
+  flex-wrap: wrap;
+  margin-top: -${ROW_SPACING};
```
andrewscfc: Not a negative, just to understand, why is negative margin needed here, is it overriding another margin applied elsewhere?

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,130 @@
...
+import { getBrevier } from '@bbc/gel-foundations/typography';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+
+const [TAG_TEXT_PAD_Y, TAG_TEXT_PAD_X, ROW_SPACING] = [
+  '0.375rem',
+  '0.4375rem',
+  '0.0625rem',
+];
+
+const topicTagParent = `
```
andrewscfc:
```suggestion
const topicTagStyles = `
```
Are these styles for an individual topic tag, just found the naming a bit confusing? If that is the case maybe `topicTagStyles`?

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,130 @@
...
+};
+
+TopicTag.propTypes = {
+  topicName: string.isRequired,
+  topicLink: string.isRequired,
+};
+
+TopicTags.propTypes = {
+  dir: oneOf(['ltr', 'rtl']),
+  children: oneOfType([arrayOf(element), object]),
```
andrewscfc: Should this propType restrict the `children` to be an array of `TopicTag` objects? You can use the propType above

packages/components/psammead-topic-tags/src/index.test.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+  shouldMatchSnapshot(
+    'should correctly render multiple topic tags for arabic',
+    <TopicTags service="arabic" script={arabic} dir="rtl">
+      <TopicTag topicName="test1" topicLink="#" />
+      <TopicTag topicName="test2" topicLink="#" />
+      <TopicTag topicName="test3" topicLink="#" />
+      <TopicTag topicName="test4" topicLink="#" />
+    </TopicTags>,
+  );
```
andrewscfc, line 63: I don't know if you want a test to cover what happens if you put something else as child of `TopicTags`, e.g. a random `div`? Maybe just ignore children that aren't `TopicTag` objects and don't render anything?

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: In the UX, there is a `1px` gap between each `44px` hit area, and so each topic tag has a `margin-top` of `1px` to give this effect, but then the first row of topic tags will be unnecessarily offset from the top of the container by `1px`, and this negative margin is to compensate 👍 I'm aware it's not a pretty solution, so open to suggestions.

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: Yeah good point that identifier is quite ambiguous, I'll make the change 👍

packages/components/psammead-topic-tags/src/index.jsx
andrewscfc: Thanks for explaining, all I can think of is something like this: https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child but you don't know how many children are in the first row. You could remove the 1px gap for the children on the first row but I don't know how you could identify them really. This seems fine to me unless there's a css expert about 💡

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: Wasn't aware this was possible 😮 Starting to really like TypeScript. In the case where only one `TopicTag` is given, the `children` prop is no longer an array, but an object instead. Think `oneOfType([arrayOf(TopicTag), object])` would be the way forward?

packages/components/psammead-topic-tags/src/index.test.jsx
HarryVerhoef: Yeah good idea, on it 👍

packages/components/psammead-topic-tags/src/index.jsx
andrewscfc: `oneOfType([arrayOf(TopicTag)])` was what I was thinking but my proptypes knowledge is a bit ropey! Don't get hung up on this too much if you can't get it to work. propTypes are okay but a miserable substitute for real types tbh, I find them better as code metadata than anything functional!

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: Yeah I was using `:not(:first-of-type)` to remove the margin of the first `TopicTag` until recently, when I found exactly what you described where there are more than 1 `TopicTag`s in the first row, so this is the best I could come up with 👍

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: That still seems to throw a console error, I think it's because when there's only 1 `TopicTag` the `children` prop isn't an array. Weirdly `oneOfType([arrayOf(TopicTag), TopicTag])` doesn't seem to work either, but `oneOfType([arrayOf(TopicTag), object])` does. I'll investigate this further on Monday to see if I can introduce a bit more specificity when there's only one `TopicTag` 👍

andrewscfc: ah `oneOfType([arrayOf(TopicTag), TopicTag])` _may work_, that matches either:
- an array of topic tags
- or a single topic tag
Again no need to sink to much time into this

HarryVerhoef: Thanks @andrewscfc ,
Yeah I gave that a go and would expect that to work, but it throws a type warning when rendering, I think because when only 1 child is given, it's formatted as a weird jsx object instead of a react element. I found the `objectOf` method and it seems to work in this instance, so the propType would be:
`children: oneOfType([arrayOf(TopicTag), objectOf(TopicTag)])`

simonsinclair: React `children` can be anything from elements to arrays. The type `node` is useful here.

If we validate that `children` is renderable, do we really need to check it's a TopicTag? Any over-engineering we do here could limit us in future – throwing a `TopicTagExpandable` into the mix for example.

HarryVerhoef: Thanks @simonsinclair, does this mean the `propType` should replace any mention of `TopicTag` with `node`?
So: `children: oneOfType([arrayOf(node), objectOf(node)])`?

simonsinclair: `children: node,` will work for all cases.

HarryVerhoef: Oh awesome, thanks!

packages/components/psammead-topic-tags/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Description
```
simonsinclair, line 7: This is a nice write-up!

packages/components/psammead-topic-tags/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@bbc/psammead-topic-tags",
+  "version": "0.1.0-alpha.1 ",
```
simonsinclair:
```suggestion
  "version": "0.1.0-alpha.1",
```

packages/components/psammead-topic-tags/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "bugs": {
+    "url": "https://github.com/bbc/psammead/issues"
+  },
+  "homepage": "https://github.com/BBC-News/psammead/blob/latest/packages/components/psammead-topic-tags/README.md",
+  "dependencies": {
+    "@bbc/psammead-styles": "^7.2.1"
+  },
+  "peerDependencies": {
+    "react": ">=16.9.0"
+  },
```
simonsinclair: It's really important to list all external dependencies. I can see quite a few that we've missed.

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,123 @@
+import React from 'react';
+import { string, oneOf, shape, node } from 'prop-types';
+import styled from '@emotion/styled';
+import { C_LUNAR, C_EBON, C_METAL } from '@bbc/psammead-styles/colours';
+import { GEL_SPACING_HLF, GEL_SPACING } from '@bbc/gel-foundations/spacings';
+import { getSansRegular } from '@bbc/psammead-styles/font-styles';
+import { getBrevier } from '@bbc/gel-foundations/typography';
+import { scriptPropType } from '@bbc/gel-foundations/prop-types';
+
+const [TAG_TEXT_PAD_Y, TAG_TEXT_PAD_X, ROW_SPACING] = [
```
simonsinclair: A style suggestion – to separate these values logically, you could keep TAG_TEXT_PAD's as a 'Tuple' using the array, as you have done, and declare ROW_SPACING separately. E.G.

```js
const [TAG_TEXT_PAD_Y, TAG_TEXT_PAD_X] = [,];
const ROW_SPACING = '';
```

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,123 @@
...
+  '0.375rem',
+  '0.4375rem',
+  '0.0625rem',
+];
+
+const topicTagsContainer = `
```
simonsinclair: It's good to see shared styles being declared in this way.

Can I suggest we be explicit with the naming? E.G. `CONTAINER_STYLES`. This will help people:
- Identify the variable as a [constant](https://www.google.com/search?q=constant+naming+convention&oq=constant+naming&aqs=chrome.1.69i57j0l7.4122j0j7&sourceid=chrome&ie=UTF-8).
- Not confuse it as component or anything else.

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  a > span {
+    background-color: ${C_LUNAR};
+    padding: ${TAG_TEXT_PAD_Y} ${TAG_TEXT_PAD_X};
+  }
+`;
+
+export const TopicTag = ({ topicName, topicLink }) => (
```
simonsinclair: If we think about the context in which these prop names are being used, do we think that prefixing them with `topic` is necessary? In this particular case - where we have no competing prop names, I would recommend we keep the API as intuitive as possible. This should also help to eliminate repetition in the code below.

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,123 @@
...
+  <a href={topicLink}>
+    <span>{topicName}</span>
+  </a>
+);
+
+export const TopicTags = ({ dir, children, script, service }) => {
+  if (children.length === 0) return null;
```
simonsinclair: These kinds of checks make a component less predictable. If I want to render this container, but haven't given it a child yet: do I go into debug mode, because it's simply not appearing on the page? Remember, I could have installed your package from NPM and have no idea how it's written. 

'[Pure functions](https://en.wikipedia.org/wiki/Pure_function)' return exactly one possible value for any given input. That means if I pass it zero or 10 children, a pure function would behave exactly the same way. Where possible, we should try to write functions that are pure. They are easier to reason, compose, and, for React, to optimise.

packages/components/psammead-topic-tags/src/index.stories.jsx
```diff
@@ -0,0 +1,59 @@
...
+storiesOf(STORY_KIND, module)
+  .addDecorator(withKnobs)
+  .addDecorator(withServicesKnob())
+  .add(
+    'default',
+    ({ dir, service, script, text }) => {
+      const shortText = service === 'news' ? text : text.trim().split(' ')[0];
+
+      return (
+        <TopicTags dir={dir} service={service} script={script}>
```
simonsinclair: Is there a reason `dir` is set here and in other places? Unless we are using the value (for example in styles, like `script`), the direction can be inferred by the global document direction, which is set at the root of every World Service page. See the root element of https://www.bbc.com/arabic.

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const hasMultipleChildren = children.length > 1;
+
+  return (
+    <>
+      {hasMultipleChildren ? (
+        <TopicsList dir={dir} role="list" service={service} script={script}>
+          {children.map((child, i) => {
+            if (child.type !== TopicTag) return null;
+            const key = i;
```
simonsinclair: Is there a reason we are not using `i` directly? If it's for readability, we could instead rename `i` to `index` or `key`.

simonsinclair (commented):
This is great work, as always, Harry. I've written a few comments - hopefully they help in some way.

packages/components/psammead-topic-tags/src/index.test.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+  shouldMatchSnapshot(
+    'should ignore non-TopicTag children',
+    <TopicTags {...baseProps}>
+      <TopicTag topicName="test1" topicLink="#" />
+      <div>
+        <p>ignore</p>
+      </div>
+      <TopicTag topicName="test2" topicLink="#" />
```
andrewscfc: For these tests that essentially yield an empty snapshot, I think it may be a bit more readable to render the component and assert that it has rendered nothing. I tend to use snapshots where you want to capture a complex nested structure of DOM elements as a snapshot; to assert the full structure would entail many assertions. 

Where you want test a single specific characteristic, e.g. a specific attribute is added like `dir` or in this case we are just rendering nothing, I tend to favour assetions for readability as they more clearly specify what you are expecting.

andrewscfc (approved):
One minor comment about tests, good to go though for me, once Simon's comments have been addressed👍

RichardPK (commented):
Lots of good feedback from others so will leave the approvals to them, but nice work team - looks great 👍

packages/components/psammead-topic-tags/src/index.jsx
HarryVerhoef: Thanks for your comments and suggestions @simonsinclair, I've addressed most of them now 👍

Regarding this, the reason `key` is declared and `i` is used indirectly is so that the linter ignores the fact that array index is being used for the component key. We discussed that this isn't an issue since the `children` array being mapped on won't change order. However, I've just realised that I can disable the linter using the wonderful `//eslint-disable-next-line react/no-array-index-key` comment in this case. Do you think that's ok? Thanks!

packages/components/psammead-topic-tags/package.json
HarryVerhoef: Thanks @simonsinclair,

I've added `prop-types` and `@emotion/styled` to `peerDependencies` and `@bbc/gel-foundations` to `dependencies`. I can't think of any more.

packages/components/psammead-topic-tags/src/index.test.jsx
HarryVerhoef: Thanks @andrewscfc,

I've rewritten these 2 tests which check for invalid usage, using the `expect` method. Is this ok? I haven't much experience with Jest, but they do seem to work. 

https://github.com/bbc/psammead/blob/a067c3767c07e9ea9c041c9e9d06ab95db1ddecf/packages/components/psammead-topic-tags/src/index.test.jsx#L65-L85

greenc05:
@HarryVerhoef would it be possible to get different text for each topic tag in storybook? This would make the screen reader testing easier as you can then be sure that the list item isn't being repeated and which item it is reading.

greenc05:
Do we need to output mark if there are are no topic tags? http://psammead-preview.tools.bbc.co.uk/4248/?path=/story/components-topictags--with-no-children Also unsure of the value of having this in storybook?

greenc05:
What is a non topictag child? http://psammead-preview.tools.bbc.co.uk/4248/?path=/story/components-topictags--with-a-non-topictag-child

HarryVerhoef:
Thanks for the comments, @greenc05 

> @HarryVerhoef would it be possible to get different text for each topic tag in storybook?

Yeah sure, I can sort that out 👍

> Do we need to output mark if there are are no topic tags?

Initially I just had it return nothing, but then Simon informed me of a design pattern called [pure functions](https://en.wikipedia.org/wiki/Pure_function), which is good for optimisation, would this cause problems for a screenreader? Happy to remove this from storybook, it was just one of the test cases.

> What is a non topictag child?

This is another test case for when a component that isn't a `TopicTag` is erroneously included within a `TopicTags` container component, such as a `div` or `p`.

greenc05:
Thanks @HarryVerhoef I was just thinking that code that isn't needed adds to page weight and data for users but if we are ok with that... Think we should remove non topictag child and no topic tags from Storybook, I can see value in a test but not for this to be in Storybook unless anyone else knows of why it's a good idea to have these cases in Storybook?

packages/components/psammead-topic-tags/src/index.test.jsx
andrewscfc: Yeah looks good, a bit more clear what you are testing for imo. Thanks 👍

simonsinclair:
Hey @HarryVerhoef, I was going to review this just now, but I can see you already have more than enough comments. Let's have a catch-up call this morning to discuss them and get this over the line. 🙂

simonsinclair (approved):

jroebu14:
Hey, with this PR I've realised that the `changeScanner` script is broken for new packages. Here is the fix https://github.com/bbc/psammead/pull/4370

andrewscfc:
Does this need testing/reviewing or is it good to merge @HarryVerhoef ?

PR #4299, opened 2021-02-01
Canonical ConsentBanner a11y changes

### LINKED ISSUE #4290: Amend Canonical Consent Banner
### Is your feature request related to a problem? Please describe.

Following a number of reviews, we must make some changes to the Canonical Consent Banner.

#### Visual UX
1. UX would like to make changes to the options container width between `400px - 599px`. Instead of options filling the entire width, they should be restricted to look more like the below image. See "New options container width" below.
2. At screen sizes where the options stack on top of each other, the last option needs to have the same padding as the one above. See "Fix bottom option padding" for context - the bottom option should have more space beneath it.
3. We should amend our `text-decoration: underline;` approach and create a faux-underline using `border-bottom` to give us control over its position. Previously, we could not see an example of this on live, but the following page shows there is a precedence for faux-underlines: https://www.bbc.com/afrique/region-55796266. (See point 4 below before starting this for links)

#### Screen reader UX / accessiblity acceptance criteria
1. Button hover state - While the grey on orange meets contrast, I’m not sure the contrast between this and the none hover state is sufficient to rely on colour alone, can we add an underline/border - see design https://app.zeplin.io/project/5c8a5e021d454124ef90669f/screen/600eb4159b96e2094d95484b
2. Button focus state - Add underline/border as per hover, and also reinstate the default focus outline (the box shadow on zeplin is an 'illustration' of this)
3. Add role list to attribue to ul. This will resinstate the list semantics on VoiceOver, which it removes due to seeing list stlye none on the CSS
4. Meet the following AAC (I suggest you implement a 2px transparent border on hover/focus to achieve this, 'transparent' will become a colour defined in the preferences in these modes) - for a code example see https://github.com/bbc/psammead/pull/707/files 

    Given I am using Windows High Contrast mode
    And also when I change my colour preferences in FireFox
    When I hover over link elements
    Then they have a visible state change
    
    Given I am using Windows High Contrast mode
    And also when I change my colour preferences in FireFox
    When I navigate through [component] by pressing the TAB key
    Then all links have a visible state change
    And this is in addition to the browser default focus outline styling
    
5. Meet the following AAC
    Given I am using Windows High Contrast mode
    And also when I change my colour preferences in FireFox
    When I view the consent/privacy banner
    Then an outline is visible for all buttons
    And I can see where the top of the compoment is **e.g. there is a line/border**

### Describe alternatives you've considered
N/A.

### Testing notes
[Tester to complete]

- [ ] ~~This feature is expected to need manual testing.~~

#### New options container width
Screenshot 2021-01-25 at 15 55 47

#### Fix bottom option padding
Screenshot 2021-01-25 at 16 18 34

HarryVerhoef:
Resolves #4290

**Overall change:** _ Some a11y fixes to the canonical consent banner._

**Code changes:**

- _Changed banner transparent border to 1px._
- _Removed outline styling so that the default is not overridden._
- _Added transparent border to the button so that the button is more visible in windows high contrast mode._

---

### DIFF
--- a/packages/components/psammead-consent-banner/src/index.jsx
+++ b/packages/components/psammead-consent-banner/src/index.jsx
@@ -6,7 +6,6 @@ import {
   C_CONSENT_BACKGROUND,
   C_CONSENT_ACTION,
   C_CONSENT_CONTENT,
-  C_CONSENT_FOCUS,
   C_WHITE,
   C_PEBBLE,
   C_EBON,
@@ -35,9 +34,6 @@ import { getSansRegular } from '@bbc/psammead-styles/font-styles';
 const transparentBorderHeight = '0.125rem';
 
 const hoverFocusStyles = `
-  &:focus {
-    outline: solid ${C_CONSENT_FOCUS};
-  }
   &:focus,
   &:hover {
     color: ${C_EBON};
@@ -48,7 +44,7 @@ const hoverFocusStyles = `
 const Wrapper = styled.div`
   ${({ service }) => getSansRegular(service)}
   background-color: ${C_CONSENT_BACKGROUND};
-  border-top: solid ${transparentBorderHeight} transparent;
+  border-top: solid 0.0625rem transparent;
 
   @media (max-width: ${GEL_GROUP_1_SCREEN_WIDTH_MAX}) {
     padding: calc(${GEL_SPACING_DBL} - ${transparentBorderHeight})
@@ -147,7 +143,7 @@ const ListItem = styled.li`
     color: ${C_EBON};
     font-weight: bold;
     background-color: ${C_GHOST};
-    border: none;
+    border: solid 0.0625rem transparent;
     margin: 0;
     cursor: pointer;
 
--- a/packages/components/psammead-consent-banner/src/index.jsx
+++ b/packages/components/psammead-consent-banner/src/index.jsx
@@ -143,7 +143,7 @@ const ListItem = styled.li`
     color: ${C_EBON};
     font-weight: bold;
     background-color: ${C_GHOST};
-    border: solid 0.0625rem transparent;
+    border: solid ${transparentBorderHeight} transparent;
     margin: 0;
     cursor: pointer;
 
--- a/packages/components/psammead-consent-banner/src/index.jsx
+++ b/packages/components/psammead-consent-banner/src/index.jsx
@@ -31,7 +31,7 @@ import {
 import { getSansRegular } from '@bbc/psammead-styles/font-styles';
 
 // Transparent border is to show edge of component and link underline on hover/focus in high-contrast mode
-const transparentBorderHeight = '0.125rem';
+const transparentBorderHeight = '0.0625rem';
 
 const hoverFocusStyles = `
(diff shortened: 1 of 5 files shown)


greenc05 (approved):
Looks good, thanks @HarryVerhoef

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -48,7 +44,7 @@ const hoverFocusStyles = `
 const Wrapper = styled.div`
   ${({ service }) => getSansRegular(service)}
   background-color: ${C_CONSENT_BACKGROUND};
-  border-top: solid ${transparentBorderHeight} transparent;
+  border-top: solid 0.0625rem transparent;
```
andrewscfc: I think this change has not been reflected in the padding calculations below, do we need seperate constants for the transparent link border (https://github.com/bbc/psammead/pull/4299/files#diff-972232a20082f3644690a9b606fed80b1787b4ad8fa8140415d5633b812636ceL81) and transparent wrapper component border here?

packages/components/psammead-consent-banner/src/index.jsx
```diff
@@ -31,7 +31,7 @@ import {
 import { getSansRegular } from '@bbc/psammead-styles/font-styles';
 
 // Transparent border is to show edge of component and link underline on hover/focus in high-contrast mode
```
greenc05: Can we update this comment please?

packages/components/psammead-consent-banner/src/index.jsx
HarryVerhoef: Thanks @andrewscfc, good spot, I've updated the `calc`s. The transparent link border is 2px upon hover/focus (0.125rem), and only used once, happy to make a constant for this but just wondering if it's necessary if only used once. I've updated the transparent wrapper border to use the `transparentBorderHeight` constant since it's also 0.0625rem (1px), happy to change this too if you feel that another constant would be required.

packages/components/psammead-consent-banner/src/index.jsx
HarryVerhoef: Thanks for the suggestion @greenc05, would "Transparent borders are used to highlight components and their borders in windows high-contrast mode." work?

greenc05: That's probably more appropriate for the a11y section of the read me? The const is just used for the top of the component and the button isn't it? Not the links? That's what I was referring to updating...

HarryVerhoef: Ah, thanks @greenc05, makes sense. I'll update now.

packages/components/psammead-consent-banner/src/index.jsx
andrewscfc: > The transparent link border is 2px upon hover/focus (0.125rem), and only used once, happy to make a constant for this but just wondering if it's necessary if only used once

If it's only used once I'd probably say it's not necessary

Changes all look good thanks 🙂

andrewscfc (approved):

PR #4303, opened 2021-02-04
Use v11 of emotion babel plugin

ryanmccombe:
Resolves [No Issue]

**Overall change:**
The babel plugin `emotion` refers to https://www.npmjs.com/package/babel-plugin-emotion
As of v11, we should be using `@emotion` which refers to https://www.npmjs.com/package/@emotion/babel-plugin

`@emotion/babel-plugin` does seem to have been correctly added to `package.json`, and `babel-plugin-emotion` was correctly removed, however, without this change to the `.babelrc` the psammead packages were still compiled using the old `babel-plugin-emotion`

Normally attempting to use a plugin that was not installed would have thrown an error, however, the old plugin was still available, as it was a transitive dependency of `react-select` which is used by `storybook`

The effect of having emotion 11 components in Psammead, compiled with an Emotion 10 plugin, then imported into Simorgh and recompiled with an emotion 11 plugin is almost certainly causing the issues we were seeing

Manual, local compilation using this babel config, and then importing of these Psammead packages into a local Simorgh does seem to indicate this is the solution to the problem.  However, applying this fix properly will require a version bump and republish of every Psammead package again

**Code changes:**

- _Update babel config to correctly use emotion plugin._
- _Bump everything to trigger rebuilds under new config._
- _Update storybook webpack config to let it work with emotion 11._
- _Update two components that output invalid CSS when a boolean is false._

---


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 4.0.92 | [PR#?](https://github.com/bbc/psammead/pull/?) Use Emotion v11 Babel Plugin |
```
ryanmccombe:
```suggestion
| 4.0.92 | [PR#4303](https://github.com/bbc/psammead/pull/4303) Use Emotion v11 Babel Plugin |
```

simonsinclair (dismissed):

joshcoventry (dismissed):

packages/components/psammead-most-read/src/Rank/__snapshots__/index.test.jsx.snap
```diff
@@ -310,16 +313,17 @@ exports[`MostReadRank should render rtl with double digits correctly 1`] = `
   color: #B80000;
   margin: 0;
   padding: 0;
+  letter-spacing: ;
```
ryanmccombe: !

ryanmccombe:
Dismissing reviews as this PR has totally changed

.storybook/main.js
```diff
@@ -1,8 +1,29 @@
+// Storybook does not currently support emotion 11
```
ryanmccombe, line 1: Relevant changes here - described in comment

jroebu14 (approved):

packages/components/psammead-most-read/src/Rank/__snapshots__/index.test.jsx.snap
```diff
@@ -225,16 +227,17 @@ exports[`MostReadRank should render rtl correctly 1`] = `
...
```
joshcoventry: Should this be here? There are numerous lines that look like this

joshcoventry (approved):

PR #4335, opened 2021-02-10
Migrate to yarn

### LINKED ISSUE #4331: Replace NPM with Yarn for package management
Similar to https://github.com/bbc/simorgh/issues/8761

We use `npm-force-resolutions` to prevent security vulnerabilities within certain versions of packages entering our builds. `npm-force-resolutions`  is run on the `preinstall` script instead of `install` which in certain situations allows security vulnerabilities to creep back into our builds.

Also, with Talos PRs we need to manually checkout the PR branch then locally `npm install` to force resolutions, push back up then seek approvals. This takes a lot of time and energy out of our day and we'd like to not have to do this.

We should switch to using Yarn for package management to make use of it's native support for force-resolutions.

Doing this work will also unlock other features of Yarn Workspaces covered in [this issue](https://github.com/bbc/psammead/issues/4332).

**Testing notes**
[Tester to complete]

**Additional context**

jroebu14:
Resolves https://github.com/bbc/psammead/issues/4331

**Overall change:** 
Replaces npm with yarn for package management.

**Code changes:**

NPM to Yarn command translations:
- `npm install` -> `yarn install` install packages
- `npm ci` -> `yarn install --frozen-lockfile` for deterministic builds based on lockfile
- `npm install --quiet` -> `yarn install --silent` installs without console noise
-  `npm update` -> `yarn upgrade` upgrades all packages
- `npm install some-package` -> `yarn add some-package` installs some-package
- `npm install --no-package-lock` -> `yarn install --no-lockfile` do no generate a lock file on installing
---


scripts/check-package-manager.js
```diff
@@ -0,0 +1,19 @@
+const ensureYarn = () => {
+  // If npm is being run, print a message and cause 'npm install' to fail.
+  if (!process.env.npm_execpath.includes('yarn')) {
+    console.log(
```
andrewscfc, line 6: Maybe add a lint ignore here to suppress this warning?

scripts/enforceVersions.sh
```diff
@@ -12,15 +12,15 @@ if [[ $(pwd) =~ .+psammead ]]; then
...
 
   # cut is used to transform 'version=6.4.1' into '6.4.1' for easy comparison
-  desired_npm_version=$(cat .npmrc | cut -d "=" -f2)
+  desired__version=$(cat .yarnrc | cut -d "=" -f2)
 
-  if [[ $(npm -v) != $desired_npm_version ]]; then
+  if [[ $(yarn -v) != $desired_yarn_version ]]; then
     tput setaf 1
-    echo "\nOh no! You're not using our version of npm, please move to version $desired_npm_version in order to reduce bundling inconsistency"
+    echo "\nOh no! You're not using our version of yarn, please move to version $desired_yarn_version in order to reduce bundling inconsistency"
```
andrewscfc, line 19: Did you do this on the Simorgh branch, I don't remember seeing it, the same script is in that repo too?

andrewscfc (commented):
Looks good, few thoughts:
- Did you you use [`yarn import`(https://classic.yarnpkg.com/en/docs/cli/import/) for the top level package lock? We needed to do this for Simorgh on the call the other day, it may well be necessary for psammead too? I guess the others were regenerated by the existing tooling.
- It seems you've wisely chosen to not remove lerna in this PR, I guess we may have a follow-up PR to either
  - Remove lerna and replicate it's functionality using yarn's built in features
  - Or investigate our existing issues with lerna and try and get local package linking working for example. A new version of lerna has appeared which gives us a bit more confidence in sticking with lerna: https://github.com/lerna/lerna/issues/2703#issuecomment-777089520

scripts/enforceVersions.sh
jroebu14: Yep this was in the original Simorgh PR

jroebu14:
@andrewscfc yeh I used `yarn import && rm package-lock.json` in the top level and then `lerna exec -- yarn import && rm package-lock.json` to do the same in every package.

Yeh this PR is just to use Yarn for installing dependencies and switching to Yarn commands. Anything else will be separate PRs.

.github/workflows/psammead-publish.yml
```diff
@@ -22,10 +22,10 @@ jobs:
           scope: '@bbc'
 
       - name: Install Node Modules
-        run: npm run ci:packages
+        run: yarn run ci:packages
```
chris-hinds: It's small and doesn't make much different but I think with yarn you can also remove the `run` command so just `yarn ci:packages` should work.

Not crucial to this work though.

.github/workflows/psammead-change-scanner.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           persist-credentials: false
 
       - name: Install Node Modules
-        run: npm run ci:packages
+        run: yarn run ci:packages
```
chris-hinds: Looks like the SetupNode GH action comes preinstalled with yarn! Nice 👍

chris-hinds (approved):

ryanmccombe (approved):

.github/workflows/psammead-publish.yml
jroebu14: Removed `run` from all yarn commands in [this commit](https://github.com/bbc/psammead/pull/4335/commits/a335bd91d50a5808b9a685b3080385b4b6f5ffb1)

chris-hinds (approved):

PR #4352, opened 2021-02-19
fix deploys - use node script publish not yarn publish

jroebu14:
**Overall change:**
Fixes package publishing by running the publish node script instead of Yarn's own publish command.

Also fixes the `changeScanner` script. It currently fails if you bump a `package.json` version without bumping the dependencies because no changes occur in the yarn.lock since yarn doesn't output the package version in `yarn.lock`
The `changeScanner` script now checks that if no dependencies have changed, then you don't need to update the `yarn.lock` file.

**Code changes:**

- replace `yarn publish` with `yarn run publish` to run the node script in `scripts/publish`

---

### DIFF
--- a/scripts/changeScanner/getChanges.js
+++ b/scripts/changeScanner/getChanges.js
@@ -20,7 +20,7 @@ const getChanges = () => {
 
     const isValidPackageName = validPackageNamePattern.test(fileName);
 
-    const [packageName, filePath] = isValidPackageName
+    const [packageName] = isValidPackageName
       ? [nameParts[2], nameParts.splice(3).join('/')]
       : [DEFAULT_PACKAGE_NAME, fileName];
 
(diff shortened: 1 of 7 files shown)


ryanmccombe (approved):

chris-hinds (approved):

chris-hinds (approved):

ryanmccombe (approved):
👍

joshcoventry (approved):

scripts/changeScanner/getChanges.js
```diff
@@ -20,7 +20,7 @@ const getChanges = () => {
 
     const isValidPackageName = validPackageNamePattern.test(fileName);
 
-    const [packageName, filePath] = isValidPackageName
+    const [packageName] = isValidPackageName
```
pharingee, line 23:
```suggestion
    const packageName = isValidPackageName ? nameParts[2] : DEFAULT_PACAKGE_NAME;
```

We can replace linens 23-25 with this and remove lines 27-29

pharingee (approved):

PR #4368, opened 2021-02-22
add yarn workspaces

### LINKED ISSUE #4332: Replace Lerna with Yarn Workspaces and enable cross-component development
Lerna is not maintained anymore and should not be relied on if bugs and security issues will not be addressed.

We think our use of Lerna can be replaced with Yarn Workpaces.

Using Yarn Workspaces will also unlock cross-component development within Psammead.

Blocked on implementation of Yarn for package management https://github.com/bbc/psammead/issues/4331

**Testing notes**
[Tester to complete]

**Additional context**

jroebu14:
Resolves https://github.com/bbc/psammead/issues/4332

**Overall change:**
Adds [Yarn Workspaces](https://classic.yarnpkg.com/en/docs/workspaces/).

**Update**
Talos depends on Lerna exec so not cannot be removed unless we find a replacement for either Lerna or Talos.

**Code changes:**

- Removes each package's lockfile. Yarn Workspaces uses a single lockfile rather than one for each workspace
- Removes `ci:packages` script:
  - What it did: installed dependencies in each package
  - Why is it removed: because Yarn installs all package's dependencies in a single pass in the root `node_modules`
- Removes `install:packages` script:
  - What it did: installed all package's dependencies and linked any cross-dependencies
  - Why is it removed: because `yarn install` installs all package's dependencies in a single pass in the root `node_modules` and links cross-dependencies by default
- Replaces `lerna exec` with `yarn workspaces run`
  - What it did: runs commands in each package - we use it for building packages with babel
  - Why is it replaced: `yarn workspaces run` is native to Yarn and does the same thing as `lerna exec`
- Changes psammead `moment-timezone-include` package to use commonjs modules. This is because the `postinstall` script in this package's `package.json` (which runs a node script in the build directory) ran before the `postinstall` script in the root `package.json` (which transpiles code and generates the build directories) and attempts to run a script that doesn't yet exist. Using commonjs modules means we don't need to transpile the code before running the node script.
- Updates the `changeScanner` script to check that the root `yarn.lock` has been updated if any package's dependencies have been updated

---


chris-hinds:
Nice work Jonathan. Does this still allow us to publish packages publicly individually to NPM as we do now. I thought the impression that by default yarn wanted packages to be private?

jroebu14:
I thought it was just the root package that had to be private

chris-hinds:
> I thought it was just the root package that had to be private

You are right, just read up on it again.

scripts/changeScanner/index.js
```diff
@@ -14,28 +14,24 @@ const requiredChanges = ['CHANGELOG.md', packageFileName];
...
-const someDepsHaveChanged = ({ localPackageFile, remotePackageFile }) => {
-  const { dependencies: localDeps, devDependencies: localDevDeps } = JSON.parse(
-    localPackageFile,
-  );
-  const {
-    dependencies: remoteDeps,
-    devDependencies: remoteDevDeps,
-  } = JSON.parse(remotePackageFile);
-  const depsHaveChanged = !equals(localDeps, remoteDeps);
-  const devDepsHaveChanged = !equals(localDevDeps, remoteDevDeps);
```
andrewscfc, line 36: Not sure I follow the changes in this file; why are devDeps no longer checked?

scripts/changeScanner/index.js
```diff
@@ -14,28 +14,24 @@ const requiredChanges = ['CHANGELOG.md', packageFileName];
...
-const getChangedFilePath = ({ packageName, matchFile }) =>
-  changedPackages[packageName].find(changedFile =>
-    changedFile.includes(matchFile),
-  );
+const getChangedFilePath = ({ packageName, matchFile }) => {
+  const packageChanges = changedPackages[packageName];
+
+  return packageChanges
+    ? packageChanges.find(changedFile => changedFile.includes(matchFile))
+    : '';
```
andrewscfc, line 22: What has changed to make `packageChanges` sometimes `undefined`?

andrewscfc (approved):
LGTM, really excited to see if this facilitates cross repo dev 🤞 Just a few questions around the change scanner but not blocking

scripts/changeScanner/index.js
jroebu14: They don't affect the root yarn.lock

scripts/changeScanner/index.js
jroebu14: this change https://github.com/bbc/psammead/pull/4368/files#diff-b91241351ff6aaef3a16d76228a999a9994e7038269586f10e5eb64e2b1d9657R45-R48

This is selecting the psammead root package which doesn't exist in `changedPackages` if there haven't been any changes in psammead root

chris-hinds (approved):
Changes look great. Thanks for this @jroebu14

FK78 (approved):

PR #4411, opened 2021-03-03
Psammead root package tidy

jroebu14:
**Overall change:**

Add dev deps to deps in root package.json.

> The root package.json of a monorepo is not published so whether a dependency is in devDependencies or dependencies does not make a difference and having one place to put dependencies in the root means that people do not have to arbitrarily decide where a dependency should go every time they install one.
_https://github.com/Thinkmill/manypkg#root-has-devdependencies_

Renovate pins root devDependencies so this change means no root dependencies will be pinned.

**Code changes:**

- Move devDependencies to dependencies
- Removes a redundant script used to update minor and patch

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 4.0.107 | [PR#4368](https://github.com/bbc/psammead/pull/4368) Remove root dev dependencies |
 | 4.0.106 | [PR#4368](https://github.com/bbc/psammead/pull/4368) use Yarn Workspaces |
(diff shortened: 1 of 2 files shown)


andrewscfc:
Did this not cause the yarn lock to change? I guess that proves the point you quoted if so?

andrewscfc (approved):

jroebu14:
@andrewscfc surprisingly no yarn.lock changes. So yeh I guess it proves the point

FK78 (approved):

FK78:
@jroebu14 any idea what might be causing the Change Scanner to fail?

jroebu14:
@FK78 I discussed with Chris - it's a problem with the GH action to detect changes between commits. We decided to merge this and remove it from our GH action later

PR #4419, opened 2021-03-03
Update topic tags component to use new designs

### LINKED ISSUE #4412: Update topic tags to new design
The topic tag designs have been updated.

The component is to be updated in alignment with the new designs, and it should keep the same alpha component version number.

New designs: https://zpl.io/2yKdGe9

n/a

**Testing notes**
a11y a/c: JIRA NEWSWORLDSERVICE-1103
Screenreader UX: https://zpl.io/2Z64m11

**Additional context**
n/a

HarryVerhoef:
Resolves https://github.com/bbc/psammead/issues/4412

**Overall change:** Updates the new topic tag component to use the new designs: https://zpl.io/2yKdGe9, essentially removing any clickable whitespace.

**Code changes:** 

- Give the `span` a display property of `inline-flex` so that it can be given a `min-height` of `44px` and allows easy centring of text using `justify-content` and `align-items`.
- Remove padding from the link for each tag so that whitespace is not clickable.

---


packages/components/psammead-topic-tags/src/__snapshots__/index.test.jsx.snap
```diff
@@ -244,8 +266,22 @@ exports[`TopicTags should correctly render a single topic tag for persian 1`] =
...
+  display: -webkit-inline-flex;
+  display: -ms-inline-flexbox;
+  display: inline-flex;
+  -webkit-align-items: center;
+  -webkit-box-align: center;
+  -ms-flex-align: center;
+  align-items: center;
+  -webkit-box-pack: center;
+  -ms-flex-pack: center;
+  -webkit-justify-content: center;
```
andrewscfc: lots of browser specific css generated by emotion, this suggests cross browser testing of this component is especially important cc @paruchurisilpa @JackJaeger

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -54,8 +55,13 @@ const SingleTopicTagItem = styled.div`
   }
 
   a > span {
+    display: inline-flex;
+    align-items: center;
```
andrewscfc: I'm not sure this is the intended usage of `inline-flex`, I've read here it is intended for laying out child elements as flex items but the `span` has no children: https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox#specifying_what_elements_to_lay_out_as_flexible_boxes ? 

Is there no other way of applying a `min-height` to the span, be interested hear what you tried?

In a similar vein, there are no child elements to align are there with `align-items`?

HarryVerhoef: Thanks @andrewscfc,

The min-height property could be applied here without `inline-flex`, but I used flexbox here so that the text of the topic tag could be centred vertically and horizontally using `align-items` and `justify-content` respectively; although there are no child HTML elements, there is the text that the span contains. And instead of using `display: flex;` here I used `inline-flex` here because span is an inline element.

packages/components/psammead-topic-tags/src/__snapshots__/index.test.jsx.snap
RichardPK: Aye I think IE 11 support is notoriously poor for some flex properties: https://caniuse.com/?search=justify-content

Otherwise should be OK. I think we already accept quite a few visual discrepancies in IE11.

packages/components/psammead-topic-tags/CHANGELOG.md
```diff
@@ -2,5 +2,5 @@
 
 | Version | Description |
 |---------|-------------|
-| 0.1.0-alpha.2 | [PR#4368](https://github.com/bbc/psammead/pull/4368) use Yarn Workspaces |
+| 0.1.0-alpha.3 | [PR#4368](https://github.com/bbc/psammead/pull/4368) use Yarn Workspaces |
```
RichardPK:
```suggestion
| 0.1.0-alpha.3 | [PR#4419](https://github.com/bbc/psammead/pull/4419) update to use inline flex styles |
| 0.1.0-alpha.2 | [PR#4368](https://github.com/bbc/psammead/pull/4368) use Yarn Workspaces |
```

RichardPK (approved):

RichardPK:
Hey Harry - just noticed the package has a `yarn.lock` file - no package should have its own yarn.lock, can you remove as part of this PR? cc @jroebu14  https://github.com/bbc/psammead/blob/8a511e9546122f7840a2886f434df1adba6fed5a/packages/components/psammead-topic-tags/yarn.lock

HarryVerhoef:
> Hey Harry - just noticed the package has a `yarn.lock` file - no package should have its own yarn.lock, can you remove as part of this PR? cc @jroebu14

Nice spot @RichardPK, I'll do that now, thanks 👍

packages/components/psammead-topic-tags/src/index.jsx
```diff
@@ -54,8 +55,13 @@ const SingleTopicTagItem = styled.div`
   }
 
   a > span {
```
simonsinclair: On the call we discussed reducing this to one element – `a`  and pruning the `span`, which was probably used in a previous iteration.

simonsinclair (approved):

packages/components/psammead-topic-tags/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 | Version | Description |
 |---------|-------------|
+| 0.1.0-alpha.4 | [PR#4419](https://github.com/bbc/psammead/pull/4419) Update designs |
```
simonsinclair, line 6: The package files will need to be bumped to this version also. 👍🏼

HarryVerhoef:
Just noticed this IE11 issue on chromatic, will look into + use a11y laptop to check.
Weirdly this must have changed after removing the span.
image

jroebu14 (approved):

PR #4420, opened 2021-03-04
Mega packages bumps

jroebu14:
**Overall change:**
Upgrades a lot of packages.

Updates all minor and patch dependencies and the following major dependencies:
- @testing-library/react-hooks
- eslint-config-prettier
- mkdirp
- object-hash

Consists of several Renovate PRs:

- Pin dependencies - renovate/pin-dependencies
- Update all 3rd party non-major dependencies - #4403
- Update dependency @testing-library/react-hooks to v5 - #4413
- Update dependency object-hash to v2- #4407
- Update dependency mkdirp to v1 - #4405
- Update dependency eslint-config-prettier to v8 - #4414

Also adds a new dependency `many-pkg` that runs [checks](https://github.com/Thinkmill/manypkg#checks) on `postinstall` to flag problems with our `package.json` files. Running this check on `postinstall` was inspired by [Chakra UI's approach](https://github.com/chakra-ui/chakra-ui/blob/main/package.json#L31).

Running `many-pkg` highlighted some problems such as dependency mismatches across packages and [peerDeps not being installed in devDeps](`many-pkg`). These issues were fixed in this PR.

**Code changes:**

Adds new root dependency `many-pkg` and 2 new scripts:
- `check:pkgs` - run this to validate our `package.json` files using these [checks](https://github.com/Thinkmill/manypkg#checks). Runs on `postinstall`
- `fix:pkgs` - run this to auto-fix problems highlighted by running `check:pkgs`

## `mkdirp` v1 breaking changes
Used in moment-timezone-include package

What changed?
- Promises instead of callbacks
- Dropped support for older versions of node

More details - https://www.npmjs.com/package/mkdirp/v/1.0.0#use-this-module-mkdirp-1x-if

**Changes needed**
None. We didn't use callbacks - are using mkdirp.sync and we use a supported version of Node.

--- 

## `object-hash` v2 breaking changes
Used in moment-timezone-include package

What changed?
- Only Node.js versions >= 6.0.0 are now supported

More details - https://github.com/puleos/object-hash#v200

**Changes needed**
None.

---

## `eslint-config-prettier` v8 breaking changes
Used in root package

What changed?
- Changed: [arrow-body-style] and [prefer-arrow-callback] are no longer turned off by default

More details
- https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#Version-700-2020-12-05
- https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#Version-800-2021-02-21

**Changes needed**
None.

---

## `@testing-library/react-hooks` v5 breaking changes
Used in root package

What changed?
- Interval and timeout time changes
- Deprecated function `suppressErrors` removed

More details
- https://github.com/testing-library/react-hooks-testing-library/releases/v5.0.0

**Changes needed**
None.


packages/components/psammead-brand/package.json
```diff
@@ -24,7 +24,9 @@
   },
   "devDependencies": {
     "@bbc/psammead-script-link": "^3.0.14",
-    "@bbc/psammead-styles": "^7.2.2"
+    "@bbc/psammead-styles": "^7.2.2",
+    "@emotion/styled": "^11.0.0",
+    "react": "^17.0.1"
```
andrewscfc, line 29: how do these dev dependencies work with the peer dependencies here? I guess the peer dependencies declare that you can integrate and use this package with a project (Simorgh in our case) that uses Emotion styled > 11 and react > 16.9 ? Should we change the react one to be react 17? I imagine this will be needed across all packages

andrewscfc (commented):
Just one point on peer dependnecies, ~otherwise I think this is good to merge~ looks like some dependencies for `psammead-timestamp-container` need pinning again looking at the failing tests

chris-hinds (approved):
LGTM. Thanks for this Jonathan

packages/components/psammead-brand/package.json
jroebu14: I _think_ if we do then we introduce a backwards-incompatible breaking change because consumers who haven't upgraded to React v17 won't be able to use new versions of Psammead components.

HarveyPeachey (approved):
Go ahead and Bump, ᴮᵘᵐᵖ!

PR #4441, opened 2021-03-17
remove Talos bumper bot

jroebu14:
**Overall change:**
Talos has been made redundant with the introduction of Yarn Workspaces* and a [check that fails our builds](https://github.com/bbc/psammead/blob/latest/package.json#L170) when there are [internal package mismatches](https://github.com/Thinkmill/manypkg#internal-mismatch).

This PR removes Talos and any associated docs and GitHub actions.

**Code changes:**

- Deletes the `scripts/talos` directory.
- Deletes talos GitHub actions in `.github/workflows` directory.
- Updates docs

_* Now that we are using Yarn Workspaces, when you install, all internal dependencies are linked locally, meaning they are not fetched from NPM and builds will not fail. This means we can release new packages and upgrade internal packages at the same time and Talos doesn't need to run at all._

---


FK78 (approved):
Goodbye Talos 👋🏼

yarn.lock
```diff
@@ -1100,19 +1100,6 @@
   resolved "https://registry.yarnpkg.com/@base2/pretty-print-object/-/pretty-print-object-1.0.0.tgz#860ce718b0b73f4009e153541faff2cb6b85d047"
   integrity sha512-4Th98KlMHr5+JkxfcoDT//6vY8vM+iSPrLNpHhRyLx2CFYi8e2RfqPLdpbnpo0Q5lQC5hNB79yes07zb02fvCw==
 
-"@bbc/psammead-radio-schedule@5.1.30":
```
andrewscfc: Should this change be in this PR?

jroebu14: I'm not sure what's happened but currently my other PRs have this change. This was the only internal dependency `@bbc/` package in the `yarn.lock` file. All other bbc/psammead packages are not in here. I'm guessing `@bbc/psammead-radio-schedule@5.1.30` found it's way in somehow but is actually not supposed to be there.

andrewscfc (approved):

PR #4445, opened 2021-03-24
remove vuln packages via force resolutions

jroebu14:
Resolves #NUMBER

**Overall change:**
Fixes a number of vulnerable packages.

**ejs**
Introduced in yeoman packages. Looks like this is fixed in the new major version of yeoman but no patch version of the current version we use is available. The major upgrade breaks the yeoman template generator in Psammead. There are no docs for the new version of yeoman so we don't know how to fix this without spending a lot of time digging through the yeoman code. We decided to force resolve a safe version of `ejs`.

**glob-parent**
Introduced in Storybook packages. We are up to date with Storybook so the only option is to force resolve a safe version of glob-parent. Psammead storybook still works as expected.

**hosted-git-info**
Introduced in many-pkg and yeoman packages. As described above we can't upgrade yeoman without potentially spending a lot of time doing so.

**semver-regex**
Introduced in yeoman package. As described above we can't upgrade yeoman without potentially spending a lot of time doing so.

**node-notifier**
Introduced in the jest package. Upgrading the jest package did not remove the vulnerable version of node-notifier so we incremented the already forced resolution.

Yeoman seems like it's not maintained too well since it's docs are quite of date and has a few vulnerable packages that affect us. Maybe worth considering one of these options:

- Removing yeoman since we don't create new packages too often and I can't remember a time I've used it or anyone else has
- Replacing yeoman with something else, either another package generator or our own simple CLI or npm/node script e.g. `yarn packages:create` that accepts arguments e.g. package name, author etc and generates the package
- Removing yeoman and having a "skeleton package" that developers can duplicate and change as needed with documentation on what exactly to change.

Issue for upgrading Yeoman https://github.com/bbc/psammead/issues/4447

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 4.1.5 | [PR#4445](https://github.com/bbc/psammead/pull/4445) security patches |
 | 4.1.4 | [PR#4444](https://github.com/bbc/psammead/pull/4444) remove lerna package |
 | 4.1.3 | [PR#4443](https://github.com/bbc/psammead/pull/4443) fixes versioning and changelogs link in docs |
 | 4.1.2 | [PR#4440](https://github.com/bbc/psammead/pull/4440) Add deprecation notice to psammead-sitewide-links |
--- a/package.json
+++ b/package.json
@@ -6,11 +6,15 @@
   "private": true,
   "resolutions": {
     "axios": "0.21.1",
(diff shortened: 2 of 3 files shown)


jamesdonoh (approved):
Thanks for looking into these @jroebu14 @FK78 @mukama, do we need an issue to look into the major Yeoman update? Otherwise feels like it might keep coming back and biting us.

jroebu14:
@jamesdonoh no worries. Yeh good point. I'll create an issue for the Yeoman update. Thanks

andrewscfc (approved):

PR #4425, opened 2021-03-09
Add deprecation notice to psammead-episode-list changelog and readme

RichardPK:
**Overall change:** 
- `psammead-episode-list` has been migrated to Simorgh, and this component should no longer be maintained in Psammead.
- This PR adds deprecation notices to the `changelog` & `readme`, prior to deprecation from NPM & removal from the repo

---

- [ ] I have assigned myself to this PR and the corresponding issues
- [ ] Automated jest tests added (for new features) or updated (for existing features)
- [ ] This PR requires manual testing

PR #4449, opened 2021-04-01
Remove psammead-amp-geo

HarryVerhoef:
**Overall change:** Removes psammead-amp-geo from psammead now that it has been migrated to Simorgh and is pending NPM deprecation.

**Code changes:**

- Removes the psammead-amp-geo code.
- Removes psammead-amp-geo from package.json.
- Remove psammead-amp-geo from the list of packages.

---


packages/components/psammead-amp-geo/README.md
```diff
@@ -1,5 +1,7 @@
...
 
+# Deprecation notice - this component is no longer being maintained.
```
andrewscfc: Do we need to use this utility: https://docs.npmjs.com/cli/v7/commands/npm-deprecate ? I don't if we've done this for other packages we've deprecated?

How have we deprecated other packages? I think we'd be better actually deleting the code as we'll end up maintaining unused code in the future if we make a change like move to `yarn` for example.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 <!-- prettier-ignore -->
 | Version | Description |
 |---------|-------------|
+| 4.1.6 | [PR#4449](https://github.com/bbc/psammead/pull/4449) Add deprecation notice to psammead-amp-geo |
```
jroebu14: I just merged in some changes so that the root CHANGELOG is not required

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bbc/psammead",
-  "version": "4.1.5",
+  "version": "4.1.6",
```
jroebu14: I just merged in some changes so that the root package version bumping is not required

yarn.lock
```diff
@@ -2,6 +2,7 @@
 # yarn lockfile v1
 
 
+
```
jroebu14: if you run yarn again you can get rid of this whitespace 🙂

packages/components/psammead-amp-geo/README.md
HarryVerhoef: Thanks @andrewscfc!
This PR mimics this other one: https://github.com/bbc/psammead/pull/4425, couldn't find any documentation for deprecating a psammead package. Happy to use the npm utility and/or delete the code but the npm utility would need to be run by the owner of the package.

CHANGELOG.md
HarryVerhoef: Awesome, thanks @jroebu14!

packages/components/psammead-amp-geo/README.md
andrewscfc: I'll broaden this conversation a bit, I'm a bit concerned about us accidently maintaining dead code

yarn.lock
```diff
@@ -9249,7 +9249,7 @@ jest-worker@^26.2.1, jest-worker@^26.6.2:
     merge-stream "^2.0.0"
     supports-color "^7.0.0"
 
-jest@^26.5.0, jest@^26.6.3:
+jest@^26.5.0:
```
HarveyPeachey: Hmm, is this an intended removal of `jest@^26.6.3`?

HarryVerhoef: Good spot, not intentional but this change occurs whenever `yarn` is run, and does so on latest for me too. Will remove this change for now 👍

HarveyPeachey (approved):
Dank 👍

andrewscfc (approved):

simonsinclair (changes requested):
The diffs on my end are showing files as deleted. For past package deprecations, we started by adding a notice to the package readme to allow a short grace period. Are we skipping this phase?

HarryVerhoef:
> The diffs on my end are showing files as deleted. For past package deprecations, we started by adding a notice to the package readme to allow a short grace period. Are we skipping this phase?

Thanks @simonsinclair

This PR originally did exactly that, but following a discussion with @andrewscfc and @jroebu14, we thought it might be better to first get an NPM admin to deprecate the package and then delete the code in order to prevent accidentally releasing new versions of deprecated components. Up for discussion though 👍

simonsinclair (approved):
> This PR originally did exactly that, but following a discussion with @andrewscfc and @jroebu14, we thought it might be better to first get an NPM admin to deprecate the package and then delete the code in order to prevent accidentally releasing new versions of deprecated components. Up for discussion though 👍

Ah! Thanks for clarifying, Harry.

PR #4511, opened 2021-07-07
Yarn v3 Berry Upgrade

### LINKED ISSUE #4527: Upgrade to Yarn v3 Berry
We should upgrade to Yarn v3 to make use of [zero-installs](https://yarnpkg.com/features/zero-installs), which will enable caching of Psammead dependencies so they can be installed offline.

Follow the migration instructions [here](https://yarnpkg.com/getting-started/migration).

CLI commands need to be migrated if they already exist. We should stick with the node_modules installation method instead of plug'n'play until we are ready to switch.

**Testing notes**
[Tester to complete]

**Checklist**

**Additional context**

HarveyPeachey:
Resolves #4527

**Overall change:** Upgrades Yarn to version 3 berry.

**Code changes:**
- Update to Yarn v3 berry release.
- Added new Yarn v3 gitignore file suggestions.
- Updated the enforce version number for Yarn.
- Removed CI environment check in preinstall script.
- Added yarn workspaces plugin and converted cli commands for workspaces.
- Made `ci` script explicit in workflows due to Yarn not running scripts unless node modules folder is available.
- Removed internal package dependencies.
- Updated getPackage script
- Updated package bumper script
- Updated publish script tests
- Bumped `psammead-detokniser` and `psammead-test-helpers` with empty dependency trees
- Removed `yarn create:package` reference in docs

---


scripts/utilities/getPackages/index.js
```diff
@@ -1,20 +1,9 @@
...
-    // handles CI shelljs output
-    packages = JSON.parse(output);
-  }
-
-  return Object.entries(packages).map(([name, { location }]) => ({
-    name,
-    location,
-  }));
+  // Filter used to remove empty strings
+  return outputStringArray.filter(item => item).map(JSON.parse);
```
jroebu14:
```suggestion
  return outputStringArray.filter(Boolean).map(JSON.parse);
```

.gitignore
```diff
@@ -18,3 +18,12 @@ packages/**/dist
 packages/**/esm
 
 published.txt
+
+# Yarn Files with Zero-Installs
+.yarn/*
+# !.yarn/cache
```
HarveyPeachey: Ignores yarn cache folder for now due to files changed becoming clogged with cache files, this will be uncommented in a future PR to make use of zero-installs feature.

.github/workflows/psammead-change-scanner.yml
```diff
@@ -17,7 +17,9 @@ jobs:
           persist-credentials: false
 
       - name: Install Node Modules
-        run: yarn ci
+        run: |
+          rm -rf node_modules
+          yarn install --immutable --check-cache
```
HarveyPeachey: Added cache checking for workflows due to this [suggestion](https://yarnpkg.com/features/zero-installs#does-it-have-security-implications). Would be better if we could create a workflow trigger that would check the cache based on the PR authors permissions in the repo, not sure if that's possible?

HarveyPeachey:
~Could potentially be blocked on Synk SCM yarn v2 [support](https://github.com/snyk/snyk/issues/1518#issuecomment-864286297).~

~Will discuss in tomorrows tech meeting on other engineer/product opinions~

Decision was to disable the failing snyk test as we have coverage from dependabot

.github/workflows/psammead-change-scanner.yml
HarveyPeachey: Double check why yarn user scripts don't run when no node_modules folder is available. See https://github.com/bbc/psammead/pull/4511#discussion_r687634538

.husky/pre-push
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-sh ./scripts/enforceVersions.sh && yarn test:unit -- --changedSince=latest
+sh ./scripts/enforceVersions.sh && yarn test:unit --changedSince=latest --passWithNoTests
```
HarveyPeachey: Check why Yarn v3 might be affecting the jest test files when they haven't been changed

scripts/utilities/getPackages/index.js
```diff
@@ -1,20 +1,9 @@
...
-  } catch (error) {
-    // handles CI shelljs output
-    packages = JSON.parse(output);
-  }
-
-  return Object.entries(packages).map(([name, { location }]) => ({
-    name,
-    location,
-  }));
+  // Filter used to remove empty strings
```
HarveyPeachey: Look to remove this comment as code explains this well enough

scripts/utilities/getPackages/index.test.js
```diff
@@ -6,18 +6,13 @@ describe(`Publish Script - getPackages`, () => {
...
-          },
-          '@bbc/psammead-bulleted-list': {
-            location: 'packages/components/psammead-bulleted-list',
-          },
-        }),
+      exec: () => ({
+        stdout: [
+          '{"location":"packages/components/psammead-consent-banner","name":"@bbc/psammead-consent-banner"}',
+          '{"location":"packages/components/psammead-brand","name":"@bbc/psammead-brand"}',
+          '{"location":"packages/components/psammead-bulleted-list","name":"@bbc/psammead-bulleted-list"}',
```
HarveyPeachey, line 13: Look to reimplementing the json stringify

.github/workflows/psammead-change-scanner.yml
HarveyPeachey: From looking at where this error is thrown from within the Yarn source code [here](https://github.com/yarnpkg/berry/blob/ebb09ae70be905fa4fa5c3fb4a73e3b1af8bf0e9/packages/plugin-nm/sources/NodeModulesLinker.ts#L51). There doesn't seem to be much of an explanation for why it was added. I have found an issue raising this as a bug on the Yarn Berry repo https://github.com/yarnpkg/berry/issues/2701#issuecomment-814462813, with a comment explaining the reason it works like this, is because Yarn exposes all dependancy binaries before running user scripts, even if a user script installs those binaries

andrewscfc: Had a bit of a think about `--check-cache`, I don't know how hard it would be but if the Actions ran where a PR didn't contain any cache changes, we could run this command without `--check-cache`; a PR including no cache changes means we have no reason to check the cache. If a PR did include cache changes we could still check the cache. This means:
- We are not vulnerable when dependencies change we always check NPM for cache accuracy
- When we want to make a change and NPM is down, assuming we aren't updating depedencies the change can go out with zero-installs. We would struggle to update dependencies  anyway while NPM was down as we wouldn't be able to get hold of a package version to put in the cache
- We accept that builds with dependency changes take longer, to my knowledge we have no major issues with build speed atm.

andrewscfc: > From looking at where this error is thrown from within the Yarn source code [here](https://github.com/yarnpkg/berry/blob/ebb09ae70be905fa4fa5c3fb4a73e3b1af8bf0e9/packages/plugin-nm/sources/NodeModulesLinker.ts#L51). There doesn't seem to be much of an explanation for why it was added. I have found an issue raising this as a bug on the Yarn Berry repo https://github.com/yarnpkg/berry/issues/2701#issuecomment-814462813, with a comment explaining the reason it works like this, is because Yarn exposes all dependancy binaries before running user scripts, even if a user script installs those binaries

I'm not sure I entirely understand it but for now I'd consider introducing a little `yarn-install-ci.sh` script to encapsulate the commands and reference them in all the actions. Out of interest why do we need to delete the node modules directory (`rm -rf node_modules`), isn't every github action ran in a clean docker instance?

scripts/utilities/getPackages/index.test.js
HarveyPeachey: I don't think we can use `JSON.stringify()` for these, as the output that we are mocking is provided in an [NDJSON](http://ndjson.org/) format, not the standard JSON format, hence the use of the `.join('\n')` to mock the behaviour of the [`yarn workspaces list --json`](https://yarnpkg.com/cli/workspaces/list) command.

@jroebu14 @andrewscfc

HarveyPeachey:
Blocked until https://github.com/bbc/psammead/issues/4527 is ready to merge, due to both repositories needing this upgrade to give devs a seamless experience

HarveyPeachey:
Unblocked as Simorgh PR is now in code review https://github.com/bbc/simorgh/pull/9318

.github/workflows/psammead-change-scanner.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           persist-credentials: false
 
       - name: Install Node Modules
-        run: yarn ci
+        run: ./scripts/yarnInstallCi.sh
```
jroebu14: this is my opinion but I liked explicitly seeing these steps. keen to hear what other think though

scripts/bumpPackages/index.js
```diff
@@ -29,7 +29,7 @@ const runExec = async (version, packageDir) => {
 
   return new Promise((resolve, reject) => {
     exec(
-      `yarn version --${versionTag} --no-git-tag-version`,
+      `yarn version ${versionTag}`,
```
jroebu14, line 32: this definitely still doesn't add any git tags?

.github/workflows/psammead-change-scanner.yml
HarveyPeachey: See https://github.com/bbc/simorgh/pull/9318#discussion_r690294866

scripts/bumpPackages/index.js
HarveyPeachey: From what I've read in the documentation I can see no mention of git tags https://yarnpkg.com/features/release-workflow. But I'm not 100% sure, would it be best if I test with this command manually?

jroebu14: what made you remove it? was it causing an error?

HarveyPeachey: The new plugin for `yarn version` doesn't support the --no-git-tag-version flag, and would error when trying to run it. And when running the package bumping script no git tags are added in the `.git/ref/tags` folder

package.json
```diff
@@ -24,18 +24,17 @@
   ],
   "scripts": {
     "build": "yarn build:cjs && yarn build:esm",
-    "build:cjs": "yarn workspaces run babel src -d dist --ignore '**/*.test.js,**/*.test.jsx,**/*.stories.js,**/*.stories.jsx' --root-mode upward",
-    "build:esm": "BABEL_TYPE=esm yarn workspaces run babel src -d esm --ignore '**/*.test.js,**/*.test.jsx,**/*.stories.js,**/*.stories.jsx' --root-mode upward",
+    "build:cjs": "yarn workspaces foreach --no-private exec babel src -d dist --ignore '**/*.test.js,**/*.test.jsx,**/*.stories.js,**/*.stories.jsx' --root-mode upward",
```
andrewscfc, line 27: Am I right in saying that `--no-private` here is exclusively to prevent these scripts running for the root package.json?

andrewscfc (approved):
Nice work, all good for me

package.json
HarveyPeachey: Yep that's right

HarryVerhoef (approved):
Really nice 👏 🚀

ryanmccombe (approved):

PR #4526, opened 2021-07-26
Add scanning to Psammead

HarryVerhoef:
**Overall change:** Adds an exposure scanning package for issues and pull requests to psammead.

**Code changes:**

- Added a script with unit tests as a psammead utility, with an alpha version. 

---

- [X] (BBC contributors only) This PR follows the [repository use guidelines](https://github.com/bbc/simorgh-infrastructure/blob/latest/documentation/repository-guidelines.md)


scripts/exposure-scanning/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "exposure-scanning",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "type": "module"
```
HarryVerhoef: I added a package.json here in so that I could use ES modules, not sure if there is an easier way... Also, unsure if there needs to be a specific value for the `license` attribute here

HarryVerhoef:
test

scripts/exposure-scanning/package.json
andrewscfc: To be clear, here you are publishing an `@bbc/exposure-scanning` package for consumption in Simorgh presumably?

I'd probably use the same licence as the parent psammead package.json: https://github.com/bbc/psammead/blob/latest/package.json#L53

scripts/exposure-scanning/scanExposures.js
```diff
@@ -0,0 +1,49 @@
...
+      owner: 'bbc',
+      repo,
+      prId: id,
+    };
+
+    const pr = await fetchPr(reqBody);
+
+    const scannedPr = await scanPr(pr, regex);
+
+    await patchPr(reqBody, scannedPr);
```
andrewscfc: Will this call the github api to patch, even when there is nothing to redact? We could hit rate limiting issues if so. It would be good to only call the api when necessary if this is the case.

andrewscfc (commented):
I'm slightly wary about this automation modifying issues/PRs in case it goes wrong and causes information to be lost. I'm happy to go with it but interested to see what others think? As long as it is scoped to a single PR/issue I'd be less concerned.

It is really impressive though and the approach you have taken is completely generic and could be used across the BBC and indeed externally I believe 💯

scripts/exposure-scanning/package.json
HarryVerhoef: Interesting. Perhaps I could take this out of `/scripts` and turn this into a psammead package that could be pulled in to simorgh?

scripts/exposure-scanning/scanExposures.js
HarryVerhoef: Really good point. I can definitely make this change 👍

packages/utilities/exposure-scanning/args/index.js
```diff
@@ -0,0 +1,45 @@
...
+  } catch {
+    return false;
+  }
+  return regex !== '';
+};
+
+const parseArgs = argv => {
+  if (argv.length !== 6) {
+    throw new Error(
+      'Incorrect number of args.\nUsage: node scripts/exposure-scanning <repo> <-pr/-issue> <id> <regex>',
```
HarryVerhoef: Update

packages/utilities/exposure-scanning/args/index.test.js
```diff
@@ -0,0 +1,54 @@
...
+      parseArgs(['node', 'scan.js', 'simorgh', '-pr', '1234', 'regex']),
+    ).toEqual({
+      repo: 'simorgh',
+      flag: '-pr',
+      id: '1234',
+      regexString: 'regex',
+    });
+  });
+
+  it('should throw an error without logging args if the number given is incorrect', () => {
```
HarryVerhoef: 'number of args'

packages/utilities/exposure-scanning/fetch/index.test.js
```diff
@@ -0,0 +1,111 @@
...
+
+jest.mock('node-fetch', () => ({
+  default: async url =>
+    ({
+      [prBodyEndpoint]: mockPrBodyRes,
+      [prCommentsEndpoint]: mockPrCommentsRes,
+      [prReviewCommentsEndpoint]: mockPrReviewCommentsRes,
+      [issueBodyEndpoint]: mockIssueBodyRes,
+      [issueCommentsEndpoint]: mockIssueCommentsRes,
+    }[url].clone()),
```
HarryVerhoef: Can remove this `clone()`

packages/utilities/exposure-scanning/patch/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+const patchPrBody = async (reqBody, body) => {
+  await octokit.request('PATCH /repos/{owner}/{repo}/pulls/{prId}', {
+    ...reqBody,
+    body,
+  });
+};
+
+const patchPrReviewComments = async (reqBody, comments) => {
+  await comments.map(async ({ id, body }) => {
```
HarryVerhoef: Could use `Promise.all` for these?

packages/utilities/exposure-scanning/patch/index.test.js
```diff
@@ -0,0 +1,96 @@
...
+      body: '* Body',
+      comments: [{ id: 5, body: 'Nice *' }],
+    };
+
+    await patchIssue(reqBody, issue);
+
+    expect(mockFetch).toHaveBeenCalledTimes(2);
+
+    expect(mockFetch).toHaveBeenCalledWith(
+      'https://api.github.com/repos/bbc/psammead/issues/4512',
```
HarryVerhoef, line 100: Try and check request body matches too

packages/utilities/exposure-scanning/scan/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+export const scanPr = (pr, regex) => {
+  let scannedPr = pr;
+
+  const { body, foundTextMatch } = scanText(pr.body, regex);
+  const { comments, foundCommentMatch } = scanComments(pr.comments, regex);
+  const {
+    comments: reviewComments,
+    foundCommentMatch: foundReviewCommentsMatch,
+  } = scanComments(pr.reviewComments, regex);
```
HarryVerhoef: This should all be in the try block

packages/utilities/exposure-scanning/args/index.js
```diff
@@ -0,0 +1,45 @@
+const isValidId = id => {
```
andrewscfc, line 1: Sorry probably should have suggested this library before, good one to avoid hand writing command line args parsing: https://www.npmjs.com/package/command-line-args

packages/utilities/exposure-scanning/scan/index.js
```diff
@@ -0,0 +1,77 @@
...
+    scannedPr = {
+      body,
+      comments,
+      reviewComments,
+    };
+
+    foundMatch =
+      foundTextMatch || foundCommentMatch || foundReviewCommentsMatch;
+  } catch (error) {
+    throw new Error(`Encountered an error when scanning.`);
```
andrewscfc: These errors will hide the cause of the original error as the stack trace will be lost; I'm not sure what catching the error achieves here as you throw another error? You normally catch an error and output to console for the user (e.g. `console.error`) and then either rethrow or return a suitable response to allow the application to continue as normal.

packages/utilities/exposure-scanning/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+    repo,
+    prId: id,
+  };
+
+  await {
+    '-pr': redactPr,
+    '-issue': redactIssue,
+  }[flag](reqBody, regex);
+
+  return Promise.resolve('No matches found.');
```
andrewscfc:
```suggestion
  return 'No matches found.';
```
I _don't think_ you need `Promise.resolve` here, `async` method implicitly return promises so your return value will automatically resolve a the implicit promise.

andrewscfc (commented):
Few last minor bits, looking good!

packages/utilities/exposure-scanning/args/index.js
HarryVerhoef: Cool, thanks, will have a look

packages/utilities/exposure-scanning/fetch/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+const fetchPrReviewComments = async reqBody => {
+  const prReviewComments = await octokit.request(
+    'GET /repos/{owner}/{repo}/pulls/{prId}/comments',
+    reqBody,
+  );
+  return prReviewComments.data.map(comment => ({
+    id: comment.id,
+    body: comment.body,
+  }));
```
jroebu14: Totally optional refactor - a little bit of destructuring for a cleaner look

```suggestion
  const { data } = await octokit.request(
    'GET /repos/{owner}/{repo}/pulls/{prId}/comments',
    reqBody,
  );
  return data.map(({ id, body }) => ({
    id,
    body,
  }));
```

packages/utilities/exposure-scanning/patch/index.js
```diff
@@ -0,0 +1,87 @@
...
+        {
+          owner: reqBody.owner,
+          repo: reqBody.repo,
+          commentId: id,
+          body,
+        },
+      );
+    }),
+  );
+};
```
jroebu14: I don't think you're supposed to await for a promise to resolve in each iteration like this. This would mean that you need to wait for the request to finish before the next iteration happens and so on until all items have been mapped. This is quite slow compared to correctly using a Promise.all where all requests happen at once e.g.

```suggestion
const patchPrReviewComments = (reqBody, comments) =>
  Promise.all(
    comments.map(({ id, body }) =>
      octokit.request(
        'PATCH /repos/{owner}/{repo}/pulls/comments/{commentId}',
        {
          owner: reqBody.owner,
          repo: reqBody.repo,
          commentId: id,
          body,
        },
      ),
    ),
  );
```

packages/utilities/exposure-scanning/scan/index.js
```diff
@@ -0,0 +1,77 @@
+const replacementStr = '*';
+
+export const scanText = (text, regex) => {
+  let foundTextMatch = false;
+  const scannedText = text.replace(regex, replacementStr);
+
+  if (scannedText !== text) {
+    foundTextMatch = true;
+  }
```
jroebu14: I think you can avoid a mutable variable here

```suggestion
  const scannedText = text.replace(regex, replacementStr);
  const foundTextMatch = scannedText !== text;
```

packages/utilities/exposure-scanning/scan/index.js
```diff
@@ -0,0 +1,77 @@
...
+      foundCommentMatch = true;
+      scannedComments.push({
+        id,
+        body: body.replace(regex, replacementStr),
+      });
+    }
+  });
+
+  return { comments: scannedComments, foundCommentMatch };
+};
```
jroebu14: 2 ways you can avoid mutable variables and an array push

One with a map and a filter
```js
export const scanComments = (comments, regex) => {
  const scannedComments = comments
    .map(({ id, body }) => {
      if (regex.test(body)) {
        return {
          id,
          body: body.replace(regex, replacementStr),
        };
      }
    })
    .filter(Boolean);
  const foundCommentMatch = scannedComments.length > 0;

  return { comments: scannedComments, foundCommentMatch };
};
```

or using a reduce

```js
export const scanComments = (comments, regex) => {
  const scannedComments = comments.reduce((acc, { id, body }) => {
    if (regex.test(body)) {
      return [
        ...acc,
        {
          id,
          body: body.replace(regex, replacementStr),
        },
      ];
    }

    return acc;
  });
  const foundCommentMatch = scannedComments.length > 0;

  return { comments: scannedComments, foundCommentMatch };
};
```

packages/utilities/exposure-scanning/scan/index.js
```diff
@@ -0,0 +1,77 @@
...
+      reviewComments,
+    };
+
+    foundMatch =
+      foundTextMatch || foundCommentMatch || foundReviewCommentsMatch;
+  } catch (error) {
+    throw new Error('Encountered an error when scanning.');
+  }
+
+  return { ...scannedPr, foundMatch };
```
jroebu14, line 55: It's easier to read this code if `foundMatch`is explicitly set to false rather than hunting down the variable definition and realising it's mutable and hunting down where it might mutate and then realising it would only ever be false here

```suggestion
  return { ...scannedPr, foundMatch: false };
```

packages/utilities/exposure-scanning/scan/index.js
```diff
@@ -0,0 +1,77 @@
...
+        body: body.replace(regex, replacementStr),
+      });
+    }
+  });
+
+  return { comments: scannedComments, foundCommentMatch };
+};
+
+export const scanPr = (pr, regex) => {
+  let scannedPr = pr;
```
jroebu14: this variable will always be mutated so it doesn't make sense to not just return inside the try like this

```js
export const scanPr = (pr, regex) => {
  try {
    const { body, foundTextMatch } = scanText(pr.body, regex);
    const { comments, foundCommentMatch } = scanComments(pr.comments, regex);
    const {
      comments: reviewComments,
      foundCommentMatch: foundReviewCommentsMatch,
    } = scanComments(pr.reviewComments, regex);

    return {
      body,
      comments,
      reviewComments,
      foundMatch:
        foundTextMatch || foundCommentMatch || foundReviewCommentsMatch,
    };
  } catch (error) {
    throw new Error('Encountered an error when scanning.');
  }
};
```

jroebu14 (commented):
Looks good. Made some suggestions to avoid let vars where possible since code is easier to reason about when you don't need to chase down where the var might mutate. I also think you have a few examples of incorrect Promise.all use. Promise.all is supposed to consume an array of promises. In your examples you're resolving the promise in each iteration (which is slower) and return the value so you're actually providing Promise.all with an array of values.

packages/utilities/exposure-scanning/patch/index.js
HarryVerhoef: Ahhh yeah that makes sense! My JS async knowledge is very limited 😂 Thanks 🙌

packages/utilities/exposure-scanning/fetch/index.js
HarryVerhoef: Yeah looks much cleaner, will do this now 👍

packages/utilities/exposure-scanning/scan/index.js
HarryVerhoef: Really good spot

packages/utilities/exposure-scanning/scan/index.js
HarryVerhoef: Thanks @jroebu14! This has been a really insightful review 🙌 Will make those changes now 🙂

packages/utilities/exposure-scanning/patch/index.js
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable global-require */
+let octokit;
+
+// Used to limit the number of calls to the patch comment and review comment endpoint
+const maxCommentPatches = 20;
```
andrewscfc, line 7: What prompted this being introduced, did you hit some rate limiting?

packages/utilities/exposure-scanning/patch/index.test.js
```diff
@@ -0,0 +1,164 @@
...
+    };
+
+    await patchPr(reqBody, pr);
+
+    expect(mockFetch).toHaveBeenCalledTimes(5);
+
+    expect(mockFetch).toHaveBeenCalledWith(
+      'https://api.github.com/repos/bbc/psamemad/pulls/1234',
+      pr.body,
+    );
```
andrewscfc, line 60: Is it worth checking the request method here to ensure it is `PATCH`: https://developer.mozilla.org/en-US/docs/Web/API/Request/method This would make sure the request actually makes a change to the PR body?

packages/utilities/exposure-scanning/src/index.js
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable no-console */
+/* eslint-disable import/extensions */
```
andrewscfc: Are these needed anymore?

packages/utilities/exposure-scanning/patch/index.js
```diff
@@ -0,0 +1,69 @@
...
+    octokit = new Octokit();
+  }
+})();
+
+const patchPrBody = async (reqBody, body) => {
+  await octokit.request('PATCH /repos/{owner}/{repo}/pulls/{id}', {
+    ...reqBody,
+    body,
+  });
+};
```
andrewscfc, line 24:
```suggestion
const patchPrBody = (reqBody, body) =>
  octokit.request('PATCH /repos/{owner}/{repo}/pulls/{id}', {
    ...reqBody,
    body,
  });
```
I _think_ following on from @jroebu14 's `async` comments this await may be redundant. As I understand it you will be returning this `Promise` returned by the Octokit api rather than generating a new one by awaiting the `Promise` from Octokit. I'm not completely certain on this, let me know if I'm wrong so I can learn 😅

packages/utilities/exposure-scanning/patch/index.js
```diff
@@ -0,0 +1,69 @@
...
+    ),
+  );
+};
+
+const patchIssueBody = async (reqBody, body) => {
+  await octokit.request('PATCH /repos/{owner}/{repo}/issues/{id}', {
+    ...reqBody,
+    body,
+  });
+};
```
andrewscfc, line 56:
```suggestion
const patchIssueBody = (reqBody, body) =>
  octokit.request('PATCH /repos/{owner}/{repo}/issues/{id}', {
    ...reqBody,
    body,
  });
```
Same as above _potentially_

andrewscfc (commented):
So sorry just a few more comments, nice learning `async` together 😅

packages/utilities/exposure-scanning/patch/index.test.js
HarryVerhoef: Good idea 👍

packages/utilities/exposure-scanning/patch/index.js
HarryVerhoef: Ahhh good point, yeah I think you're right I'll give it a go 👍

packages/utilities/exposure-scanning/patch/index.js
HarryVerhoef: Not so much rate limiting, but rather I found that the endpoint for updating comments and review comments doesn't take the PR or issue id as an argument (I initially thought it took it as an argument in the body), but instead it gives each comment across the repo a unique id and that's used to identify the comment we'll update. So it's kind of a way of mitigating the scenario you described in the mob review where it goes and updates a load of comments outside the PR/issue in question. It's very unlikely considering the comment id's are obtained from [an endpoint](https://docs.github.com/en/rest/reference/pulls#list-review-comments-on-a-pull-request) which **does** take in the PR/issue id, but I think it's useful for if someone were to change the way that the fetch and patch systems interact

packages/utilities/exposure-scanning/src/index.js
HarryVerhoef: Good spot, will remove 👍

packages/utilities/exposure-scanning/patch/index.js
jroebu14: This might help - if you want the `patchPrBody` to resolve with an `undefined` value then Harry's current code is what you want. If you want to return whatever value the `octokit.request` promise resolves to then Andrew's suggestion is what you want. If you don't use the resolved value then Andrew's suggestion has the benefit of being terser.

andrewscfc: Greats thoughts @jroebu14 , I'd actually probably leave it as it is on that basis as it better encapsulation to not return the result from this function; there is no need for the caller to know (or care) that octokit was used to make the request.

packages/utilities/exposure-scanning/patch/index.js
andrewscfc: Thanks for the explaination @HarryVerhoef you might want to add why you want to limit the calls in the comment as you've explained here

andrewscfc (approved):
LGTM, thanks for sticking with all my comments!

ryanmccombe (commented):
Nice one 🚀 

I think a README.md might be nice to explain how to use it?  Especially if it's going to be on npm

packages/utilities/exposure-scanning/fetch/index.js
```diff
@@ -0,0 +1,58 @@
...
+    data: { body },
+  } = await octokit.request('GET /repos/{owner}/{repo}/issues/{id}', reqBody);
+  return body;
+};
+
+export const fetchPr = async reqBody => ({
+  body: await fetchPrBody(reqBody),
+  comments: await fetchComments(reqBody),
+  reviewComments: await fetchReviewComments(reqBody),
+});
```
jroebu14: would something like this be quicker/more efficient?

```js
export const fetchPr = async reqBody => {
  const [body, comments, reviewComments] = Promise.all([
    fetchPrBody(reqBody),
    fetchComments(reqBody),
    fetchReviewComments(reqBody),
  ]);

  return {
    body,
    comments,
    reviewComments,
  };
};
```

HarryVerhoef:
> I think a README.md might be nice to explain how to use it? Especially if it's going to be on npm

Thanks @ryanmccombe! Yeah good idea, I'll add a README at some point today 👍

packages/utilities/exposure-scanning/src/index.js
```diff
@@ -0,0 +1,47 @@
...
+import { patchPr, patchIssue } from '../patch/index';
+
+const redactPr = async (reqBody, regex) => {
+  const pr = await fetchPr(reqBody);
+
+  const scannedPr = await scanPr(pr, regex);
+
+  if (scannedPr.foundMatch) {
+    await patchPr(reqBody, scannedPr);
+    throw new Error('Match found.');
```
HarveyPeachey: I'm not familiar with throw and it's usage so bare with. 

So when the PR/Issue has been patched after a match was found, does the throw just exit the code here and return the error string `Match found`, which in turn means it won't reach line 44 where it returns `No matches found.`? If so shall it also include that it has been patched alongside finding a match in the error string?

HarryVerhoef: Yes that's correct 👍 And yeah good idea regarding the error message, I'll make that change now

HarveyPeachey (approved):
This looks great!

Not sure what the scope is for this scanning stuff but it would be cool to have some sort of log for when these exposures happen, or a way to store the history privately before they're patched

packages/utilities/exposure-scanning/args/index.js
```diff
@@ -0,0 +1,45 @@
...
+  const flag = argv[3]; // -pr or -issue
+  const id = argv[4];
+  const regexString = argv[5];
+
+  if (!isValidRegex(regexString)) {
+    throw new Error('Invalid regex argument given.');
+  }
+
+  if (!isValidId(id)) {
+    throw new Error('Invalid issue/pr id.');
```
jroebu14, line 35: would this ever happen?

packages/utilities/exposure-scanning/args/index.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  if (flag !== '-pr' && flag !== 'issue') {
```
jroebu14: maybe this is more readable

```suggestion
  const isValidFlag = ['-pr', 'issue'].includes(flag)

  if (!isValidFlag) {
```

but would this ever occur?

HarryVerhoef:
Thanks @HarveyPeachey! The scope is to identify and redact any exposures in PRs and issues 👍 Yeah we have to be careful when logging in this instance, but agree that a private logging solution would be ideal.

packages/utilities/exposure-scanning/fetch/index.js
HarryVerhoef: Good idea 👍

packages/utilities/exposure-scanning/args/index.js
HarryVerhoef: Yeah if a non-integer argument is supplied then this should throw. It's unlikely to happen once integrated with the GitHub action, but since it's a psammead package and may be run by others outside of GitHub actions I think it may be necessary

packages/utilities/exposure-scanning/args/index.js
HarryVerhoef: Good idea 👍 And yeah same as the id parsing - Unlikely to happen in our case since the action will invoke the script, but can also be run with user-input command line args

packages/utilities/exposure-scanning/fetch/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+export const fetchIssue = async reqBody => ({
+  body: await fetchIssueBody(reqBody),
```
jroebu14: same as above here

PR #4536, opened 2021-08-17
Remove dependency on fixture for onRender prop

HarryVerhoef:
**Overall change:** Allows the psammead-social-embed to correctly function when integrated into Simorgh.

**Code changes:**

- Uses a string literal for the provider name instead of referring to a fixture

---

- [X] (BBC contributors only) This PR follows the [repository use guidelines](https://github.com/bbc/simorgh-infrastructure/blob/latest/documentation/repository-guidelines.md)

### DIFF
--- a/packages/components/psammead-social-embed/src/Canonical/index.jsx
+++ b/packages/components/psammead-social-embed/src/Canonical/index.jsx
@@ -73,14 +73,14 @@ export const providers = {
 };
 
 const CanonicalEmbed = ({ provider, oEmbed, onRender }) => {
-  const hasLibraryLoaded = useScript(providers[provider].script);
+  const libraryHasLoaded = useScript(providers[provider].script);
   useEffect(providers[provider].enrich);
 
   useEffect(() => {
-    if (provider === 'twitter' && hasLibraryLoaded && onRender) {
+    if (provider === 'twitter' && libraryHasLoaded && onRender) {
       providers.twitter.onLibraryLoad(onRender);
     }
-  }, [hasLibraryLoaded]);
+  }, [libraryHasLoaded]);
 
   return (
     <OEmbed
--- a/packages/components/psammead-social-embed/src/Canonical/index.jsx
+++ b/packages/components/psammead-social-embed/src/Canonical/index.jsx
@@ -73,14 +73,14 @@ export const providers = {
 };
 
 const CanonicalEmbed = ({ provider, oEmbed, onRender }) => {
-  const isSdkLoaded = useScript(providers[provider].script);
+  const hasLibraryLoaded = useScript(providers[provider].script);
   useEffect(providers[provider].enrich);
 
   useEffect(() => {
-    if (provider === 'twitter' && isSdkLoaded && onRender) {
+    if (provider === 'twitter' && hasLibraryLoaded && onRender) {
       providers.twitter.onSdkLoad(onRender);
     }
-  }, [isSdkLoaded]);
+  }, [hasLibraryLoaded]);
 
   return (
     <OEmbed
--- a/packages/components/psammead-social-embed/src/Canonical/index.jsx
+++ b/packages/components/psammead-social-embed/src/Canonical/index.jsx
@@ -47,7 +47,7 @@ export const providers = {
         window.twttr.widgets.load();
       }
     },
-    onSdkLoad: onRender => {
+    onLibraryLoad: onRender => {
       window.twttr.ready(twttr => {
         twttr.events.bind('rendered', onRender);
       });
@@ -78,7 +78,7 @@ const CanonicalEmbed = ({ provider, oEmbed, onRender }) => {
 
   useEffect(() => {
     if (provider === 'twitter' && hasLibraryLoaded && onRender) {
-      providers.twitter.onSdkLoad(onRender);
+      providers.twitter.onLibraryLoad(onRender);
     }
   }, [hasLibraryLoaded]);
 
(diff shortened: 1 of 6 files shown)


Louis-Matsika (approved):

packages/components/psammead-social-embed/src/Canonical/index.jsx
```diff
@@ -78,8 +77,8 @@ const CanonicalEmbed = ({ provider, oEmbed, onRender }) => {
   useEffect(providers[provider].enrich);
 
   useEffect(() => {
-    if (provider === fixtures.twitter.source && isSdkLoaded && onRender) {
-      providers[fixtures.twitter.source].onSdkLoad(onRender);
+    if (provider === 'twitter' && isSdkLoaded && onRender) {
```
andrewscfc: Not one for this PR but looking back at the previous PR, I think it might have been clearer if `onRender` was `onTwitterEmbedRender` or something. Better to genericise it later if the functionality was added to other Embed types.

andrewscfc (approved):
LGTM, just a note to consider following up on

packages/components/psammead-social-embed/src/Canonical/index.jsx
jroebu14: being nitpicky but `isSdkLoaded` is like a question rather than a fact. What about renaming to `hasLoadedLibrary` or something instead?

packages/components/psammead-social-embed/src/Canonical/index.jsx
```diff
@@ -78,8 +77,8 @@ const CanonicalEmbed = ({ provider, oEmbed, onRender }) => {
...
+      providers.twitter.onSdkLoad(onRender);
```
jroebu14: then maybe 
```suggestion
      providers.twitter.onLibraryLoaded(onRender);
```

packages/components/psammead-social-embed/src/Canonical/index.jsx
jroebu14: what about making this like:

```jsx
  useEffect(() => {
    const { onLibraryLoad } = providers[provider]

    if (hasLoadedLibrary && onLibraryLoad) {
      onLibraryLoad(onRender)
    }
  }, [hasLoadedLibrary]);
```

and making `onRender` a no-op function in default props e.g.

```jsx
CanonicalEmbed.defaultProps = {
  onRender: () => {},
};
```

HarryVerhoef: Thanks @andrewscfc and @jroebu14!

These are both great suggestions, but I think they lean in opposite directions - I'm not sure if it makes sense to combine them since one increases specificity and the other generalises/future-proofs. A bit torn 😂

jroebu14: For the YouTube and Instagram `onLibraryLoad` methods you could console log an error `onRender callback function not implemented for YouTube` or something

HarryVerhoef: Yeah good idea 👍

packages/components/psammead-social-embed/src/Canonical/index.jsx
```diff
@@ -74,14 +73,14 @@ export const providers = {
 };
 
 const CanonicalEmbed = ({ provider, oEmbed, onRender }) => {
-  const isSdkLoaded = useScript(providers[provider].script);
+  const hasLibraryLoaded = useScript(providers[provider].script);
```
jroebu14: ~sorry can it be `libraryHasLoaded`?~

jroebu14: sorry I originally wrote `hasLoadedLibrary `

packages/components/psammead-social-embed/src/Canonical/index.jsx
```diff
@@ -34,6 +34,8 @@ export const providers = {
         window.instgrm.Embeds.process();
       }
     },
+    onLibraryLoad: () =>
+      console.error('onRender callback function not implemented for Instagram'),
```
jroebu14: You'll only want to log an error when the consumer provides an `onRender` function for YT and Insta.

To do this you could make the `onRender` default prop `null` again and something like

```suggestion
    onLibraryLoad: onRender => {
      if (onRender) {
        console.error(getOnRenderError('Instagram'))
      }
    }
```

jroebu14: sorry for all the changes!

jroebu14: you'll also need to check to do `if (onRender) {` in the twitter one too before binding to the event

HarryVerhoef: Thanks @jroebu14, these are really helpful suggestions 🙂 Just a query, would we not be able to add the `onRender` condition to the `onLibraryLoad` call in the `useEffect`?
Like
```jsx
if (onRender && hasLoadedLibrary && onLibraryLoad) {
      onLibraryLoad(onRender);
}
```

then

```jsx
onLibraryLoad: () => getOnRenderError('Instagram'),
```

jroebu14: yeh that sounds better! sorry for going a little full circle on this one

jroebu14 (approved):

PR #4563, opened 2021-09-21
Fix package versioning script

jroebu14:
**Overall change:**
Fixes bug where the package versioning script broke after versioning the first package from a list of packages as the input.

This occurred because Yarn 3 needs to version packages synchronously.

**Code changes:**

- Removes the async versioning of packages and uses synchronous commands

---


scripts/bumpPackages/index.js
```diff
@@ -1,56 +1,49 @@
-const { exec } = require('child_process');
+const { execSync } = require('child_process');
```
HarveyPeachey, line 1: Ahh I attempted to use this, I tried to get it to work with promises which makes sense why it didn't work. I need to look up more about async and sync in JS.

HarveyPeachey (approved):
Looks great! There is a command introduced in Yarn v2/v3 [`yarn version check --interactive`](https://yarnpkg.com/cli/version/check/), which is similar and provides an interactive way to bump, it's a combination of this script and change scanner, and it also allows you to pick the strategy for each individual workspace. But it errors as it [doesn't support prerelease packages](https://github.com/yarnpkg/berry/blob/a7cf5e5b23200299c17c6965afcd0d83d56f6281/packages/plugin-version/sources/commands/version/check.tsx#L104-L105). I might look into submitting a PR to the Yarn repo on a learning day to try and support prerelease packages

jroebu14:
@HarveyPeachey thanks for reviewing. That yarn command sounds really good. I'll check it out 👍

ryanmccombe (approved):

PR #4608, opened 2021-12-15
Add `primaryMimeType` and `fallbackMimeType` props to `psammead-image`

amoore108:
Part of https://github.com/bbc/simorgh/issues/9628

**Overall change:**
Derives the `<source />` mime-type by checking the srcsets of the image. This appears to be required by some older browsers that cannot derive the mime-type themselves from the server response.

**Code changes:**

- Updates the `psammead-image` canonical component to check the `srcset` and `fallbackSrcset` URLs to determine what the mime-type of the respective `<source />` tag should be set to.
---

### DIFF
--- a/packages/components/psammead-image/src/index.jsx
+++ b/packages/components/psammead-image/src/index.jsx
@@ -45,7 +45,7 @@ export const Img = props => {
 
   return (
     <StyledPicture onLoad={onLoad}>
-      <source srcSet={srcset} type={primaryMimeType} />
+      {srcset && <source srcSet={srcset} type={primaryMimeType} />}
       {fallbackSrcset && (
         <source srcSet={fallbackSrcset} type={fallbackMimeType} />
       )}
@@ -74,8 +74,8 @@ Img.defaultProps = {
   sizes: null,
   srcset: null,
   fallbackSrcset: null,
-  primaryMimeType: null,
-  fallbackMimeType: null,
+  primaryMimeType: 'image/jpeg',
+  fallbackMimeType: 'image/jpeg',
   width: null,
   onLoad: () => {},
 };
--- a/packages/components/psammead-image/src/index.jsx
+++ b/packages/components/psammead-image/src/index.jsx
@@ -32,13 +32,43 @@ const StyledImg = styled.img`
   width: 100%;
 `;
 
+const getMimeType = srcset => {
+  if (!srcset) return null;
+
+  const firstSrcsetUrl = srcset.split(',')[0].split(' ')[0];
+  const urlFileExtension = firstSrcsetUrl.split('.').pop();
+
+  switch (urlFileExtension) {
+    case 'webp':
+      return 'image/webp';
+    case 'jpg':
+    case 'jpeg':
+      return 'image/jpeg';
+    case 'png':
+      return 'image/png';
+    default:
+      return null;
+  }
+};
+
 export const Img = props => {
   const { src, srcset, fallbackSrcset, onLoad, ...otherProps } = props;
 
+  const primaryMimeType = getMimeType(srcset);
+  const secondaryMimeType = getMimeType(fallbackSrcset);
+
   return (
     <StyledPicture onLoad={onLoad}>
-      <source srcSet={srcset} />
-      {fallbackSrcset && <source srcSet={fallbackSrcset} />}
+      <source
+        srcSet={srcset}
+        {...(primaryMimeType && { type: primaryMimeType })}
+      />
+      {fallbackSrcset && (
+        <source
+          srcSet={fallbackSrcset}
+          {...(secondaryMimeType && { type: secondaryMimeType })}
+        />
+      )}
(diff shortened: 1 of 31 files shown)


ryanmccombe (approved):

packages/components/psammead-image/src/index.jsx
```diff
@@ -32,13 +32,45 @@ const StyledImg = styled.img`
...
+  const primaryMimeType = getMimeType(srcset);
+  const secondaryMimeType = getMimeType(fallbackSrcset);
+
   return (
     <StyledPicture onLoad={onLoad}>
-      <source srcSet={srcset} />
-      {fallbackSrcset && <source srcSet={fallbackSrcset} />}
+      <source
+        srcSet={srcset}
+        {...(primaryMimeType && { type: primaryMimeType })}
```
jroebu14: not sure you need to conditionally spread like this. I think if the value is null then the attribute doesn't get rendered

packages/components/psammead-image/src/index.jsx
```diff
@@ -32,13 +32,45 @@ const StyledImg = styled.img`
   width: 100%;
 `;
 
+const getMimeType = srcset => {
+  if (!srcset) return null;
+
+  const firstSrcsetUrl = srcset.split(',')[0].split(' ')[0];
```
jroebu14: just wondering if this is easier to work out in simorgh

amoore108: ~I think its much of a muchness really. It would probably be a bit more work in Simorgh to pass through a `mimeType` prop or something similar to the Psammead component, with similar checking done in Simorgh. Conscious not having too much rendering logic in Psammead though.~

Moved this over to Simorgh and passed through as props

packages/components/psammead-image/src/index.jsx
```diff
@@ -32,13 +32,39 @@ const StyledImg = styled.img`
...
+      return 'image/webp';
+    case 'jpg':
+    case 'jpeg':
+      return 'image/jpeg';
+    case 'png':
+      return 'image/png';
+    case 'gif':
+      return 'image/gif';
+    default:
+      return null;
```
andrewscfc: don't we support the `raw` extension as well? Are there any other extensions missing here?

packages/components/psammead-image/src/index.jsx
```diff
@@ -32,13 +32,39 @@ const StyledImg = styled.img`
   width: 100%;
 `;
 
+const getMimeType = srcset => {
+  if (!srcset) return null;
+
+  const firstSrcsetUrl = srcset.split(',')[0].split(' ')[0];
```
andrewscfc:
```suggestion
  // ?. captures an edge case where the srcset passed in is null or invalid
  const [ firstSrcSet ] = srcset?.split(',');
  const [ firstSrcSetUrl ] = firstSrcSet?.split(' ');
```
This reads a bit more clearly imo. The main bit is we need to be a bit defensive if the src set is invalid, so null or contains an unexpected format. I believe if the srcset was invalid in this way the page would not render at all with the code as it is; I'd write a test to cover this eventuality.

I'd agree with @jroebu14 that this logic may be better to be in Simorgh, is that where the srcset is constructed; you could pass the mime type in as a prop?

packages/components/psammead-image/src/index.jsx
amoore108: `raw` is used as an iChef recipe, so something like `https://ichef.bbci.co.uk/news/raw/myfile.jpg`. It doesn't support the '.raw' file extension as it were.

packages/components/psammead-image/src/index.jsx
amoore108: I'll move this logic over to Simorgh and pass through the mime-types as props for each of the srcsets

jroebu14 (approved):
The component API looks good. Do you think we should have defaults for the new props though?

andrewscfc (approved):

amoore108:
> The component API looks good. Do you think we should have defaults for the new props though?

Its probably safe enough to have defaults of `image/jpeg`, although we do use PNG a couple of times thoughout, so I wouldn't want to have the situation of a PNG srcset and an `image/jpeg` type being set for example. Unlikely, but possible I think, although happy to add them if its more sensible?

amoore108:
> > The component API looks good. Do you think we should have defaults for the new props though?
> 
> Its probably safe enough to have defaults of `image/jpeg`, although we do use PNG a couple of times thoughout, so I wouldn't want to have the situation of a PNG srcset and an `image/jpeg` type being set for example. Unlikely, but possible I think, although happy to add them if its more sensible?

I've added a conditional check for the primary `srcset` as well and set mime-type defaults. This should cover most cases I think, as anywhere in Simorgh that is using the psammead-image should now be deriving mime-types and srcsets. And if not, it won't render an empty `<source />` tag and just return the basic `` tag inside the picture element.

ryanmccombe (approved):

PR #4615, opened 2022-02-09
Add data-hero to Psammead's amp-img

DarioR01:
Partial resolution of [#9801](https://github.com/bbc/simorgh/issues/9801)

Adds data-hero attribute to Psammead amp-img.

**Code changes:**

- adds data-hero to all amp-img in Psammead

---

- [ ] (BBC contributors only) This PR follows the [repository use guidelines](https://github.com/bbc/simorgh-infrastructure/blob/latest/documentation/repository-guidelines.md)


amoore108 (approved):

pvaliani (approved):

HarveyPeachey:
@DarioR01 Is this `data-hero` attribute being used correctly here? My understanding is that it should only be put on the most substantial image on the page so it loads that image first as that's what a user would be shown in the viewport. It looks like it's being applied to all images looking at the PR? Or have I missed something?

DarioR01:
> ding is that it should only be put on the

My understanding is that data-hero is used for AMP Caches and AMP Optimizers to server-side render the image. Normally, as you suggested, this would only be implemented on the most important images ("hero images"). Since there is no list of images that are "most important" on our pages, and technically speaking there is nothing that does not allow me to place this in every single image, I assumed this should have been applied to every image.
I will double-check to be on the safe side.

HarveyPeachey:
> My understanding is that data-hero is used for AMP Caches and AMP Optimizers to server-side render the image. 

Yeah from that explanation then, if it serverside renders all the images first as a priority, let's say on a PGL page with lots of images, the server will prioritise loading rendering those images first as they all have the `data-hero` attribute, which I suspect would increase the Total Blocking Time as the page won't be sent to the user until all those images have been loaded server-side, which may have a worse impact on the core web vitals score. It's best if you test to see if it makes a difference in the score as this is probably okay for assets where there are only a couple of images, but might be bad for pages with a lot of images.

DarioR01:
You are right, I should have had data-hero only on the first image... oh well, time to revert

PR #4620, opened 2022-02-10
Update http-signature to fix jsprim vulnerability

MuhammadAHussain:
Resolves N/A

**Overall change:** 
Updated http-signature version which uses jsprim patch version

**Code changes:**

- Add resolution for `http-signature` package.

---


MuhammadAHussain:
Got a question regarding this package bump. `http-signature` uses `jsprim` version 2.0.2 which has the `js-schema` patch version. However I noticed on the `jsprim` repo, a tag for backport was created. I opted to update the `http-signature` package instead of adding a resolution to `jsprim` version `1.4.2`. Which is the better option (unsure what a backport does tbh)?

amoore108:
> Got a question regarding this package bump. `http-signature` uses `jsprim` version 2.0.2 which has the `js-schema` patch version. However I noticed on the `jsprim` repo, a tag for backport was created. I opted to update the `http-signature` package instead of adding a resolution to `jsprim` version `1.4.2`. Which is the better option (unsure what a backport does tbh)?

I think updating `http-signature` makes sense since its the only library using `jsprim` and `json-schema`.

The backporting just applies fixes to older versions of the library. So for example if the library updated with breaking changes, some people would still want critical security fixes on older versions of the same library without having to update to the version with breaking changes.

amoore108 (approved):

MuhammadAHussain:
> > Got a question regarding this package bump. `http-signature` uses `jsprim` version 2.0.2 which has the `js-schema` patch version. However I noticed on the `jsprim` repo, a tag for backport was created. I opted to update the `http-signature` package instead of adding a resolution to `jsprim` version `1.4.2`. Which is the better option (unsure what a backport does tbh)?
> 
> I think updating `http-signature` makes sense since its the only library using `jsprim` and `json-schema`.
> 
> The backporting just applies fixes to older versions of the library. So for example if the library updated with breaking changes, some people would still want critical security fixes on older versions of the same library without having to update to the version with breaking changes.

Sweet. Thanks for the explanation. I agree, probably best to update the `http-signature` package then.

ryanmccombe (approved):

